Issue #41, opened 2014-11-27
manage per box settings


yohanboniface:
Like we have a user that can be specific to a box, but versionned, we want to have some settings specific for some box BUT still versionned.

yohanboniface:
Done via #62

PR #62, opened 2014-12-17
Add per box settings (cf #41)

yohanboniface:
This PR also:

- fixes model mixins that was not inheriting from `models.Model`, so fields was not active
- fixes `form-fullpage.html` template that was overriding `title` block
- introduces a `user.public_fields` method, allowing to loop over user fields, even with a custom user model

If it's better, I can split this into other PRs. :)

### DIFF
--- a/ideasbox/settings.py
+++ b/ideasbox/settings.py
@@ -1,161 +1,25 @@
-"""
-Django settings for ideasbox project.
-
-For more information on this file, see
-https://docs.djangoproject.com/en/1.7/topics/settings/
-
-For the full list of settings and their values, see
-https://docs.djangoproject.com/en/1.7/ref/settings/
-"""
-
+import importlib
 import os
+import socket
+import sys
 
-from django.utils.translation import ugettext_lazy as _
+# The normal scenario is that we use the hostname, but let's make it
+# overridable, this is useful for dev and debugging.
+IDEASBOX_ID = os.environ.get('IDEASBOX_ID', socket.gethostname())
 
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(__file__))
-PROJECT_DIR = os.path.join(BASE_DIR, 'ideasbox')
-
-# Quick-start development settings - unsuitable for production
-# See https://docs.djangoproject.com/en/1.7/howto/deployment/checklist/
-
-# SECURITY WARNING: keep the secret key used in production secret!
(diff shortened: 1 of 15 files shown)


ideasbox/settings.py
```diff
@@ -1,161 +1,25 @@
...
+except ImportError:
+    from .conf import base as sub
+finally:
+    SETTINGS_MODULE = sub.__name__  # Make if available as a settings, to
+                                    # be able to display it in the admin
+    sys.stdout.write('Importing settings from %s\n' % SETTINGS_MODULE)
+    ldict = locals()
+    for k in sub.__dict__:
+        if k.isupper() and not k.startswith('__') or not k.endswith('__'):
+            ldict[k] = sub.__dict__[k]
```
magopian: I understand your point, and I may be thick, but really I'm convinced this isn't the right place to do it (now that you kind of convinced me you need to do it).

Settings are a place to set settings, not dynamically do anything. I'd recommend doing this hackery (which will thus be much more limited and much simpler) in the `manage.py`, using the hostname to choose what to store in the `DJANGO_SETTINGS_MODULE` (this way it'll simply be a string interpolation, instead of this module import fiddling).

What do you think?

yohanboniface: Humm, but, if we do that in the manage.py, it will not work when calling the wsgi module directly from gunicorn, isn't it?
Also, in any case, we need to be sure the hostname specific setting exist in any case before trying to load it, and if not fallback on the default, so we will certainly need to do some more hackery than just building the string to give Django, I guess.

magopian: I didn't think about that, but that leads me to another question ;)

If you need to create a custom settings file (different than the default one), why not simply use this one? You could then simply use the "usual" 
```python
try:
    from local_settings import *
except ImportError:
    pass
```
trick. If the `local_settings` is needed (because you need a custom settings file for this box), then you need to create it anyway, no need to add an extra level of complexity. Not sure this is clear enough, I'm available to discuss it via irc, maybe I'm just missing the point ;)

yohanboniface: If I get your point correctly, the answer is that we need each "per box" setting file to be versionned, so we can't only use a generic local_settings. We try to avoid any "hidden" setting.

Also, if we try to import "per box" settings from the settings file itself, instead of importing the base settings from the "per box" settings files, this need that we can't only adapt base settings (like adding an entry in a list or a dict), we will always need to override all the entry of a list or tuple or dict each time we wan't to have one custom element inside.

If I'm out of the point, let's discuss this in IRC, yep :)

yohanboniface:
r+ed in IRC :)

Issue #38, opened 2014-11-26
Search for blog posts


yohanboniface:

yohanboniface:
On SQLite FTS also: http://charlesleifer.com/blog/using-sqlite-full-text-search-with-python/

yohanboniface:
Also: http://chipaca.com/post/16877190061/doing-full-text-search-in-sqlite-from-python

yohanboniface:
Done via #64.

Issue #12, opened 2014-11-16
books browser


yohanboniface:
A page to:
- search for books
- browse books by: author, alphabetical order, tags

PR #64, opened 2015-01-04
First shot on full-text search

yohanboniface:
Ref: #12, #38

Still following the "make it simple for now, avoid as much as possible dependencies" mantra, I've implemented a quick full text search using the FTS functionality of SQLite.

Basically, we create some special virtual table at post migrate, and we have a dedicated model that will index data on it.
Then any model that we want searchable should inherit from a `SearchableMixin` that will give this model the needed methods (`index` and `deindex` for example), and also make that the instance will be indexed at post save and deindexed at post delete.
A QuerySet mixin is also provided to add a `search` method.
It's all very simple and djangoic; the only tricky part is the ranking: the SQLite FST doesn't provide any rank function to score the result by relevancy, so we needed to code our.

Also a fist version of a search box and search page has been added.


blog/models.py
```diff
@@ -63,3 +54,11 @@ def get_absolute_url(self):
 
     def get_author_display(self):
         return self.author_text or unicode(self.author)
+
+    @property
+    def index_strings(self):
+        return [self.title, self.text, self.author_text, unicode(self.author)]
+
+    @property
+    def index_public(self):
```
davidbgk: I would have named that one `is_published`.

blog/models.py
```diff
@@ -4,28 +4,22 @@
 from django.utils.translation import ugettext_lazy as _
 
 
+from search.models import SearchMixin, SearchableQuerySet
```
davidbgk: Sort imports alphabetically :p #OhWait

ideasbox/templatetags/ideasbox_tags.py
```diff
@@ -0,0 +1,23 @@
...
+}
+THEMES = {
+    'Book': 'read',
+    'Content': 'create'
+}
+
+
+@register.filter(is_safe=True)
+def theme_slug(inst, slug=None):
+    tpl = '<span class="theme {0}">{1}</span>'
```
davidbgk: Named vars in a template are way more readable.

library/models.py
```diff
@@ -54,6 +55,11 @@ def __unicode__(self):
     def get_absolute_url(self):
         return reverse('library:book_detail', kwargs={'pk': self.pk})
 
+    @property
+    def index_strings(self):
+        return [self.title, self.isbn, self.authors, self.subtitle,
```
davidbgk: Why not a tuple?

search/templates/search/box.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+
+<div class="card tinted search">
+    <form action="{% url "search:search" %}">
+        <input name="q" type="text" placeholder="{% trans 'search' %}" value="{{ q }}" /><!--
```
davidbgk: `{% spaceless %}`?

search/models.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @classmethod
+    def search(cls, **kwargs):
+        qs = Search.objects.filter(**kwargs).order_by_relevancy()
+        for row in qs:
+            key = '.'.join([row.module, row.model])
+            if not key in _SEARCHABLE:
+                _SEARCHABLE[key] = import_string(key)
+            yield _SEARCHABLE[key].objects.get(pk=row.model_id)
+_SEARCHABLE = {}
```
davidbgk: Smells badly.

blog/models.py
yohanboniface: Fixed :) #amostyle

ideasbox/templatetags/ideasbox_tags.py
yohanboniface: Done :)

library/models.py
yohanboniface: Good question :)
But, out of curiosity, why a tuple? :)
(I've moved to tuples, anyway.)

search/templates/search/box.html
yohanboniface: Good point!

search/models.py
yohanboniface: Meeeeeeh.
I first thought about a metaclass, then I remembered that metaclass are the start of unreadable code, so I've ended with this Q&D option.
Do you think it's worth a registry and a metaclass? Or any other option in mind?

library/models.py
davidbgk: For [performances](https://stackoverflow.com/questions/3340539/why-is-tuple-faster-than-list).

blog/models.py
davidbgk: You forgot to bubble with the line below :D

search/models.py
davidbgk: Do you really need to cache the import mechanism? How does it affect performances?

yohanboniface: Are you asking me to change lists by tuples in one hand and to import modules at each method call in another? :p

Honestly, ideas box challenge is not really the performances (but the simplicity, robustness, offline…), so I can remove this caching without any real issue on the usability I think.

But still it'd we a bit clod to make this import at each row of each search result, nope?

davidbgk: :p

Well, you can start by caching it for the forloop at first in this case.

yohanboniface: Humm, but if I want this to be dynamic, I need to loop over the result to know which models are available and need to be imported.

Another option, is to have them hardcoded. DRY alert, but maybe less ugly than the on the fly caching, and less complex/overkill than a metaclass plus a registry.

yohanboniface: `class_prepared` signal was the solution I think: https://github.com/ideas-box/ideasbox.lan/commit/d7ab7a53013785170520b752ab5bb9ed01944927

davidbgk: Still not fond of the global dict but that smells better to me ;)

blog/tests/test_search.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from search.models import Search
+
+from ..models import Content
+
+pytestmark = pytest.mark.django_db
```
magopian: Flake8 doesn't complain that this variable isn't used?

blog/tests/test_search.py
```diff
@@ -0,0 +1,67 @@
...
+def test_nothing_is_indexed_without_any_fixture():
+    assert Search.objects.count() == 0
+
+
+def test_draft_is_indexed(draft):
+    assert Search.objects.count() == 1
+    assert len(list(Search.search(public=True))) == 0
+
+
+def test_deleted_is_indexed(deleted):
```
magopian: typo in the name? Shouldn't it be "deindexed"?

magopian: Ah, forget it, my bad, sorry ;)

blog/tests/test_search.py
```diff
@@ -0,0 +1,67 @@
...
+    assert Content.objects.search("ikinyugunyugu").count() == 1
+
+
+def test_we_can_search_arabic_content(published):
+    published.title = u"أكثر من خمسين لغة،"
+    published.save()
+    assert Content.objects.search(u"خمسين").count() == 1
+
+
+def test_we_can_search_with_jocker(published):
```
magopian: I don't think "joker" takes a "c"

blog/tests/test_search.py
yohanboniface: Humm, I don't think so because it's on the first level of the module, so it can be imported from elsewhere.

search/apps.py
```diff
@@ -0,0 +1,16 @@
+from django.apps import AppConfig
+from django.db.models.signals import post_migrate
+
+from .utils import create_index_table
+
+
+def create_index(sender, **kwargs):
+    create_index_table()
```
magopian: What's the use of this indirection?

search/management/commands/reindex.py
```diff
@@ -0,0 +1,19 @@
...
+
+from blog.models import Content
+from library.models import Book
+from search.utils import create_index_table
+
+
+class Command(BaseCommand):
+    help = 'Reindex all the searchable objects'
+
+    def handle(self, *args, **options):
```
magopian: why `options` and not `kwargs`?

search/models.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class Match(models.Lookup):
+    lookup_name = 'match'
+
+    def as_sql(self, qn, connection):
+        lhs, lhs_params = self.process_lhs(qn, connection)
+        rhs, rhs_params = self.process_rhs(qn, connection)
+        params = lhs_params + rhs_params
+        return '%s MATCH %s' % (lhs, rhs), params
```
magopian: use `.format`?

search/apps.py
yohanboniface: Humm, you mean why not registering directly create_index_table` to the signal instead? If yes, the answer is that the function is also used in the management command, so it needs a neutral signature.

search/management/commands/reindex.py
yohanboniface: bad habits ;)

search/models.py
yohanboniface: fixed :)

search/models.py
```diff
@@ -0,0 +1,120 @@
...
+    @property
+    def index_strings(self):
+        return []
+
+    @property
+    def index_public(self):
+        return True
+
+    def is_indexable(self):
+        return True
```
magopian: What is that for? Do you think you'll have models inheriting from `SearchMixin` but aren't indexable?

yohanboniface: Not the model itself, but some instances under some circumstance yep.

search/models.py
```diff
@@ -0,0 +1,120 @@
...
+    def search(self, query, **kwargs):
+        kwargs['text__match'] = query
+        kwargs['model'] = self.model.__name__
+        ids = Search.ids(**kwargs)
+        return self.filter(pk__in=ids)
+
+
+@receiver(post_save)
+def index(sender, instance, **kwargs):
+    if SearchMixin in sender.__mro__:
```
magopian: wait, what? Isn't it possible to simply use `issubclass`?

search/tests/test_models.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+from blog.tests.factories import ContentFactory
+from ..models import Search
+
+
+pytestmark = pytest.mark.django_db
+
+
+def test_nothing_is_indexed_without_any_fixture():
```
magopian: exact same test as https://github.com/ideas-box/ideasbox.lan/pull/64/files#diff-f1381635bf10f8354e878ef3e776a5c0R11 ?

search/models.py
yohanboniface: Damn, yes, good point, thanks :)

search/tests/test_models.py
yohanboniface: Good catch, I'm removing the other one :)

search/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    cursor = connection.cursor()
+    cursor.execute("DROP TABLE IF EXISTS idx")
+    cursor.execute("CREATE VIRTUAL TABLE idx using "
+                   "FTS4(id, model, model_id, public, text)")
+
+
+def rank(match_info):
+    # Handle match_info called w/default args 'pcx' - based on the example
+    # rank function http://sqlite.org/fts3.html#appendix_a
+    # From github.com/coleifer/peewee/master/playhouse/sqlite_ext.py
```
magopian: I know the following code is taken from somewhere else, but it would help to have some more comments in the following lines to understand what it's doing. It's a bit cryptic, as is.

magopian:
I can't say much on the full text search implementation itself, I'm not well aware of sqlite or the FTS functionality, but it looks good from what I can tell.

r+wc

yohanboniface:
Thanks for the review, @davidbgk and @magopian :)

Issue #27, opened 2014-11-21
Media Center


yohanboniface:
UI to:

- upload a document (can be PDF, doc, vidéo, etc.) with basic metadatas (title, description, tags, datetime of creation/edition) ; only staff users for now
- browse those documents (list, pagination)
- download those documents

Optionally for iteration 1:
- manage previews images to have a more friendly browsing

Apps to consider:
- http://django-media-tree.readthedocs.org/en/latest/
- https://github.com/shelldweller/mediabrowser
- http://mayan.readthedocs.org/en/latest/ (http://www.mayan-edms.com/, certainly overkill)
- https://github.com/caktus/django-file-picker

PR #67, opened 2015-01-15
Media center (cf #27)

yohanboniface:
Ref: #27


ideasbox/views.py
```diff
@@ -15,9 +16,11 @@
 def index(request):
     contents = Content.objects.published()[:3]
     random_book = Book.objects.available().order_by('?').first()
+    random_doc = Document.objects.order_by('?').first()
```
twidi: I don't know for sqlite but generally it's not recommended to sort sql data by "random"

mediacenter/tests/test_views.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_can_create_document(staffapp):
+    assert not Document.objects.count()
+    form = staffapp.get(reverse('mediacenter:document_create')).form
+    form['title'] = 'my document title'
+    form['summary'] = 'my document summary'
+    form['credits'] = 'my document credits'
+    form['original'] = Upload('image.jpg', 'xxxxxx')
+    form.submit().follow()
+    assert Document.objects.count()
```
twidi: better assert that count == 1 to be sure that the form will create only one doc

ideasbox/views.py
yohanboniface: One of the specifics of Ideas Box server, is that we will be running with very small dataset and under very low load, so I think this will scale in our context :)

ideasbox/templates/ideasbox/includes/document_card.html
```diff
@@ -0,0 +1,7 @@
+{% load i18n mediacenter_tags %}
+<div class="card tinted document">
+    <img src="{{ document|preview_url }}" />
```
twidi: title ?

ideasbox/templates/ideasbox/includes/document_card.html
```diff
@@ -0,0 +1,7 @@
...
+    <h4 class="flow"><span class="theme read">read</span> <span>{% trans "medias center" %}</span></h4>
+    <h3 class="flow"><a href="{% url 'mediacenter:document_detail' pk=document.pk %}">{{ document }}</a></h3>
```
twidi: why not get_absolute_url ?

ideasbox/views.py
twidi: ok, I preferred to ask anyway ;)

mediacenter/models.py
```diff
@@ -0,0 +1,78 @@
...
+    title = models.CharField(_('title'), max_length=100)
+    summary = models.TextField(_('summary'))
+    lang = models.CharField(_('Language'), max_length=10, blank=True,
+                            choices=settings.LANGUAGES)
+    original = models.FileField(_('original'),
+                                upload_to='mediacenter/document')
+    preview = models.ImageField(_('preview'), upload_to='mediacenter/preview',
+                                blank=True)
+    credits = models.CharField(_('credit'), max_length=300)
+    type = models.CharField(_('type'), max_length=5, choices=TYPE_CHOICES,
```
twidi: not fan to use python "reserved" words

mediacenter/models.py
```diff
@@ -0,0 +1,78 @@
...
+        return self.title
+
+    def get_absolute_url(self):
+        return reverse('mediacenter:document_detail', kwargs={'pk': self.pk})
+
+    def save(self, *args, **kwargs):
+        self.set_type()
+        super(Document, self).save(*args, **kwargs)
+
+    def set_type(self):
```
twidi: I think a docstring is needed

mediacenter/models.py
```diff
@@ -0,0 +1,78 @@
...
+        return self.filter(type=Document.PDF)
+
+    def text(self):
+        return self.filter(type=Document.TEXT)
+
+    def audio(self):
+        return self.filter(type=Document.AUDIO)
+
+
+class Document(SearchMixin, TimeStampedModel, models.Model):
```
twidi: (I know it's not related to this PR, but there is some inconsistency in the names, here: `TimeStampedModel` is a mixin or a model.Model subclass ?)

mediacenter/models.py
```diff
@@ -0,0 +1,78 @@
...
+    OTHER = 'other'
+
+    TYPE_CHOICES = (
+        (IMAGE, _('image')),
+        (AUDIO, _('sound')),
+        (VIDEO, _('video')),
+        (PDF, _('pdf')),
+        (TEXT, _('text')),
+        (OTHER, _('other')),
+    )
```
twidi: why not django-extended-choices ? :)

mediacenter/templates/mediacenter/document_confirm_delete.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'one-row.html' %}
+{% load i18n %}
+
+{% block row %}
+    <h2>{% blocktrans %}Are you sure you want to delete document "{{ document }}"?{% endblocktrans %}</h2>
```
twidi: I always prefer to use `document.title` than to refer to the `__unicode__` method, but it's my point of view. Explicit is better than implcit. By using the field, I don't need to go to the model to see what will be displayed

mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'two-third-third.html' %}
+
+{% load i18n static %}
+
+{% block twothird %}
+    <h2><span class="theme read">{{ document.get_type_display }}</span> {{ document }}</h2>
+    {% if document.type == document.IMAGE %}
+        <img src="{{ document.original.url }}" />
+    {% elif document.type == document.VIDEO %}
+        <video controls width="100%">
```
twidi: I guess that you know what you're doing with the video tag :)

mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'two-third-third.html' %}
+
+{% load i18n static %}
+
+{% block twothird %}
+    <h2><span class="theme read">{{ document.get_type_display }}</span> {{ document }}</h2>
+    {% if document.type == document.IMAGE %}
+        <img src="{{ document.original.url }}" />
```
twidi: use `document.title` for the `title` attribute ?

mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -0,0 +1,29 @@
...
+                <img src="{% static 'mediacenter/document.svg' %}" /><br />{% trans 'Download document' %}
+            </a>
+        </div>
+    {% endif %}
+    <div class="text">{{ document.summary }}</div>
+{% endblock twothird %}
+{% block third %}
+    {% if user.is_staff %}
+        <ul class="card tinted admin">
+            <li><a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
```
twidi: instead to rely on long urls, I prefer defining quick methods in the model (`get_update_url`...)... More maintainable and readable, for me.

mediacenter/templates/mediacenter/index.html
```diff
@@ -0,0 +1,19 @@
+{% extends 'two-third-third.html' %}
+{% load i18n %}
+
+{% block twothird %}
+    <h2><span class="theme read">{% trans "read" %}</span> {% trans "Medias Center" %}</h2>
+    <div class="grid document-list">
+        {% for document in document_list  %}
+            {% include "mediacenter/document_card.html" %}
```
twidi: It's ok because you say you don't have a lot of date. But using `include` in a loop is a performance killer. Really. A lot.

ideasbox/templates/ideasbox/includes/document_card.html
yohanboniface: Good point :)

mediacenter/templatetags/mediacenter_tags.py
```diff
@@ -0,0 +1,14 @@
+from django import template
+from django.contrib.staticfiles.templatetags.staticfiles import static
+
+register = template.Library()
+
+
+@register.filter()
+def preview_url(inst):
```
twidi: Here a docstring may be cool too, i don't know

mediacenter/tests/test_managers.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+
+from ..models import Document
+
+pytestmark = pytest.mark.django_db
+
+
+def test_video_should_return_only_video(video, image, audio):
```
twidi: why not add pdf and text here and test on them too ? (same for all tests in this file, using all doc types)

mediacenter/models.py
yohanboniface: Both :)
I mean, it does inherit from models.Model because if define some fields, but it's clearly used as a mixin. Do you think name your be `TimeStampMixin`, or something like that?

mediacenter/models.py
yohanboniface: I'm trying to avoid dependencies if I'm not really sure we really need it: this django app will be deployed in a embedded scenario, plus very often with close to no possibility to debug it in case of problem, so I'm really trying to keep the stack under control as much as possible. :)

mediacenter/models.py
yohanboniface: Yeah, I thought about that, but the devil in me won. What about `kind`? :)

mediacenter/tests/test_search.py
```diff
@@ -0,0 +1,24 @@
...
+
+from ..models import Document
+
+pytestmark = pytest.mark.django_db
+
+
+def test_document_are_indexed(video):
+    assert Search.objects.count() == 1
+    assert len(list(Search.search(public=True))) == 1
+    assert Document.objects.search("Ikinyugunyugu").count() == 0
```
twidi: Here I think you just test the pytest fixtures. Pytest is already tested, you are sure that there is no data, except ones added by your fixtures.

twidi: this is the same for every test where your start by checking that there is something at the start of the function

mediacenter/utils.py
```diff
@@ -0,0 +1,12 @@
+import mimetypes
+
+
+def guess_type(filepath):
```
twidi: here you need a docstring

mediacenter/templates/mediacenter/document_detail.html
yohanboniface: Yep, one good point of the project: we now which browsers will connect to the server, only FF on desktop, and Webkit default on the tablets, so only modern ones :)

mediacenter/templates/mediacenter/index.html
yohanboniface: Humm, so I would you to instead?

mediacenter/views.py
```diff
@@ -0,0 +1,38 @@
...
+from django.core.urlresolvers import reverse_lazy
+from django.views.generic import (CreateView, DeleteView, DetailView, ListView,
+                                  UpdateView)
+
+from .models import Document
+
+
+class Index(ListView):
+    model = Document
+    template_name = 'mediacenter/index.html'
```
twidi: everywhere else you rely on the name computed by the django generic views, why not here ?

mediacenter/tests/test_managers.py
yohanboniface: Humm, I just thought three was enough to the purpose of the test :)

mediacenter/models.py
twidi: If it inherites from model.Model, you don't need to use model.Model when defined your model here. Your choice, but either you use it as a model subclass, or a mixin, but the name and usage should be consistent

mediacenter/models.py
twidi: ok understood

mediacenter/models.py
twidi: would be better

mediacenter/templates/mediacenter/document_detail.html
twidi: great

mediacenter/templates/mediacenter/index.html
twidi: There is no good answer to this point. Everybody should know that, that's all. And if it's really a big loop, use external solutions, as caching, pre-renderer (django-template-preprocessor)...

mediacenter/views.py
yohanboniface: No really strong argument actually, I just thought this page was more about a section home page than a list of objects, so that the template will ends with more elements than just a list of object we would expect from a "document_list.html" template.

mediacenter/tests/test_search.py
yohanboniface: That's true, but, as someone said somewhere in this review, "explicit is better than implicit" ;)

What I mean is that I find both clearer and stronger to make sure the context is as expected before doing the actual "unit" test.

I'll not fight to convince that's a good practice, but I think when in doubt, better more tests than not enough :)

mediacenter/templates/mediacenter/index.html
yohanboniface: OK, now I know it. :)

mediacenter/tests/test_search.py
twidi: Ok as you want, you're the boss. But I don't agree. It's different, where is the limit of what you know that is already done by the framework and you don't test, and the parts that you test ? I won't argue ;)

mediacenter/views.py
twidi: ok i see

mediacenter/tests/test_managers.py
twidi: why not two only in this case ? one good, one bad :)

ideasbox/templates/ideasbox/includes/header.html
```diff
@@ -5,7 +5,7 @@
       --><a class="mwide tthird" href="{% url 'blog:index' %}">{% trans "blog" %}</a><!--
       --><a class="mwide tthird" href="{% url 'library:index' %}">{% trans "library" %}</a><!--
       --><a class="mwide tthird" href="#">Contents</a><!--
-      --><a href="#" class="mwide tthird">Medias</a><!--
+      --><a class="mwide tthird" href="{% url 'mediacenter:index' %}">Medias</a><!--
```
davidbgk: `{% spaceless %}` again?

mediacenter/views.py
```diff
@@ -0,0 +1,38 @@
...
+    paginate_by = 10
+index = Index.as_view()
+
+
+class DocumentDetail(DetailView):
```
davidbgk: If you don't plan to add anything to these classes, you might find shorter to pass the model directly to the `as_view()` method within the URLs. Does it beak the MVC pattern? Sure, but it's already broken anyway ;)

yohanboniface: Oh, I was not aware of that option. But I'm not sure moving views definition to the `urls.py` file passes the magopian test ;) @magopian, thoughts? :)

ideasbox/templates/ideasbox/includes/header.html
yohanboniface: Right (again) :)

mediacenter/tests/test_managers.py
yohanboniface: It's one good, one bad, and one for the entropy ;)

yohanboniface:
@twidi @davidbgk Any remaining point I need to fix to get a green light for merge on this? :)

mediacenter/templates/mediacenter/document_confirm_delete.html
twidi: (to resume: explicit is better than implicit ;))

mediacenter/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+    ['pouet.jpg', 'image'],
+    ['pouet.jpeg', 'image'],
+    ['pouet.png', 'image'],
+    ['pouet.mp4', 'video'],
+    ['pouet.avi', 'video'],
+    ['pouet.mp3', 'audio'],
+    ['pouet.ogg', 'audio'],
+    ['pouet.pdf', 'pdf'],
+    ['pouet', None],
+])
```
twidi: missing a test with an unknown extension (which is different than no extension at all)

mediacenter/utils.py
```diff
@@ -0,0 +1,14 @@
...
+    """Return a document kind (image, audio...) guessed from a filename. If
+    no kind can be guessed, returns None."""
+    lookups = ['image', 'video', 'audio', 'text', 'pdf']
+    if filepath:
+        kind, encoding = mimetypes.guess_type(filepath)
+        if not kind:
+            return
+        for lookup in lookups:
+            if lookup in kind:
+                return lookup
```
twidi: I didn't do a lot of file uploading in django, so i don't know, but don't we have a mime type when the user upload its document ?

yohanboniface: Right! But we are doing our guess_type at save time, which doesn't means we are on the upload file process. The fact is that later we'll have command line for batch import of documents, which then will not come from a Form. Do that makes sense? :)

twidi: more or less, you could still guess it "better" when the user upload it :)

As you want, it's not blocker for me

twidi:
@yohanboniface: Good for me

yohanboniface:
Thanks a lot @twidi and @davidbgk! :)

PR #68, opened 2015-01-21
First shot in notice import (cf #8 #9)

yohanboniface:
Ref: #8 #9 

Initial plan was to support  [UniMARC](http://fr.wikipedia.org/wiki/Machine-Readable_Cataloging) standard and [Z39.50](http://fr.wikipedia.org/wiki/Z39.50) protocol. I've looked at them a bit more, and it appears that, despite the name (MAchine Readable Cataloging), both are surviving from XIXe century, and we should absolutely avoid them.

For now, I've coded two identified scenarios: notice import from Moccam en ligne, and automatic import with an ISBN.
For the first, I've chosen the CSV format, that appears to be the only usable choice.
For the second, I'm relying on OpenLibrary, which, even being not perfect (missing many summaries), is the only API freely available, without licence constraint, and with a simple REST procotol.

Let's build with this for now and see if we need a more complex case.

There is one coming for sure, but not in the first iteration, which will be exporting notices from old servers and reimporting. But we can just have a SQL one shot script.

This PR also starts using the Django messages app.


library/forms.py
```diff
@@ -14,9 +17,52 @@ class Meta:
...
+    FORMATS = (
+        (MOCCAM_CSV, 'CSV from "Mocam-en-ligne"'),
+    )
+
+    from_files = forms.FileField(required=False)
+    files_format = forms.ChoiceField(choices=FORMATS)
+    from_isbn = forms.CharField(widget=forms.Textarea, required=False)
+
+    def save_from_files(self):
+        """Save book from given files."""
```
magopian: `Create or update books from given files`?

library/forms.py
```diff
@@ -14,9 +17,52 @@ class Meta:
...
+        for f in files:
+            for notice in handler(f):
+                notice['section'] = Book.OTHER
+                book, _ = Book.objects.update_or_create(isbn=notice['isbn'],
+                                                        defaults=notice)
+                books.append(book)
+        return books
+
+    def save_from_isbn(self):
+        """Retrieve books notices from isbn, and create them."""
```
magopian: `Create or update books from isbn.`?

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+import urllib2
+
+from django.core.files.base import ContentFile
+from django.utils.translation import ugettext as _
+
+OPENLIBRARY_API_URL = 'https://openlibrary.org/api/books?'
+
+
+def to_unicode(text):
```
magopian: I believe `django.utils.encoding.smart_text` does [exactly that](https://docs.djangoproject.com/en/1.7/ref/unicode/#conversion-functions), don't you think?

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    """Do its best to return an unicode string."""
+    if isinstance(text, unicode):
+        return text
+    else:
+        try:
+            text = text.decode('utf-8')
+        except UnicodeDecodeError:
+            text = text.decode('latin-1')
```
magopian: I don't understand: if it can't decode from `utf-8` it sure won't be able to decode from `latin-1` (which is a subset of `utf-8`).

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+def fetch_from_openlibrary(isbn):
+    """Fetch notice from Open Library and return a ready to use dict.
+    Doc: https://openlibrary.org/dev/docs/api/books."""
+    args = {
+        'jscmd': 'data',
+        'format': 'json',
+    }
+    key = 'ISBN:{isbn}'.format(isbn=isbn)
+    args['bibkeys'] = key
+    args = '&'.join('{0}={1}'.format(k, v) for k, v in args.iteritems())
```
magopian: I'd rather use [urllib.urlencode](https://docs.python.org/2/library/urllib.html#urllib.urlencode)

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+        cover = None
+    publishers = data.get('publishers', [])
+    if publishers:
+        publisher = publishers[0]['name']
+    notice = {
+        'isbn': isbn,
+        'title': data.get('title'),
+        'authors': ', '.join([a['name'] for a in data.get('authors', [])]),
+        'cover': cover,
+        'publisher': publisher
```
magopian: Careful here, if there's no `publishers` line 48, then `publisher` won't be initialized.

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+def load_from_moccam_csv(content):
+    """Handle Moccam CSV import.
+    See http://www.moccam-en-ligne.fr
+    """
+    FIELDS = [
+        'isbn', 'title', 'authors', 'publisher', 'collection', 'year', 'price',
+        'summary', 'small_cover', 'cover'
+    ]
+    content = content.split('\n')
+    if not content or content[0].count('	') != 9:
```
magopian: Maybe add a comment here explaining that it's a tab-separated csv file, and thus you want to make sure there's the correct number of columns?

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+        raise ValueError(_('Unable to parse file'))
```
magopian: `raise ValueError(_('Empty content, or badly formatted file'))`?

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    rows = csv.DictReader(content, fieldnames=FIELDS,
+                          delimiter='	')
```
magopian: Just wondering if it would be possible to use `\t` for the delimiter value, instead of a real tab? It would be easier to read and understand (for example, it wouldn't have a different size when reviewing in github: compare with line 85 above.

magopian: Also, instead of manually checking the content line 84 and 85, why not simply initialize `rows` like you did on line 87, and use the `restkey` and `restval` (check https://docs.python.org/2/library/csv.html?highlight=csv.dictreader#csv.DictReader).

Eg: if `rows.keys()` has the `restkey`, or if the last col has the `restval`, then you know the csv isn't correct.

library/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    for row in rows:
+        if row['cover']:
+            cover = load_cover_from_url(row['cover'])
+        else:
+            cover = None
+        authors = row.get('authors', '')
+        if ',' in authors:
+            authors = authors.split(', ')
+            authors.reverse()
```
magopian: Maybe add a comment to explain why you reverse the order?

library/views.py
```diff
@@ -44,6 +46,36 @@ class BookDelete(DeleteView):
...
+
+    def form_valid(self, form):
+        if form.cleaned_data['from_files']:
+            handler = form.save_from_files
+        elif form.cleaned_data['from_isbn']:
+            handler = form.save_from_isbn
+        try:
+            notices = handler()
+        except ValueError:
+            msg = _('Unable to process notices.')
```
magopian: why use a variable `msg` here? you're not using it anymore later

library/views.py
```diff
@@ -44,6 +46,36 @@ class BookDelete(DeleteView):
...
+            messages.add_message(self.request, messages.ERROR, msg)
+        else:
+            if notices:
+                msg = _('Successfully processed {count} notices.')
+                msg = msg.format(count=len(notices))
```
magopian: did you split this on two lines to cope with the 79 char limit?

library/views.py
```diff
@@ -44,6 +46,36 @@ class BookDelete(DeleteView):
...
+                messages.add_message(self.request, messages.SUCCESS, msg)
+                if len(notices) == 1:
+                    self.success_url = notices[0].get_absolute_url()
+            else:
+                msg = _('No notice processed')
```
magopian: Same here, no need for an intermediate variable

library/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+from library.utils import (load_from_moccam_csv, to_unicode,
+                           fetch_from_openlibrary)
```
magopian: Alphabetical order?

library/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
...
+        assert len(notices) == 2
+        assert notices[0]['title'] == 'Les Enchanteurs'
+        assert notices[0]['authors'] == 'Romain Gary'
+        assert notices[0]['cover'] == 'xxx'
+        assert notices[0]['publisher'] == 'Gallimard'
+        assert notices[0]['summary'].startswith('Le narrateur')
+
+
+def test_fetch_from_openlibrary(monkeypatch):
+    doc = """{"ISBN:2070379043": {"publishers": [{"name": "Gallimard"}], "identifiers": {"isbn_13": ["9782070379040"], "openlibrary": ["OL8838456M"], "isbn_10": ["2070379043"], "goodreads": ["118988"], "librarything": ["1655982"]}, "weight": "7 ounces", "title": "Les Enchanteurs", "url": "https://openlibrary.org/books/OL8838456M/Les_enchanteurs", "number_of_pages": 373, "cover": {"small": "https://covers.openlibrary.org/b/id/967767-S.jpg", "large": "https://covers.openlibrary.org/b/id/967767-L.jpg", "medium": "https://covers.openlibrary.org/b/id/967767-M.jpg"}, "publish_date": "January 22, 1988", "key": "/books/OL8838456M", "authors": [{"url": "https://openlibrary.org/authors/OL123692A/Romain_Gary", "name": "Romain Gary"}]}}"""
```
magopian: Isn't this line missing a `# noqa` to make flake8 happy?

library/tests/test_views.py
```diff
@@ -174,3 +175,61 @@ def test_it_should_be_possible_to_remove_isbn_from_books(staffapp):
...
+    form['title'] = 'My book title'
+    form['summary'] = 'My book summary'
+    form['section'] = '1'
+    form['isbn'] = '2-7071-2402-8'
+    form.submit().follow()
+    assert Book.objects.get(isbn='2707124028')
+
+
+def test_import_from_isbn(staffapp, monkeypatch):
+    doc = """{"ISBN:2070379043": {"publishers": [{"name": "Gallimard"}], "identifiers": {"isbn_13": ["9782070379040"], "openlibrary": ["OL8838456M"], "isbn_10": ["2070379043"], "goodreads": ["118988"], "librarything": ["1655982"]}, "weight": "7 ounces", "title": "Les Enchanteurs", "url": "https://openlibrary.org/books/OL8838456M/Les_enchanteurs", "number_of_pages": 373, "cover": {"small": "https://covers.openlibrary.org/b/id/967767-S.jpg", "large": "https://covers.openlibrary.org/b/id/967767-L.jpg", "medium": "https://covers.openlibrary.org/b/id/967767-M.jpg"}, "publish_date": "January 22, 1988", "key": "/books/OL8838456M", "authors": [{"url": "https://openlibrary.org/authors/OL123692A/Romain_Gary", "name": "Romain Gary"}]}}"""
```
magopian: Same here, missing `# noqa`?

library/tests/test_views.py
```diff
@@ -174,3 +175,61 @@ def test_it_should_be_possible_to_remove_isbn_from_books(staffapp):
...
+    form = staffapp.get(reverse('library:book_import')).form
+    form['from_files'] = Upload('library/tests/data/moccam.csv')
+    response = form.submit()
+    response.follow()
+    assert Book.objects.count() == 2
+
+
+def test_import_from_files_does_not_duplicate(staffapp, monkeypatch):
+    monkeypatch.setattr('library.utils.read_url', lambda x: None)
+    BookFactory(isbn='9782070379040')  # Same as first entry of CSV.
```
magopian: Maybe assert that here, so that if you change the csv later, without remembering about this, the test will still fail?

library/tests/test_views.py
```diff
@@ -174,3 +175,61 @@ def test_it_should_be_possible_to_remove_isbn_from_books(staffapp):
...
+    monkeypatch.setattr(
+        'library.utils.read_url',
+        lambda x: open('ideasbox/tests/data/the-prophet.jpg').read()
+    )
+    form = staffapp.get(reverse('library:book_import')).form
+    form['from_files'] = Upload('library/tests/data/moccam.csv')
+    response = form.submit()
+    response.follow()
+    assert Book.objects.count() == 2
+    assert Book.objects.last().cover
```
magopian: You specified the cover line 228 right? Why not assert that the content is coherent here?

magopian:
Excellent work!

r+wc

library/utils.py
yohanboniface: Humm, I'm not sure. It seems we need to specify the encoding, which in our case we don't know, so we need to sort of guess it.

library/utils.py
yohanboniface: Not sure about that: try this in a console ```u'é'.encode('latin1').decode('utf-8')``` it will fail.

library/utils.py
yohanboniface: Good point! :)

library/utils.py
yohanboniface: Wow, good catch! :)

library/utils.py
yohanboniface: Nice idea, I'll do that. :)

library/views.py
yohanboniface: yes :p

library/utils.py
magopian: My bad: latin-1 is a subset of unicode, not utf-8.

library/utils.py
magopian: I understand, fair point.

yohanboniface:
Thanks again for the review @magopian :)

PR #69, opened 2015-01-25
Backup and restore management (fix #13)

### LINKED ISSUE #13: backup / restore
We should be able to totally backup / restore the server data.

Backup means:
- zip all the storage folder, and save it with the box id and date in the "backup" folder

Restore means:
- list the available backups in the backup folder and ask for one to use for backup
- optionally allow to upload a new zip

yohanboniface:
Ref #13 

The idea is that we have a "backuped" folder, where we store everything that we want to backup, for example the database or the user medias, but this will also the place to store some other apps data (example: KA-lite SQLite database).
Backups are just a zip of this folder, and restoring means replacing the current folder by the unzipped files of a backup.


ideasbox/__init__.py
```diff
@@ -0,0 +1,3 @@
+VERSION = (0, 0, 0)
```
davidbgk: I would start with `0.0.1` at least ;)

serveradmin/backup.py
```diff
@@ -0,0 +1,96 @@
...
+        settings.IDEASBOX_ID,
+        __version__,
+        datetime.now().strftime(Backup.DATE_FORMAT)
+    ]))
+
+
+class Backup(object):
+
+    FORMAT = 'zip'
+    DATE_FORMAT = "%Y%m%d%H%M"
```
davidbgk: Why not an iso/standard format by default?

serveradmin/backup.py
```diff
@@ -0,0 +1,96 @@
...
+    ROOT = os.path.join(settings.STORAGE_ROOT, 'backups')
+
+    def __init__(self, name):
+        if not name.endswith('.zip'):
+            raise ValueError
```
davidbgk: You should raise a specific error here.

serveradmin/templates/serveradmin/backup.html
```diff
@@ -0,0 +1,33 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans "Manage server backups" %}</h2>
+<form name="backups" method="POST" enctype="multipart/form-data">
```
davidbgk: Why uppercased method?

serveradmin/views.py
```diff
@@ -42,3 +46,54 @@ def power(request):
...
+                    messages.add_message(request, messages.ERROR,
+                                         _('Unable to load file.'))
+                else:
+                    msg = _('File {name} has been loaded.').format(
+                        name=backup.name)
+                    messages.add_message(request, messages.SUCCESS, msg)
+            else:
+                messages.add_message(request, messages.ERROR,
+                                     _('No file found to upload.'))
+        elif 'backup' in request.POST:
```
davidbgk: `do_backup` vs. `backup` is confusing. Why don't you use regular forms btw?

serveradmin/backup.py
yohanboniface: Humm, my rationale was to go on the simpler (no space, comma, etc.) while human readable format.

Are you thinking about one specific?

serveradmin/templates/serveradmin/backup.html
yohanboniface: Recollection from 90's ;)
(Fixing.)

serveradmin/views.py
yohanboniface: All the `do_` named fields are the submit buttons, which are not anyway shipped with the Django forms.

I've chosen not to use Django forms because at the end the form is a lot of submit buttons (5 actually), which are not part of Django forms, one field that would require a custom field, because it needs Backup instance to create the HTML (a table) but on the form level it only deals with backup names, and a file field. So I made a little equation and found a Django form would mean more complexity and boostrap code.

Back to the initial question, is `do_create` a bit more clear? As we are always dealing with backup, I understand your point that having 'backup' in an action is a bit confusing.

serveradmin/backup.py
davidbgk: [isoformat](https://docs.python.org/2/library/datetime.html#datetime.datetime.isoformat) looks good to me. I know that there are some issues with `:` within filenames on Windows though.

serveradmin/views.py
davidbgk: I'm always a bit scarred when not using Django's forms and doing that kind of things `return os.remove(self.path)` without any check ;)

serveradmin/backup.py
yohanboniface: Yeah, my idea is to limit as much as possible human based errors on those files, that will be transferred by email, USB key, etc., so I think avoiding spaces or any non numeric char is stronger. Make sense?

davidbgk: My concern is that I find `20150127223034` hardly readable. Something like `2015-01-27T22:30:34` looks more readable at first sight to me. Other option: just keep the iso date + counter `2015-01-27-001` for instance.

yohanboniface: I've thought about it since then, because I think your readability point is a real one. But then I ended to the conclusion that those file names are not made to be *read* by humans, but made to be *readable* by humans, and at the end, as soon as they are *readable*, the important point is robustness across many unknown workflows.

So, as for now, I'm going this way. We still have one month before release, so we have a bit of time for a better idea to rise up :)

yohanboniface:
Inch Allah!

Thanks for the review, amigo :)

PR #70, opened 2015-01-26
Add lang switcher in footer (fix #56)

### LINKED ISSUE #56: We need a lang switcher
Certainly a (nice) dropdown menu in the header and/or in the footer. We also need to display which language is active.

yohanboniface:
Ref: #56

This PR just adds a select lang switcher in the footer, but there are two tricks:
- firstly, as it adds a form in every page, I've needed to add ids to every form, in order to make them selectable in unittests, which create a lot of noise in the diff
- secondly, as we are using i18n URLs, we needed to workaround a limitation of the Django view: if we just give the current URL as "next" parameter, it will never work, because the URL contains the i18n prefix (for example `/en/`, so the Django view will set the new lang, then redirect to the next URL, which contains the i19n prefix, so will be caught by the i18n middleware, which at its turn will set the current language (so using the initial language)

### DIFF
--- a/ideasbox/static/ideasbox/main.css
+++ b/ideasbox/static/ideasbox/main.css
@@ -313,7 +313,7 @@ html[dir='rtl'] .boxid {
     float: left;
     right: -31px;
 }
-div.i18n_switch select {
+.i18n_switch select {
     display: inline;
     width: calc(100% - 100px);
     margin: 0;
@@ -322,7 +322,7 @@ div.i18n_switch select {
     vertical-align: top;
     height: 40px;
 }
-div.i18n_switch input[type='submit'] {
+.i18n_switch input[type='submit'] {
     display: inline;
     width: 100px;
     border-bottom-left-radius: 0;
--- a/ideasbox/static/ideasbox/main.css
+++ b/ideasbox/static/ideasbox/main.css
@@ -313,7 +313,33 @@ html[dir='rtl'] .boxid {
     float: left;
     right: -31px;
 }
-
+div.i18n_switch select {
+    display: inline;
+    width: calc(100% - 100px);
+    margin: 0;
+    border-bottom-right-radius: 0;
+    border-top-right-radius: 0;
+    vertical-align: top;
+    height: 40px;
+}
+div.i18n_switch input[type='submit'] {
+    display: inline;
+    width: 100px;
+    border-bottom-left-radius: 0;
+    border-top-left-radius: 0;
+    margin: 0;
+    vertical-align: top;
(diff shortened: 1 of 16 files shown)


ideasbox/static/ideasbox/main.css
```diff
@@ -313,7 +313,33 @@ html[dir='rtl'] .boxid {
     float: left;
     right: -31px;
 }
-
+div.i18n_switch select {
```
Exirel: Better not to use `elem.class child` but `.class child` instead, as explained in [Don’t qualify class rules with tag names](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Writing_efficient_CSS#Don.27t_qualify_class-categorized_rules_with_tag_names).

So I would write: `.i18n_switch select` here.

yohanboniface:
Preview (the form is on the right of the footer ­— when in ltr, otherwise in the right ;) ): 
screenshot from 2015-01-31 17 56 10

ideasbox/templates/ideasbox/form_confirm_delete.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'one-row.html' %}
+{% load i18n %}
+
+{% block row %}
+    <h2>{% block confirm_message %}{% endblock %}</h2>
+    <p>{% block confirm_warning %}{% endblock %}</p>
+    <form method="POST" action="." id="delete_form">
+        {% csrf_token %}
+        {{ form.as_p }}
+        <input type="submit" name="confirm" value="{% trans 'yes' %}" />
```
Exirel: This means all your fields will be inside a `<p>` elements, except for the submit button? For consistency, I tend to always put `<input>` element inside a `<p>` one.

But that's nitpicking. :smile:

ideasbox/templatetags/ideasbox_tags.py
```diff
@@ -21,3 +24,10 @@ def theme_slug(inst, slug=None):
         slug = SLUGS.get(name, name.lower())
     klass = THEMES.get(name, slug)
     return mark_safe(tpl.format(klass=klass, slug=slug))
+
+
+@register.filter()
+def remove_i18n(url):
+    """Remove i18n prefix from an URL."""
+    return i18n_pattern.sub("/", url)
+i18n_pattern = re.compile(language_code_prefix_re)
```
Exirel: My PEP8 sense tells me to always put 2 blank lines between any module-level declarations (classes, functions, variables).

yohanboniface:
Thanks for the review @Exirel :)

Issue #32, opened 2014-11-21
debian package


yohanboniface:
We want to be able to install the project from with `apt-get install` in our server, so we'll need a package.

Things to take care of:
- [x] templates
- [x] static files
- [ ] user uploaded content
- [x] nginx vhost activation (or other server strategy)
- [x] i18n
- [x] migrations
- [x] service like daemon to run django (with gunicorn?)
- [x] create a first user at first run
- [x] autostart daemon at server boot

yohanboniface:
A bit of doc:
- https://wiki.debian.org/DjangoPackagingDraft
- http://www.laurentluce.com/posts/hello-world/

yohanboniface:
Examples:
- http://bazaar.launchpad.net/~zkrynicki/+junk/django_hello/files
- http://bazaar.launchpad.net/~zkrynicki/ubuntu/lucid/django-testproject/packaging/files

yohanboniface:
- http://stackoverflow.com/questions/4452208/package-a-django-project-and-its-dependencies-for-a-standalone-product
- http://www.radix50.net/~ibr/notes/20080323-0-django-autotools-debian.html

yohanboniface:
https://hynek.me/articles/python-app-deployment-with-native-packages/

yohanboniface:
Parcel is a suite of classes and helper functions designed to be used in conjunction with Fabric for building and deploying Python web applications using native packages.

http://parcel.readthedocs.org/en/latest/user/quickstart.html

yohanboniface:
https://hynek.me/articles/virtualenv-lives/

xordoquy:
https://github.com/spotify/dh-virtualenv

yohanboniface:
Example of Nginx postinst scripts:
- https://github.com/quarkslab/irma-frontend/blob/master/debian/irma-frontend-nginx.postinst
- https://github.com/codebutler/ubuntu-packages/blob/master/nginx/debian/postinst

yohanboniface:
Example using dh_virtualenv: https://github.com/goodguy/debian-virtualenv-skeleton

yohanboniface:
May be interesting to publish packages, as we can't publish dh_virtualenv made packages from launchpad: https://packagecloud.io/

yohanboniface:
For the record: https://nylas.com/blog/packaging-deploying-python

yohanboniface:
https://github.com/nylas/make-deb

PR #74, opened 2015-02-05
Debian

yohanboniface:
Ref #32 

I've chosen to use [dh_virtualenv](http://dh-virtualenv.readthedocs.org/en/0.8/), which allows to package all the deps like in a virtualenv, Python included. This sounds the nicer method, specifically because doing so we can add more dependencies even if they are not already Debian packaged.

This PR includes:
- debian files
- a service init script (ideasbox.init)
- a /etc/default/ideasbox file (ideasbox.default)
- an nginx VHost (nginx/ideasbox)

Also, I've chosen to move all the applications inside `ideasbox` namespace, to avoid import clash in the future (i.e. not to have "search" or "blog" as first level module). This create a bit of noice in the diff of the PR sadly.

The most important things that need to be reviewed are:
- the debian package files (control, postinst...)
- the service init file (ideasbox.init)

I need to solve one last thing (at least): there is no user created on the installation, and we need one at first install (but only first install).


debian/changelog
```diff
@@ -0,0 +1,5 @@
+ideasbox (0.0.1) unstable; urgency=low
```
dzen: ideasbox (0.0.1-1) unstable; urgency=low

-1 parce que c'est le premier paquet debian que tu as fait. Si il est foiré pour cette meme version, tu referas une version de ton paquet avec -2 (en utilisant dch)

yohanboniface: Ah, yes, good point :)

PR #79, opened 2015-02-19
Added python-pytest to build-depends.

chrisglass:
This change allows to build cleanly on an Ubuntu trusty PPA (trusty has dh-virtualenv 0.6.1)


yohanboniface:
Thanks! :)

PR #138, opened 2015-05-06
Fix service getter clashing with status key

yohanboniface:
When defining a service in settings, it's possible to override the way to interact the service, this was clashing with the `status` key.


ametaireau:
go go go!

Issue #149, opened 2015-07-14
CSS bugs in iPad Chrome


yohanboniface:
Sylvain is now in Australia in a mission where they use iPad (yeah I know…).
He reports that CSS is broken on Chrome (we can't even use Safari, as it does not support flex box):
![ipad3](https://cloud.githubusercontent.com/assets/146023/8668755/ab75b3a2-2a0e-11e5-8d0d-1bc0a10e4e63.jpg)
![ipad2](https://cloud.githubusercontent.com/assets/146023/8668757/ab7b021c-2a0e-11e5-83be-75e720cdaa60.jpg)
![ipad](https://cloud.githubusercontent.com/assets/146023/8668756/ab788cf8-2a0e-11e5-9523-2d674df64a1f.jpg)

When I test it with what I think is the size of the iPad (1024x768), I don't reproduce.

![screenshot from 2015-07-14 09 55 27](https://cloud.githubusercontent.com/assets/146023/8668791/0abf4d3c-2a0f-11e5-90ad-a2e593ca816e.png)

@Dadouw do we have an iPad at BSF office?
@fheslouin @providenz @magopian @davidbgk anyone having an iPad to hand able to have a quick debug on this?

providenz:
I'll check this tomorrow.

yohanboniface:
Cool, thanks Laurent! :)
Let's see also if we have one in the office just in case.

yohanboniface:
Should not be an issue with flexbox itself: http://caniuse.com/#feat=flexbox

providenz:
Css bug confirmed.

yohanboniface:
OK, thanks for checking!
Have you been able to track down to the guilty CSS part and/or do you have (a sort of) an idea of fix?

providenz:
I'll try to find out that. I've read tha ios chrome is supposed to use the safari rendering engine, so it might be related to flexbox support.

yohanboniface:
Aïe :(

yohanboniface:
Maybe we just need something like `-webkit` prefix or something like this?

yohanboniface:
ah yes, that's what says http://caniuse.com/#feat=flexbox about iOS 8.3, I've read a bit fast the other day, just looking at the version 9.

yohanboniface:
@providenz around to make a try with `-webkit-` prefix added to flexbox related rules? I'd like to push a new version to the Australian box while we have someone there to assist. :)

[...]

PR #155, opened 2015-07-17
Fix ios chrome layout issues

providenz:
On ipad chrome rendering engine does not support flexbox and needs proper prefixing.

img_0013


yohanboniface:
Excellent, thanks a lot!
Fix #149

yohanboniface:
New build has been sent to Australia and installed, and the bug is confirmed fixed, thanks again! :)

PR #159, opened 2015-08-24
Wifi

Erf-:
Hi Yohan. Here is the integration of the wifi library I have done. It looks like very perfectible in terms of performance, especially about connection issues. Your comments are welcomed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 language: python
 python:
   - "2.7"
-sudo: false
+before_install:
+    - pip install git+https://github.com/Erf-/wifi.git@adding-sudo-args#egg=wifi
+sudo: required
 install:
+    - sudo apt-get install wireless-tools
     - make devinstall
 script: make test
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 python:
   - "2.7"
 before_install:
-    - pip install git+https://github.com/Erf-/wifi.git@adding-sudo-args#egg=wifi
+    - pip install git+https://github.com/Erf-/wifi.git@connection-status#egg=wifi
 sudo: required
 install:
     - sudo apt-get install wireless-tools
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,8 @@ python:
   - "2.7"
 before_install:
     - pip install git+https://github.com/Erf-/wifi.git@connection-status#egg=wifi
-sudo: required
+sudo: false
 install:
-    - sudo apt-get install wireless-tools
     - make devinstall
 script: make test
 notifications:
--- a/ideasbox/serveradmin/tests/conftest.py
+++ b/ideasbox/serveradmin/tests/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+from django.utils.translation import ugettext as _
 
 from ..backup import Backup
 
@@ -7,9 +8,6 @@ try:
     from io import StringIO
 except ImportError:  # Python < 3
     from StringIO import StringIO
-from mock import patch
-from django.utils.translation import ugettext as _
-from wifi.utils import get_property, set_properties
 
 
 @pytest.fixture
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: python
 python:
   - "2.7"
-sudo: false
+before_install:
+    - pip install git+https://github.com/Erf-/wifi.git@connection-status#egg=wifi
+    - sudo chmod a+r+w /etc/network/interfaces
```
yohanboniface: Tests should not touch this file for real, is there some place we are not mocking?

.travis.yml
```diff
@@ -1,8 +1,12 @@
...
+#sudo: required
 install:
+    - sudo apt-get install wireless-tools
```
yohanboniface: Same here, we should not call this system lib from tests for real IMHO.

ideasbox/serveradmin/tests/conftest.py
```diff
@@ -10,3 +17,173 @@ def backup(monkeypatch):
...
+        self.truncate = True
+        self.seek(0)
+
+@pytest.fixture
+def properties_file():
+    properties_file_content = """scheme_current=test-scheme
+    interface_current=test-interface
+    scheme_active=True
+    """
+    content = properties_file_content
```
yohanboniface: humm?

ideasbox/serveradmin/views.py
```diff
@@ -11,6 +14,7 @@
 from .backup import Backup
 from .utils import call_service
 
+interface = 'wlan0'
```
yohanboniface: Needs to be in the config file (/etc/default/ideasbox, then an env var, because it's the system to decide this).

ideasbox/serveradmin/views.py
```diff
@@ -110,3 +114,81 @@ def backup(request):
...
                   {'batteries': batinfo.batteries()})
+
+
+unset_config_data = """scheme_current=None
+interface_current={interface}
+scheme_active=False
+""".format(interface=interface)
+
+def set_wifilist():
+    f = open('.runningconfig', 'r+')
```
yohanboniface: why do we need this file?

ideasbox/serveradmin/views.py
```diff
@@ -110,3 +114,81 @@ def backup(request):
...
+    read_data = f.read()
+    if len(read_data) == 0:
+        f.write(unset_config_data)
+    print f.read()
+    from wifi.utils import get_property, set_properties
+    to_bool = {'True' : True, 'False' : False, None : False, 'None' : False}
+    # reset scheme_active property
+    essid = get_property('scheme_current')
+    try:
```
yohanboniface: you can replace this by something like `is_essid == 'True'`

ideasbox/serveradmin/views.py
```diff
@@ -110,3 +114,81 @@ def backup(request):
...
+        wifi = Cell.all(interface, sudo=True)
+        # Florian --> config file with interface name
+        for hotspot in wifi:
+            # find out if connected with
+            id_ = '--'.join([hotspot.ssid, hotspot.address])
+            hotspot.is_active = (get_property('scheme_current') == id_)
+            hotspot.is_connected = (hotspot.is_active and is_connected)
+            hotspot.action = action[hotspot.is_connected]
+    except InterfaceError:
```
yohanboniface: only one line should be in a `try` block, with one or more named exception catched.

ideasbox/serveradmin/views.py
```diff
@@ -110,3 +114,81 @@ def backup(request):
...
+                              lambda cell: cell.address == address)[0]
+            cell_kwargs['cell'] = cell
+            scheme = Scheme.for_cell(**cell_kwargs)
+            if not Scheme.find(cell_kwargs['interface'], scheme.name):
+                scheme.save()
+            try:
+                scheme.activate()
+            except ConnectionError:
+                # just to verify in terminal
+                print "erreur"
```
yohanboniface: Clearer and English message, if we really need that :)

ideasbox/serveradmin/views.py
```diff
@@ -110,3 +114,81 @@ def backup(request):
...
+        else:
+            # disconnect
+            import subprocess
```
yohanboniface: why import are not on top of tile?

ideasbox/serveradmin/views.py
```diff
@@ -110,3 +114,81 @@ def backup(request):
...
+            # to avoid errors in tests
+            subprocess.call(["/sbin/ifdown", interface])
+    # refresh wifilist
+    wifilist, auth = set_wifilist()
+    # causes the long load time because of the scan
+    return render(request, 'serveradmin/wifi.html',
+                  {'wifiList': wifilist,
+                   'AuthOK' : auth})
```
yohanboniface: no camel case in python: `auth_ok`
btw, this name is a bit ambiguous: are we saying `is_connected` ?

yohanboniface:
Generic comment: seems we should be testing the wifi on a higher level.
Ideasbox package only deals with `wifi` package, it then assumes that this package is well tested. In other words we should not deal with any system call ourselves.

ideasbox/serveradmin/views.py
Erf-: It is probably justified rigth below. It is in order to avoid some incoherence in tests because depending on the type of import (top file import or within function import) the mock could be unimplemented because python do not recognize the call as the mocked function.

ideasbox/serveradmin/views.py
Erf-: It is implemented in wifi lib to identify the connection status and related config data.

ideasbox/serveradmin/views.py
Erf-: Is it actually the system or rather the admin who make the decision?

ideasbox/serveradmin/views.py
Erf-: I have not understood...

ideasbox/serveradmin/tests/conftest.py
```diff
@@ -3,6 +3,13 @@
 from ..backup import Backup
 
 DATA_ROOT = 'ideasbox/serveradmin/tests/data'
+try:
+    from io import StringIO
+except ImportError:  # Python < 3
+    from StringIO import StringIO
+from mock import patch
```
yohanboniface: Wrong import order, use [isort](https://github.com/timothycrosley/isort) to help you follow the conventions.

ideasbox/serveradmin/tests/conftest.py
```diff
@@ -10,3 +17,172 @@ def backup(monkeypatch):
...
+                    Bit Rates:54 Mb/s
+                    Extra:wpa_ie=dd160050f20101000050f20201000050f20201000050f202
+                    IE: WPA Version 1
+                        Group Cipher : TKIP
+                        Pairwise Ciphers (1) : TKIP
+                        Authentication Suites (1) : PSK
+                    Quality=100/100  Signal level=74/100
+"""
+    output = '\n'.join([bsf01, bsf02])
+    return output
```
yohanboniface: `return '\n'.join()`

ideasbox/serveradmin/tests/conftest.py
```diff
@@ -10,3 +17,172 @@ def backup(monkeypatch):
...
+def disconnected_output():
+    disconnected_output = """test-interface00     IEEE 802.11bgn  ESSID:off/any  
+          Mode:Managed  Access Point: Not-Associated   Tx-Power=20 dBm   
+          Retry short limit:7   RTS thr=2347 B   Fragment thr:off
+          Power Management:off
+"""
+    return disconnected_output
+
+class MyStringIO(StringIO):
+# extends StringIO buit-in class to override write and close methods
```
yohanboniface: why?

ideasbox/serveradmin/tests/test_views.py
```diff
@@ -181,3 +186,88 @@ def test_upload_button_do_not_create_backup_with_bad_file_name(staffapp,
...
+
+def test_wifi_connection_to_hotspot(monkeypatch,
+                                    staffapp,
+                                    success_output,
+                                    failure_output,
+                                    list_output):
+    def save_mock():
+        pass
+
+    monkeypatch.setattr("wifi.scheme.Scheme.save",
```
yohanboniface: why not just "save_mock" as temp attribute?

fheslouin:
How far did we go ? Can we call the wifi library in the IdeasBox code ? 
I should be able to manage the front-end UI part if everything else is in order

Erf-:
At this stage we can actually connect and disconnect to available hotspots but I am waiting for @yohanboniface to review the current code for improvements. Then I have to send you a progress status. I have not got the time to do it but I will try to do it by the end of the week.

yohanboniface:
Superseded by #197

PR #163, opened 2015-09-29
Add conf file for the KoomBook

fheslouin:
Variables have been adapted for the context. 
STORAGE_ROOT is pointing to the Hard Drive : /media/hdd/ideasbox/storage


yohanboniface:
You should make it much shorter (only setting the value you need to override), take example on [azraq.py](https://github.com/ideas-box/ideasbox.lan/blob/master/ideasbox/conf/azraq.py) one :)

PR #165, opened 2015-10-07
Pop up confirm

parmeso6:
A try for a confirm pop-up before to stop/restart the server.
The JS function is in a seperate file which is probably to be changed, as well as the form a bit convoluted.


yohanboniface:
Can you rebase on master? :)

yohanboniface:
Done with https://github.com/ideastube/ideastube/commit/7d03801b12c7fc1a166f05196cc93fbd4c547f00

PR #168, opened 2015-10-12
Power btn

mariefontanel:
The power button in the header for the admin


yohanboniface:
Can you rebase on master? :)

Issue #132, opened 2015-04-01
Add default image when book has no cover


yohanboniface:

yohanboniface:
Fixed by #175.
We may ask designers for a custom image though.

PR #175, opened 2015-10-19
add default book display

eregnier:
Solve #132 

The default book image can be switched easyly as it lives in the static folder.


yohanboniface:
Thanks!

PR #178, opened 2015-10-19
A begginer may forget this step

Lcaracol:
As a begginer I could have easily forgotten this step.


yohanboniface:
When using virtualenvwrapper, and thus mkvirtualenv command, it's yet done.

We may switch to using virtualenv only maybe, because I've seen issues from people installing the wrapper on ZSH or things like this.

Anyway, as it is in this PR I don't think it's needed, but the overall point needs to be make clearer, I agree :)

yohanboniface:
Maybe for now adding a mention of `workon ideasbox` to reactivate the venv?

PR #179, opened 2015-10-20
Adds two packages (libxml2-dev libxslt1-dev) to dependencies.

Lcaracol:
To execute without errors "make devinstall" I had to install these additionnal packages. Installing 'libxml2-dev' only wasn't enough.

It may be because I'm running an old 12.04 ubuntu desktop.

The log message is here:
https://zerobin.net/?1b0815e147ec6d05#WNN8th4y3K9FUGEHGUBakUo2KqzN5l3tJHzmQXEi25s=

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ Project's typical development environment requires:
 
 On a Debian-based system, you may use:
 
-    sudo apt-get install python-pip python-virtualenv virtualenvwrapper python-dev libjpeg-dev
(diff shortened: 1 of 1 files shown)


yohanboniface:
Sounds legit, thanks :)

Lcaracol:
Note: the dependencies were found on this webpage:
https://stackoverflow.com/questions/6504810/how-to-install-lxml-on-ubuntu

Found after looking for " libxml/xmlversion.h " on a Web Search Engine.

It seems that on some ubuntu distros there could have been some more dependencies, as seen in the comments, but with these two only everything was fine for ubuntu 12.04

The possible additionnal dependencies found in the comments where 
lib32z1-dev and zlib1g-dev

for ubuntu 13.04 and 14.04

PR #196, opened 2015-10-29
tests: Show which lines are not covered

bochecha:
This is a useful indication as to which parts of the code might need
more unit tests.


yohanboniface:
I'm thinking about not running cov at all by default, but let's think about it with more calm ;)

PR #197, opened 2015-10-29
Wifi management

bochecha:
Here is wifi management for the admin interface.

It is still a work in progress, so don't merge it just yet.

However, it does enough that it's worth sharing, if only to enable early review.

Also, note that I'll be rebasing this branch, so if you fetch it locally be prepared to have to force things a bit. :wink: 

What's done and missing:

* [x] List all wifi networks
* [x] Connect to a wifi network (without sudo)
 * [x] Open network
 * [x] WPA-protected network
 * [ ] Other network security types?
* [x] Forget a network
* [ ] Connect to hidden network?

Do we want to support other types of network security? WEP just shouldn't be used any more now, and I'm not sure the various "enterprise" types are very useful for our target. :smiley: 

Do we want to support connecting to a hidden network? (that is, one which does not advertise its SSID)

Is there anything else the admin app should be able to do with Wifi networks?

Screenshots of how it looks like are at: https://bochecha.fedorapeople.org/ideascube/

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -47,7 +47,7 @@ Project's typical development environment requires:
 
 On a Debian-based system, you may use:
 
-    sudo apt-get install python-pip python-virtualenv virtualenvwrapper python-dev libjpeg-dev libxml2-dev libxslt1-dev zlib1g-dev build-essential
+    sudo apt-get install python-pip python-virtualenv virtualenvwrapper python-dev libjpeg-dev libxml2-dev libxslt1-dev zlib1g-dev build-essential libdbus-glib-1-dev
 
 ### Download the souce code
 
--- a/ideascube/serveradmin/templates/serveradmin/wifi.html
+++ b/ideascube/serveradmin/templates/serveradmin/wifi.html
@@ -6,7 +6,7 @@
 {% for wifi in wifi_list %}
 {% if forloop.first %}<ul class="wifi-list">{% endif %}
     <li>
-        <a class="wifi-network" data-ssid="{{ wifi.ssid }}" data-known="{{ wifi.known }}">
+        <a class="wifi-network" data-ssid="{{ wifi.ssid }}" data-known="{{ wifi.known }}" data-secure="{{ wifi.secure }}">
             {{ wifi.ssid }}
             {% if wifi.connected %}
             <span><i class="fa fa-check"></i></span>
--- a/ideascube/serveradmin/templates/serveradmin/wifi.html
+++ b/ideascube/serveradmin/templates/serveradmin/wifi.html
@@ -24,8 +24,23 @@
 <p>{% trans 'No Wi-Fi available' %}</p>
 {% endfor %}
 
+<div id="passphrase-popup" class="overlay">
+    <div class="popup">
+        <h2>{% trans 'Authentication required' %}</h2>
+        <a class="close" href="#">&times;</a>
+        <div class="content">
+            <p>{% trans 'An encryption key is required to access this network.' %}</p>
+            <form method="POST" action="">
+                {% csrf_token %}
+                <input type="password" id="wifi_key" name="wifi_key" autocomplete="off" />
+                <input type="submit" value="{% trans 'Connect' %}" />
+            </form>
+        </div>
+    </div>
+</div>
+
 <script type="text/javascript">
-    ID.initWifiList('.wifi-network', "{% url 'server:wifi' %}");
+    ID.initWifiList('.wifi-network', '#passphrase-popup', "{% url 'server:wifi' %}");
 </script>
 
 {% endblock twothird %}
--- a/ideascube/serveradmin/templates/serveradmin/wifi.html
+++ b/ideascube/serveradmin/templates/serveradmin/wifi.html
@@ -1,0 +1,27 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans "Manage Wi-Fi" %}</h2>
+{% for wifi in wifi_list %}
+{% if forloop.first %}<ul class="wifi-list">{% endif %}
+    <li>
+        <a class="wifi-network">
+            {{ wifi.ssid }}
+            {% if wifi.connected %}
+            <span><i class="fa fa-check"></i></span>
+            {% endif %}
+            <span class="right">{{ wifi.strength }}%</span>
+            {% if wifi.secure %}
+            <span class="right"><i class="fa fa-lock"></i></span>
+            {% endif %}
+        </a>
+    </li>
+{% if forloop.last %}
+</ul>
+{% endif %}
+{% empty %}
+<p>{% trans 'No Wi-Fi available' %}</p>
+{% endfor %}
+
+{% endblock twothird %}
--- a/ideascube/serveradmin/templates/serveradmin/wifi.html
+++ b/ideascube/serveradmin/templates/serveradmin/wifi.html
@@ -6,7 +6,7 @@
 {% for wifi in wifi_list %}
 {% if forloop.first %}<ul class="wifi-list">{% endif %}
     <li>
-        <a class="wifi-network">
+        <a class="wifi-network" data-ssid="{{ wifi.ssid }}" data-known="{{ wifi.known }}">
             {{ wifi.ssid }}
             {% if wifi.connected %}
             <span><i class="fa fa-check"></i></span>
@@ -23,5 +23,9 @@
 {% empty %}
 <p>{% trans 'No Wi-Fi available' %}</p>
 {% endfor %}
+
+<script type="text/javascript">
+    ID.initWifiList('.wifi-network', "{% url 'server:wifi' %}");
+</script>
 
 {% endblock twothird %}
--- a/ideascube/serveradmin/templates/serveradmin/wifi.html
+++ b/ideascube/serveradmin/templates/serveradmin/wifi.html
@@ -19,6 +19,7 @@
     </li>
 {% if forloop.last %}
 </ul>
+<p><a href="{% url 'server:wifi_history' %}">{% trans 'History' %}</a></p>
 {% endif %}
 {% empty %}
 <p>{% trans 'No Wi-Fi available' %}</p>
--- a/ideascube/serveradmin/templates/serveradmin/wifi_history.html
+++ b/ideascube/serveradmin/templates/serveradmin/wifi_history.html
@@ -1,0 +1,28 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans 'Wi-Fi History' %}</h2>
+{% for wifi in wifi_list %}
+{% if forloop.first %}
+<form method="POST" action="{% url 'server:wifi_history' %}" id="wifi_history">
+    {% csrf_token %}
+    <fieldset class="wifi-list" name="ssids">
+{% endif %}
+        <label class="wifi-network" for="{{ wifi.ssid }}">
+            <input type="checkbox" id="{{ wifi.ssid }}" name="{{ wifi.ssid }}" />
+            {{ wifi.ssid }}
+            {% if wifi.connected %}
+            <span><i class="fa fa-check"></i></span>
+            {% endif %}
+        </label>
+{% if forloop.last %}
+    </fieldset>
+    <input type="submit" value="{% trans 'Forget Selected' %}" />
+</form>
+{% endif %}
+{% empty %}
+<p>{% trans 'No known Wi-Fi networks' %}</p>
+{% endfor %}
+
+{% endblock twothird %}
--- a/ideascube/serveradmin/tests/__init__.py
+++ b/ideascube/serveradmin/tests/__init__.py
@@ -1,3 +1,12 @@
+try:
+    from NetworkManager import NM_DEVICE_TYPE_ETHERNET, NM_DEVICE_TYPE_WIFI
+
+except Exception:
+    # Use hard-coded values for the tests, these are constants anyway
+    NM_DEVICE_TYPE_ETHERNET = 1
+    NM_DEVICE_TYPE_WIFI = 2
+
+
 class NMActiveConnection(object):
     def __init__(self, ssid='', is_secure=True):
         self.Connection = NMConnection(True, ssid=ssid, is_secure=is_secure)
@@ -22,3 +31,10 @@ class NMConnection(object):
             settings.update({'802-11-wireless-security': {}})
 
         return settings
+
+
+class NMDevice(object):
+    def __init__(self, is_wifi):
+        self.DeviceType = (
+            NM_DEVICE_TYPE_WIFI if is_wifi else NM_DEVICE_TYPE_ETHERNET)
+        self.Managed = True
--- a/ideascube/serveradmin/tests/__init__.py
+++ b/ideascube/serveradmin/tests/__init__.py
@@ -7,6 +7,14 @@ except Exception:
     NM_DEVICE_TYPE_WIFI = 2
 
 
+class NMAccessPoint(object):
+    def __init__(self, ssid, strength, wpa_flags):
+        self.Ssid = ssid
+        self.Strength = strength
+        self.WpaFlags = wpa_flags
+        self.Frequency = 5000
+
+
 class NMActiveConnection(object):
     def __init__(self, ssid='', is_secure=True):
         self.Connection = NMConnection(True, ssid=ssid, is_secure=is_secure)
@@ -38,3 +46,14 @@ class NMDevice(object):
         self.DeviceType = (
             NM_DEVICE_TYPE_WIFI if is_wifi else NM_DEVICE_TYPE_ETHERNET)
         self.Managed = True
+
+    def SpecificDevice(self):
+        return NMSpecificDevice()
+
+
+class NMSpecificDevice(object):
+    def GetAllAccessPoints(self):
+        return [
+            NMAccessPoint('random open network', 42, 0),
+            NMAccessPoint('', 2, 332),
+            NMAccessPoint('my home network', 99, 332)]
--- a/ideascube/serveradmin/tests/__init__.py
+++ b/ideascube/serveradmin/tests/__init__.py
@@ -1,0 +1,24 @@
+class NMActiveConnection(object):
+    def __init__(self, ssid='', is_secure=True):
+        self.Connection = NMConnection(True, ssid=ssid, is_secure=is_secure)
+
+
+class NMConnection(object):
+    def __init__(self, is_wifi, ssid='', is_secure=True, delete_func=None):
+        self.is_wifi = is_wifi
+        self.is_secure = is_secure
+        self.ssid = ssid
+
+        if delete_func is not None:
+            self.Delete = lambda: delete_func(self)
+
+    def GetSettings(self):
+        if not self.is_wifi:
+            return {}
+
+        settings = {'802-11-wireless': {'ssid': self.ssid}}
+
+        if self.is_secure:
+            settings.update({'802-11-wireless-security': {}})
+
+        return settings
--- a/ideascube/serveradmin/views.py
+++ b/ideascube/serveradmin/views.py
@@ -11,7 +11,8 @@ from ideascube.decorators import staff_member_required
 
 from .backup import Backup
 from .utils import call_service
-from .wifi import AvailableWifiNetwork, enable_wifi, WifiError
+from .wifi import (
+    AvailableWifiNetwork, KnownWifiConnection, enable_wifi, WifiError)
 
 
 @staff_member_required
@@ -141,3 +142,28 @@ def wifi(request, ssid=''):
 
     return render(
         request, 'serveradmin/wifi.html', {'wifi_list': wifi_list.values()})
+
+
+@staff_member_required
+def wifi_history(request):
+    try:
+        enable_wifi()
+        wifi_list = KnownWifiConnection.all()
+
+    except WifiError as e:
+        messages.error(request, e)
+        return render(request, 'serveradmin/wifi_history.html')
+
+    for ssid, checked in request.POST.items():
+        if checked.lower() in ('on', 'true'):
+            try:
+                connection = wifi_list.pop(ssid)
+                connection.forget()
+
+            except KeyError:
+                # Someone tried to forget a connection we don't know.
+                continue
+
+    return render(
+        request, 'serveradmin/wifi_history.html',
+        {'wifi_list': wifi_list.values()})
--- a/ideascube/serveradmin/views.py
+++ b/ideascube/serveradmin/views.py
@@ -115,7 +115,7 @@ def battery(request):
 
 
 @staff_member_required
-def wifi(request):
+def wifi(request, ssid=''):
     try:
         enable_wifi()
         wifi_list = AvailableWifiNetwork.all()
@@ -124,6 +124,19 @@ def wifi(request):
         messages.error(request, e)
         return render(request, 'serveradmin/wifi.html')
 
+    if ssid:
+        try:
+            network = wifi_list[ssid]
+            network.connect()
+            messages.success(
+                request, _('Connected to {ssid}').format(ssid=ssid))
+
+        except KeyError:
+            messages.error(
+                request, _('No such network: {ssid}').format(ssid=ssid))
+
+        except WifiError as e:
+            messages.error(request, e)
 
     return render(
         request, 'serveradmin/wifi.html', {'wifi_list': wifi_list.values()})
--- a/ideascube/serveradmin/views.py
+++ b/ideascube/serveradmin/views.py
@@ -127,7 +127,8 @@ def wifi(request, ssid=''):
     if ssid:
         try:
             network = wifi_list[ssid]
-            network.connect()
+            wifi_key = request.POST.get('wifi_key', '')
+            network.connect(wifi_key=wifi_key)
             messages.success(
                 request, _('Connected to {ssid}').format(ssid=ssid))
 
--- a/ideascube/serveradmin/views.py
+++ b/ideascube/serveradmin/views.py
@@ -11,6 +11,7 @@ from ideascube.decorators import staff_member_required
 
 from .backup import Backup
 from .utils import call_service
+from .wifi import AvailableWifiNetwork, enable_wifi, WifiError
 
 
 @staff_member_required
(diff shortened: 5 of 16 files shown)


docs/contributing.md
```diff
@@ -64,6 +64,17 @@ Create a virtualenv (we name it `ideascube` here but the name is up to you):
...
+
+    wget http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.0.tar.gz
+    tar xf dbus-python-1.2.0.tar.gz
+    cd dbus-python-1.2.0
+    ./configure --prefix=$VIRTUAL_ENV
+    make
+    make install
+    cd ..
+    rm -fr dbus-python-1.2.0*
+
```
yohanboniface: Not pip installable?

bochecha: Unfortunately, no. :disappointed: 

Unless pip is capable of installing something that uses the Autotools to build?

bochecha:
Talking with @yohanboniface and Grégoire, supporting WEP networks is desirable, but not required in a first iteration.

Supporting hidden networks is not required or even desired at the moment, it may or may not come later on.

Therefore, I think this PR is now in a state where it could eventually get merged (pending review approval of course :wink: )

I had to do a few ugly things with the `NetworkManager` and `dbus` python modules. Discussing with @yohanboniface, this would work for a first iteration but long term we need to do something better (probably involving `txdbus`).

ideascube/serveradmin/templates/serveradmin/wifi.html
```diff
@@ -0,0 +1,47 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans "Manage wifi" %}</h2>
+{% if wifi_list %}
+   <ul class="wifi_list">
```
yohanboniface: Small detail: `wifi-list` to be consistent with other classes.

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+try:
+    from NetworkManager import (
+        NetworkManager, Settings as NMSettings, NM_DEVICE_TYPE_WIFI)
+
+except Exception as e:
+    import sys
+    sys.stderr.write("Could not import the NetworkManager module: %s\n" % e)
+    sys.stderr.write("Wifi management will not work properly\n")
+    sys.stderr.flush()
+
```
yohanboniface: I think we need this to be a bit more friendly, with a flag like `wifi_available`, that we'd check to display or not the wifi link in the admin and to call or not wifi related code.

requirements.txt
```diff
@@ -1,8 +1,10 @@
 batinfo==0.3
+-e git+https://github.com/bochecha/dbus-python.git@pipified-1.1.1#egg=dbus
```
yohanboniface: Not sure why we are not using upstream fork instead; https://github.com/posborne/dbus-python/network graph is not clear. Can you elaborate a bit?
Also, it would be great to push it on pypi. Have you been in contact with the upstream fork maintainers about that?
(Also, no need to -e in this case :) )

ideascube/serveradmin/templates/serveradmin/wifi.html
```diff
@@ -0,0 +1,47 @@
...
+               <span class="right">{{ wifi.strength }}%</span>  {# TODO: Maybe use a nice looking icon with bars? #}
+               {% if wifi.secure %}
+               <span class="right"><i class="fa fa-lock"></i></span>
+               {% endif %}
+           </a>
+       </li>
+   {% endfor %}
+   </ul>
+   <p><a href="{% url 'server:wifi_history' %}">History</a></p>
+{% else %}
```
yohanboniface: You can use `empty` tag instead of if/else.

ideascube/serveradmin/templates/serveradmin/wifi.html
```diff
@@ -0,0 +1,47 @@
...
+<div id="passphrase_popup" class="overlay">
+    <div class="popup">
+        <h2>Authentication required</h2>
+        <a class="close" href="#">&times;</a>
+        <div class="content">
+            <p>An encryption key is required to access this network.</p>
+            <form method="POST" action="">
+                {% csrf_token %}
+                <input type="password" id="wifi_key" name="wifi_key" autocomplete="off" />
+                <input type="submit" value="Connect" />
```
yohanboniface: i18n ;)

ideascube/serveradmin/templates/serveradmin/wifi.html
```diff
@@ -0,0 +1,47 @@
...
+{% else %}
+   <p>{% trans 'No wifi available' %}</p>
+{% endif %}
+
+<div id="passphrase_popup" class="overlay">
+    <div class="popup">
+        <h2>Authentication required</h2>
+        <a class="close" href="#">&times;</a>
+        <div class="content">
+            <p>An encryption key is required to access this network.</p>
```
yohanboniface: i18n ;)

ideascube/static/ideascube/js/ideascube.js
```diff
@@ -265,3 +265,26 @@ ID.confirmClick = function (selector) {
     };
     el.addEventListener('click', ask, false);
 };
+
+
+ID.initWifiList = function(item_selector, popup_selector, urlroot) {
```
yohanboniface: Seems like eslinting is missing ;)

ideascube/static/ideascube/js/ideascube.js
```diff
@@ -265,3 +265,26 @@ ID.confirmClick = function (selector) {
...
+        var known = element.getAttribute('data-known') === 'True';
+        var secure = element.getAttribute('data-secure') === 'True';
+
+        if (known || !secure) {
+            var ssid = element.getAttribute('data-ssid');
+            element.setAttribute('href', urlroot + ssid);
+        } else {
+            element.setAttribute('href', popup_selector);
+
+            element.addEventListener('click', function(evt) {
```
yohanboniface: This would deserve a generic calling method, that would take the popup element selector as argument and a callback at minimum. So we can reuse it elsewhere.

ideascube/static/ideascube/main.css
```diff
@@ -737,6 +737,71 @@ img.cover {
...
+    cursor: pointer;
+}
+.wifi_network span {
+    margin: auto 1em;
+}
+.right {
+    float: right;
+}
+
+.overlay {
```
yohanboniface: Move that section on a more generic section of the CSS.

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+        for connection in NMSettings.ListConnections():
+            if '802-11-wireless' not in connection.GetSettings():
+                continue
+
+            result.append(cls(connection))
+
+        return sorted(result, key=attrgetter('ssid'))
+
+    def __init__(self, connection):
+        self.__connection = connection
```
yohanboniface: Why two `__`? Ref: https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+    def __get_connection(self):
+        for connection in NMSettings.ListConnections():
+            settings = connection.GetSettings()
+
+            try:
+                if settings['802-11-wireless']['ssid'] == self.ssid:
+                    return connection
+
+            except KeyError:
+                # This is not a Wi-Fi connection
```
yohanboniface: Needs a dot at the end of the sentence :)

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+    # and hardcoding it allows the tests to run anyway
+    NM_DEVICE_TYPE_WIFI = 2
+
+
+class AvailableWifiNetwork(object):
+    @classmethod
+    def all(cls):
+        result = []
+
+        if not NetworkManager.WirelessEnabled:
```
yohanboniface: I'd add a check covering case when NetworkManager is not available at all (cf my previous comment :) ).

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+    sys.stderr.write("Wifi management will not work properly\n")
+    sys.stderr.flush()
+
+    # Set these to None, people will get AttributeError if they try calling
+    # the wifi views, but they have been warned above
+    NetworkManager = None
+    NMSettings = None
+
+    # Hardcode this one, it is a constant anyway in the NetworkManager module,
+    # and hardcoding it allows the tests to run anyway
```
yohanboniface: dot at the end of the comment :)

ideascube/serveradmin/views.py
```diff
@@ -111,3 +112,47 @@ def backup(request):
...
+                break
+
+    return render(request, 'serveradmin/wifi.html', {'wifi_list': wifi_list})
+
+
+@staff_member_required
+def wifi_history(request):
+    wifi_list = KnownWifiConnection.all()
+
+    if request.method == 'GET':
```
yohanboniface: I'd do the opposite: check if we are in POST to process request.POST, OR in any other case respond as if we are in GET mode. That's stronger (if you have a HEAD or OPTIONS request, you'll in trouble, with the current implementation ;) ).

ideascube/serveradmin/views.py
```diff
@@ -111,3 +112,47 @@ def backup(request):
...
     return render(request, 'serveradmin/battery.html',
                   {'batteries': batinfo.batteries()})
+
+
+@staff_member_required
+def wifi(request, ssid=''):
+    wifi_list = AvailableWifiNetwork.all()
+
+    if ssid:
+        for wifi in wifi_list:
```
yohanboniface: Seems that you are often looping through the wifi list to match the ssid: I think that would at least deserve a method of the AvailableWifiNetwork (something like `get(ssid)`), plus it may be worth an OrderedDict instead of a list, so we don't have to loop.

ideascube/serveradmin/templates/serveradmin/wifi_history.html
```diff
@@ -0,0 +1,26 @@
...
+       <label class="wifi_network" for="{{ wifi.ssid }}">
+           <input type="checkbox" id="{{ wifi.ssid }}" name="{{ wifi.ssid }}" />
+           {{ wifi.ssid }}
+           {% if wifi.connected %}
+           <span><i class="fa fa-check"></i></span>
+           {% endif %}
+       </label>
+   {% endfor %}
+   </fieldset>
+   <input type="submit" value="Forget Selected" />
```
yohanboniface: i18n :)

ideascube/serveradmin/templates/serveradmin/wifi.html
```diff
@@ -0,0 +1,47 @@
...
+{% if wifi_list %}
+   <ul class="wifi_list">
+   {% for wifi in wifi_list %}
+       <li>
+           <a class="wifi_network" data-ssid="{{ wifi.ssid }}" data-known="{{ wifi.known }}" data-secure="{{ wifi.secure }}">
+               {{ wifi.ssid }}
+               {% if wifi.connected %}
+               <span><i class="fa fa-check"></i></span>
+               {% endif %}
+               <span class="right">{{ wifi.strength }}%</span>  {# TODO: Maybe use a nice looking icon with bars? #}
```
yohanboniface: I think you can remove the TODO, the percentage is good enough :)

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+    NM_DEVICE_TYPE_WIFI = 2
+
+
+class AvailableWifiNetwork(object):
```
yohanboniface: `__str__/__unicode__` would help when debugging :)

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,183 @@
...
+
+                # There usually are more than one network for a given SSID
+                # (for example at 2.4 and 5GHz)
+                # Displaying each more than once would confuse users, so let's
+                # only pick the one with the highest frequency.
+                for ap in sorted(grouper, key=attrgetter('Frequency')):
+                    result.append(cls(ap, device))
+                    break
+
+            break
```
yohanboniface: koombooks will have two wifi cards, not sure how we can handle that the cleanway, but here we have one chance to loop on the bad one :/

requirements.txt
bochecha: Right, so here's the thing.

Upstream dbus-python is at http://cgit.freedesktop.org/dbus/dbus-python/. It can't be installed by pip, because it uses the Autotools, not the typical `setup.py`.

Somebody forked dbus-python to add a `setup.py` and make it pip-installable, it is at : https://github.com/posborne/dbus-python

Unfortunately, they only made the `master` branch pip-installable. That requires DBus >= 1.6. Our target for deployment is Jessie, which has (as of today) DBus 1.8.20.

Our CI is Ubuntu Precise, though, which has only 1.4.18, and given that it's an old LTS, it is very unlikely they will upgrade to DBus 1.6.

So our deployment could very well just use the fork from posborne, but not our CI.

So I forked the fork, and I **only** reapplied the patches that make it pip-installable onto the 1.1.1 release, and pointed the `requirements.txt` file to it.

This was a quick and dirty way to make both the CI and the target deployment work.

(I'll note that this was explained in the commit message, did you found it unclear?)

I could sure get that merged back into posborne's fork, but what's the point, if we move to something else than dbus-python anyway? (we were talking about txdbus the other day)

As for getting it on pypi, we would then use the "dbus-python" namespace for a fork, taking it away from the upstream maintainer. I find this rather hostile. To me, if anything goes on pypi, it has to come from upstream itself (which means posbourne's patches need to get merged upstream at fdo)

ideascube/static/ideascube/js/ideascube.js
bochecha: i'm not sure I understand what you mean here.

ideascube/static/ideascube/js/ideascube.js
bochecha: The popup is only displayed in some cases (the `else` case), and it seems very dependent on the list, its items, and what's inside.

I'm not sure that could be reused directly anywhere else.

In general, I'm a bit cautious about making reusable functions before there's actually a place where it would be nice to reuse it, because only when you have that need can you really see how you'd want to reuse it. (i.e the needed API)

Can't we wait for having a need to display other popups, and at that point we can change that code, make it reusable, and reuse it?

ideascube/static/ideascube/main.css
bochecha: Just to be sure, you don't mean the `wifi_*` (which I need to rename `wifi-*` :stuck_out_tongue_winking_eye:) classes, only the `right`, `overlay`, and all the popup-related stuff?

ideascube/serveradmin/wifi.py
bochecha: That `__connection` attribute is a NetworkManager object. That is, it is an implementation detail, and should not be considered public API.

Sure, I could use only one `_` and rely on convention. But does it actually matter whether we rely on convention or on name mangling?

All I think matters is that this should be a private API, so that if we change the underlying implementation (say we move from NetworkManager to something else) then the rest of the app (the API user) is completely unaffected.

If you feel strongly about using `_` instead of `__`, I'll change it.

yohanboniface:
`AvailableWifiNetwork.all()` seems a bit slow. Do you know why?

yohanboniface:
Also, we should now add a home card for wifi management :)

ideascube/static/ideascube/js/ideascube.js
yohanboniface: http://eslint.org/ :)

ideascube/static/ideascube/main.css
yohanboniface: Yep :)

ideascube/serveradmin/wifi.py
bochecha: The clean way is to just not `break`.

Then we'll loop through all devices, append all the found access points, and return them all as one big list.

It will be completely transparent to users (because they really shouldn't care which AP was found by which device, or even that there are more than one device).

The only issue I can see if is if both device find the same SSIDs, then the list will have things doubled.

bochecha:
> AvailableWifiNetwork.all() seems a bit slow. Do you know why?

No idea. It's completely instantaneous here. :confused: 

I'll see if I can figure it out.

> Also, we should now add a home card for wifi management :)

Really?

Won't that make the home page very long and unwieldy if we just add home cards for everything?

I thought Wifi management would be fine inside some kind of "admin" card, and was instead going to suggest that we rename the "stop/restart" card to be a more general server management card. :smile:

requirements.txt
yohanboniface: Thanks, that's clear now :)
But let's then move the fork on ideascube org :)

> (I'll note that this was explained in the commit message, did you found it unclear?)

Ooops, sorry, missed that: I'm not looking at each commit when reviewing a PR ;)

> I could sure get that merged back into posborne's fork, but what's the point, if we move to something else than dbus-python anyway?

I'd submit a PR anyway if it's not too much work, we are never sure what's the future will be ;)

> As for getting it on pypi, we would then use the "dbus-python" namespace for a fork, taking it away from the upstream maintainer.

You mean it's already on pypi even without setup.py? o_O

yohanboniface:
> I'll see if I can figure it out.

Takes close to 10 seconds here :s

```
In [1]: from ideascube.serveradmin.wifi import AvailableWifiNetwork

In [2]: import time

In [3]: before = time.time(); AvailableWifiNetwork.all(); print(time.time() - before)
9.08472299576
```

I'll try to investigate a bit if I'm the only one to reproduce ;)

yohanboniface:
> Won't that make the home page very long and unwieldy if we just add home cards for everything?

Those cards are only visible by staff members, so the home page for them is more or less the admin page with all useful links, so I think it's ok :)

ideascube/serveradmin/wifi.py
yohanboniface: @fheslouin can we use indistinctly both wifi cards?

> The only issue I can see if is if both device find the same SSIDs, then the list will have things doubled.

Using an OrderedDict may help :)

ideascube/serveradmin/wifi.py
yohanboniface: > That is, it is an implementation detail, and should not be considered public API.

One `_` means just that, while two means something else, so let's go on one `_` yes :)

ideascube/static/ideascube/js/ideascube.js
yohanboniface: > In general, I'm a bit cautious about making reusable functions before there's actually a place where it would be nice to reuse it

Agreed :)

> Can't we wait for having a need to display other popups, and at that point we can change that code, make it reusable, and reuse it?

I already have those use case, but OK let's move as is, I'll refactor in a later step :)

yohanboniface:
r+wc, very good job! :)

requirements.txt
bochecha: > You mean it's already on pypi even without setup.py? o_O

Even if it wasn't, it's still not very nice to take pypi/dbus-python when we're not dbus-python upstream.

But in any case, yes, it already is there:
https://pypi.python.org/pypi/dbus-python/

It's a very old version, though.

My guess is that upstream uploaded it there after being told that dbus-python should be on pypi, without knowing that it wouldn't work without a `setup.py` file (keep in mind they are OS-level, C hackers, not Python devs), and then when they realized it didn't work they just stopped sending newer versions.

I'm sure they'd be happy to upload newer version again after they add the `setup.py` upstream, if that ever happens.

bochecha:
Just pushed some updates to the code.

All of the minor things you pointed out should be fixed.

As for the big ones...

We now tell the user (in the web UI) that the Wi-Fi hardware is disabled (I have an old laptop with an actual **physical** switch to test that :smile:), and if it is only disabled in software (i.e the "airplane mode"), we just enable it automatically.

I'm now looping **much** less when listing the available networks, so it shouldn't be as horribly slow as it was. Can you try it again? :wink: 

I still need to add a card on the home page.

One last thing is the management of the two wifi cards on the Koombook. Talking with @yohanboniface it came out that one of these cards is used to connect to the internet with the feature implemented in this pull-request, and the other one is exposing a local access point.

That AP is done with hostapd, not NetworkManager, so we should make sure that NetworkManager knows not to touch this second card, and then we won't have any problem with the wifi management.

This is something to do in https://github.com/ideascube/ansiblecube though, I'll send a pull-request doing that soon.

And while typing this, I see that the tests failed in Travis. I'll fix it soon, but not tonight. :stuck_out_tongue_winking_eye:

ideascube/serveradmin/tests/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+
+except Exception:
+    # Use hard-coded values for the tests, these are constants anyway
+    NM_DEVICE_TYPE_ETHERNET = 1
+    NM_DEVICE_TYPE_WIFI = 2
+
+
+class NMAccessPoint(object):
+    def __init__(self, ssid, strength, wpa_flags):
+        self.Ssid = ssid
```
yohanboniface: Humm, not sure about using capitalized properties. I guess it's for consistency with NetworkManager conventions, but then we are hitting python itself (and ideascube) conventions as you know, and I'd better arbitrate for the later :)
(Eg. https://github.com/ideascube/ideascube/pull/197/files#diff-7f6f126831854a0d948da9d936257df2R27 ;) )

ideascube/serveradmin/tests/__init__.py
```diff
@@ -0,0 +1,59 @@
+try:
+    from NetworkManager import NM_DEVICE_TYPE_ETHERNET, NM_DEVICE_TYPE_WIFI
+
```
yohanboniface: Humm, is this line pep8? :)

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,208 @@
...
+
+        return OrderedDict(
+            [(n.ssid, n) for n in sorted(
+                result, key=attrgetter('strength'), reverse=True)])
+
+    def __init__(self, access_point, device, known_connections=None):
+        self._access_point = access_point
+        self._device = device
+
+        known_connections = known_connections if known_connections else {}
```
yohanboniface: `known_connections or {}` ?

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,208 @@
...
+        self._connection = known_connections.get(self.ssid, None)
+
+    def __str__(self):
+        return 'Available Wi-Fi network: %r' % (self.ssid)
```
yohanboniface: We may have non ascii network name :/

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,208 @@
...
+        if self._connection is not None:
+            NetworkManager.ActivateConnection(
+                self._connection._connection, self._device, "/")
+
+        else:
+            self._connection = self._new_connection(wifi_key=wifi_key)
+
+    @property
+    def connected(self):
+        if self._connection is None:
```
yohanboniface: return `self._connection and self._connection.connected` ?

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,208 @@
...
+            result.append(cls(connection))
+
+        return OrderedDict(
+            [(c.ssid, c) for c in sorted(result, key=attrgetter('ssid'))])
+
+    def __init__(self, connection):
+        self._connection = connection
+
+    def __str__(self):
+        return 'Known Wi-Fi connection: %r' % (self.ssid)
```
yohanboniface: Same here :)

requirements.txt
```diff
@@ -1,8 +1,10 @@
 batinfo==0.3
+git+https://github.com/bochecha/dbus-python.git@pipified-1.1.1#egg=dbus
```
yohanboniface: Can you add here the comment you made on the first review on why we use your fork?
Also, can you move the fork on ideascube organization?

yohanboniface: Also, I'd like to be sure we have made everything we can for this to be solved asap: ticket on Travis so we can use a newer version, ticket on the github dbus fork so they merge our changes, ticket on the upstream dbus repo so they release on pypi, etc. :)
We will not wait for anything, but I just want to help pushing the community forward on those topic so the situation may get cleaner at some point :)

ideascube/serveradmin/tests/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        self.Connection = NMConnection(True, ssid=ssid, is_secure=is_secure)
+
+
+class NMConnection(object):
+    def __init__(self, is_wifi, ssid='', is_secure=True, delete_func=None):
+        self.is_wifi = is_wifi
+        self.is_secure = is_secure
+        self.ssid = ssid
+
+        if delete_func is not None:
```
yohanboniface: This seems only used for testing, why not using monkeypatch instead?

yohanboniface: Meeeeh, forget, we are in tests :p

ideascube/serveradmin/tests/__init__.py
yohanboniface: Forget it too, now I understand those are test features :s

ideascube/serveradmin/wifi.py
```diff
@@ -0,0 +1,208 @@
...
+        return OrderedDict(
+            [(n.ssid, n) for n in sorted(
+                result, key=attrgetter('strength'), reverse=True)])
+
+    def __init__(self, access_point, device, known_connections=None):
+        self._access_point = access_point
+        self._device = device
+
+        known_connections = known_connections if known_connections else {}
+        self._connection = known_connections.get(self.ssid, None)
```
yohanboniface: `None` is useless here AFAIK :)

yohanboniface:
r+wc :)

ideascube/serveradmin/tests/__init__.py
bochecha: Yeah, I'm mocking the whole NetworkManager dbus API, replacing some objects by these.

Therefore, I have to use the same API they export, and that means capitalized properties.

ideascube/serveradmin/wifi.py
bochecha: I know, I add it out of habit, to be entirely explicit on the expected default value.

yohanboniface:
Yeeeeee-haaaaaa :)

fheslouin:
Great ! Can we change the release number thought ?

yohanboniface:
As we are in a work session with @bochecha I'd wait the other bits of its work. Does that urge on your side? :)

fheslouin:
Ok, I'm gonna build a package and try it on the KoomBook. I'll let you know if I found anything weird.

The next 15 KoomBook have to be ready by the end of the day. It would be great if they can have this feature.

requirements.txt
```diff
@@ -5,4 +5,19 @@ django-select-multiple-field==0.4.1
...
+#
+# So there is a fork on Github that **only** makes it pipi-installable:
+#     https://github.com/posborne/dbus-python
+#
+# But that's for the master branch... which we can't use, because our Travis
+# CI base is Ubuntu 12.04, which has DBus 1.4.18, and dbus-python master
+# requires DBus >= 1.6.
+#
+# So we forked the forked, backporting the pip-related patches to the 1.1.1
+# branch, and we use that here.
```
Natim: OMG

PR #210, opened 2015-11-17
wifi: Properly handle users making mistakes with the Wi-Fi key

bochecha:
NetworkManager allows creating a new connection either for just the current user, or shared with all other users on the system.

Before this commit, we were doing the latter, which requires additional permissions.

Since we only really care about the ideascube user, let's only create a connection for that user.

### DIFF
--- a/ideascube/serveradmin/wifi.py
+++ b/ideascube/serveradmin/wifi.py
@@ -3,6 +3,8 @@ from itertools import groupby
 from operator import attrgetter
 from uuid import uuid4
 
+from dbus import DBusException
+
 from django.utils.translation import ugettext as _
 
 try:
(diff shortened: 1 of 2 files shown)


ideascube/serveradmin/wifi.py
```diff
@@ -72,6 +72,7 @@ def _new_connection(self, wifi_key=None):
                 'id': self.ssid,
                 'type': '802-11-wireless',
                 'uuid': str(uuid4()),
+                'permissions': ['user:mathieu:'],
```
yohanboniface: mathieu? :)

ideascube/serveradmin/wifi.py
```diff
@@ -97,7 +103,20 @@ def _new_connection(self, wifi_key=None):
...
+
+        except DBusException as e:
+            excname = e.get_dbus_name()
+            name = 'org.freedesktop.NetworkManager.Settings.InvalidProperty'
+
+            excmsg = e.get_dbus_message()
+            keyprop = '802-11-wireless-security.psk:'
+
+            if excname == name and excmsg.startswith(keyprop):
+                raise WifiError(_('Invalid Wi-Fi key'))
```
yohanboniface: Can you add a minimal unittest for that case?

bochecha:
So I went back on this one, and removed the code that was creating per-user connections.

When storing per-user connections, NetworkManager wants an actual user session (login).

Given that out target for deployment is a system-level service, we don't have user sessions.

Storing a system-level session also means that the connection is activated at boot-time, rather than when the user logs in.

All in all, the change I had made which was storing per-user connections was wrong, and I now just removed it from this pull request.

The other one still holds though: we can't just return a 500 error when the user sends an invalid Wi-Fi passphrase.

I only added unit tests for it as you requested @yohanboniface.

PR #212, opened 2015-11-18
csv: Do not export private user information

bochecha:
We don't need to know about a user's real name, the (hopefully anonymous) serial is enough for our data collection purposes.

I'm only hiding the fields @barbayellow explicitly mentioned to me. However, there seems to be other potentially private informations in other fields, that we probably shouldn't have in CSV exports. Things like the family or marital status, or even the gender.

In any case, those would just need to be added to the new `User.PRIVATE_DATA` field, and then they would not be exported to CSVs any more.

### DIFF
--- a/ideascube/models.py
+++ b/ideascube/models.py
@@ -62,6 +62,11 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
         ('r', _('Read')),
     )
 
+    PRIVATE_DATA = (
+        'short_name',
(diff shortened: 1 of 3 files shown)


barbayellow:
Family status, marital status or gender are not personal data as long as long as these information are not connected to any kind of name. If there is no way to make the connection between those information and a real person, this is acceptable. This is why names must not be sent. Never. And I am wondering if the serial should even be exported.

Issue #202, opened 2015-11-04
Actvate / de activate any service or application on ideascube


barbayellow:
The ideascube software can run on several hardware. Amongst those, BSF' KoomBook. It's a lightweight computer powered by an Olimex card with only 1 Go of ram memory. On this kind of light configuration, Khalite for instance, is already using 1/3 of the ram.

Therefore, we must offer the ability to shut down some services and allow the administrator to choose which services will be running on the server. The admin will choose to run only a few services from an extensive list. Not only will it resources but it will also be an improvement for the users : the admin will be able to offer targeted services to a specific audience. For instance : games and educational content for kids. 

For a better understanding of this issue, a UI mockup is available here : https://projects.invisionapp.com/share/PJ4F6QFUA#/screens/105752891

barbayellow:
The ideascube server has been running and tested on 2 different OS : Lubuntu 14.04 and Debian 8.0. The ultimate goal is to run the ideascube on any kind of GnuLinux system but - given our limited time and ressources -  we should stick to one OS preferably and give up this dual support. Debian 8 shoud be the OS of choice.

PR #213, opened 2015-11-18
Rework the management of the services

bochecha:
This ports the service management view to systemd and PolicyKit.

It does not change the UX at all, this can come in a further pull request.

This is part of the work towards fixing issue #202.


ideascube/conf/base.py
```diff
@@ -252,7 +252,6 @@
     {
         'name': 'kalite',
         'description': _('Daemon which provides KhanAcademy on lan'),
-        'status_caller': lambda x: {'status': False} if subprocess.call(['pgrep', 'kalite', '-f']) else {'status': True}  # noqa
```
yohanboniface: We can remove "ideascube" too, and add "uwsgi" instead.

bochecha: Sure.

I figured that choosing which services we manage was orthogonal to this rework of how we managed, and it could come in another pull-request. :smile:

ideascube/serveradmin/views.py
```diff
@@ -18,33 +18,39 @@
...
-        elif 'stop' in request.POST:
-            service_action = 'stop'
-        elif 'restart' in request.POST:
-            service_action = 'restart'
-    else:
-        active_service = None
+    manager = Manager()
+    to_manage = request.POST.get('name')
 
     for service in services:
```
yohanboniface: Do we still need this loop?

yohanboniface:
Nice move! :)

PR #221, opened 2015-11-28
WPA config — Fixes #98

### LINKED ISSUE #98: Create an UI to change the wifi password
Replacing by regex on the hostapd.conf file?

Natim:


ideascube/serveradmin/wifi.py
```diff
@@ -218,3 +218,67 @@ def get_wifi_device():
...
+        with open(configfile) as fd:
+            self.data = fd.readlines()
+        # update option with current passphrase and check if WPA is enable
+        for line in self.data:
+            if '=' in line and 'wpa_passphrase' in line:
+                 self.is_enable = not line.startswith('#')
+                 line = line.strip('# ')
+                 passphrase = line.split('=', 1)[1].strip()
+                 self.options['wpa_passphrase'] = passphrase
+        self.new_data = self.data[:]
```
Natim: is ``self.data[:]`` better than ``list(self.data)`` ?

ideascube/serveradmin/views.py
```diff
@@ -167,3 +168,26 @@ def wifi_history(request):
     return render(
         request, 'serveradmin/wifi_history.html',
         {'wifi_list': wifi_list.values()})
+
+
+@staff_member_required
+def wpa_config(request):
```
yohanboniface: This view needs to be unittested :)

ideascube/serveradmin/wifi.py
```diff
@@ -218,3 +218,67 @@ def get_wifi_device():
...
+         'rsn_pairwise': 'CCMP',
+    }
+
+    def __init__(self, configfile='/etc/hostapd/hostapd.conf'):
+        self.is_enable = False
+        # read data from config
+        with open(configfile) as fd:
+            self.data = fd.readlines()
+        # update option with current passphrase and check if WPA is enable
+        for line in self.data:
```
yohanboniface: This piece of code also needs a bit of unittests :)

ideascube/serveradmin/wifi.py
```diff
@@ -218,3 +218,67 @@ def get_wifi_device():
...
+    options = {
+         'wpa': '2',
+         'wpa_passphrase': '',
+         'wpa_key_mgmt': 'WPA-PSK',
+         'wpa_pairwise': 'TKIP',
+         'rsn_pairwise': 'CCMP',
+    }
+
+    def __init__(self, configfile='/etc/hostapd/hostapd.conf'):
+        self.is_enable = False
```
yohanboniface: is_enabled?

ideascube/serveradmin/wifi.py
```diff
@@ -218,3 +218,67 @@ def get_wifi_device():
...
+        # read data from config
+        with open(configfile) as fd:
```
yohanboniface: catch file not found?

ideascube/serveradmin/wifi.py
```diff
@@ -218,3 +218,67 @@ def get_wifi_device():
...
+                 line = line.strip('# ')
+                 passphrase = line.split('=', 1)[1].strip()
+                 self.options['wpa_passphrase'] = passphrase
+        self.new_data = self.data[:]
+        self.configfile = configfile
+
+    def save(self):
+        """save data to configfile"""
+        with open(self.configfile, 'w') as fd:
+            fd.write('\n'.join(self.new_data))
```
bochecha: How will that save a file in `/etc`?

The application does not run as root.

Natim: What we did is to create a group with the ideascube user and give write permission to that group on that file. It will come in another PR about the Ansible recipe.

yohanboniface:
Let's have a quick discussion to see if it's worth using NetworkManager instead of making ideascube sudoer on /etc/hostapd conf.

See: http://cgit.freedesktop.org/NetworkManager/NetworkManager/tree/examples/python/dbus/wifi-hotspot.py

cc @bochecha

yohanboniface:
One point to keep in mind (but I've not solution for that), is: what happen when we update hostapd.conf from ansible runs?

fheslouin:
We have 2 options to update hostapd.conf : 
  - use regex to add/delete/modify some of the parameters in the conf file 
  - simply erase the file with the new version and the settings will be back to default, they will have to modify them to set theire conf again (this is not the best solution)

First option is the best and I think it is doable with regex to update some part of the file.

fheslouin:
It would be nice to have this feature merge in the next release. 
Currently all the KoomBook and IdeasBox are configured with their machine name instead of a nice hotspot name.

It's easier for us to configure and manage the device, but not really nice for the end user.

yohanboniface:
I'd like to dig on NetworkManager option before (see https://github.com/ideascube/ideascube/pull/221#issuecomment-160427997)

bochecha:
Ideascube moved to Gitlab.

https://framagit.org/ideascube/ideascube/merge_requests/221

PR #227, opened 2015-11-29
Really fast forward PoC of a package manager.

Natim:
Just for the demo.

### DIFF
--- a/ideascube/serveradmin/views.py
+++ b/ideascube/serveradmin/views.py
@@ -1,11 +1,14 @@
 from subprocess import call
 
 import batinfo
+import os
+import requests
 from django.conf import settings
 from django.contrib import messages
 from django.http import StreamingHttpResponse
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.utils.translation import ugettext as _
+from kinto_client import Client
 
 from ideascube.decorators import staff_member_required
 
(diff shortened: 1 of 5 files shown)


ideascube/serveradmin/views.py
```diff
@@ -167,3 +170,41 @@ def wifi_history(request):
...
+    packages = client.get_records()
+
+    return render(
+        request, 'serveradmin/packages.html',
+        {'packages': packages})
+
+
+@staff_member_required
+def package_detail(request, package_id):
+    client = Client(server_url='http://kinto.bsf-intranet.org/v1',
```
almet: the url should come from the configuration.

almet: The server might not have access to an external repository here. Thhis brings a use case of an offline-first client for Kinto.py (which is good) but also means that this PoC doesn't handle the offline case just yet.

Natim: Note that you cannot install a remote package while offline, so this is a limitation of the feature.
But yeah, it definitely brings the usecase for offline first kinto.py :) I opened  about this https://github.com/Kinto/kinto.py/issues/25

almet: You could at least list them so someone can select them without any connection and then create a script (or anything) that retrieves the packages to put on an USB stick, for instance.

yohanboniface:
#281 ;)

PR #228, opened 2015-12-03
Do not load files into memory when importing medias

yohanboniface:


yohanboniface:
@bochecha around for a quick review?

(Or any other volunteer? :) )

almet:
Looks good to me, don't we want tests for this as well?

yohanboniface:
> don't we want tests for this as well?

There are already for the import itself. Or do you mean a test that would check that the file is not totally loaded in memory? Not sure how I would do that, if so ;)

yohanboniface:
Merging, let's add specific tests if you have a point on that :)

PR #243, opened 2015-12-14
Debian packaging fixes

bochecha:
With this, we can build a Debian package for ideascube with the [sbuild](https://wiki.debian.org/sbuild) command, and with correct dependencies.

`sbuild` allows building the package in a chroot, populated only with packages from the Jessie repos (which entirely avoids issues like the one @fheslouin regularly has with libjpeg), in a completely unattended and (almost) reproducible way.

This in turn will allow us to build packages in a [Buildbot](https://buildbot.net/). And in fact, I have a local Buildbot which just successfully built a Debian package for Ideascube, which should make @barbayellow happy. :smile:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,14 @@ Source: ideascube
 Section: python
 Priority: extra
 Maintainer: Yohan Boniface <hi@yohanboniface.me>
-Build-Depends: debhelper (>= 9), python, dh-virtualenv (>= 0.6)
+Build-Depends: debhelper (>= 9), dh-virtualenv (>= 0.6),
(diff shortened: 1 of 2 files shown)


barbayellow:
That makes me happy indeed ! If I understood correctly what buildbot does, that means that we must now install and setup a buildbot master somewhere on the Internet. This master will make the slaves building the .deb on an arm ou amd platform depending on the platform they are installed on ? If I understood correctly, I am happy indeed :)

bochecha:
@barbayellow You understood correctly.

PR #259, opened 2016-01-07
do not limit imported path names (fix #229)

### LINKED ISSUE #229: cut filename at import
From a media import:
```
⚠ Skipping. original: * Ensure this filename has at most 100 characters (it has 102)..
```

dachary:
import_medias.add verifies if the path exists and will therefore fail if
the path length exceeds the underlying file system limits. Relax the
form constraints on the file name sizes for the document and the
associated preview.

Signed-off-by: Loic Dachary <loic@dachary.org>


yohanboniface:
Why not cutting names too long instead?

ideascube/mediacenter/tests/test_import_medias.py
```diff
@@ -99,6 +99,30 @@ def test_should_guess_kind_from_path():
     assert video.kind == Document.VIDEO
 
 
+def test_long_path():
+    assert not Document.objects.count()
+    long = 'AAAAAAAAAA' \
```
yohanboniface: `'A' * 80` (or whatever the number) should do the job ;)

yohanboniface: Also, it's better not to use python reserved keywords (`long`)

dachary:
> Why not cutting names too long instead?

I'm not sure I understand what is the problem with file names longer than 100 chars ?

dachary:
@yohanboniface thanks for your quick review :-) I addressed your comments and re-pushed.

yohanboniface:
> I'm not sure I understand what is the problem with file names longer than 100 chars ? 

What about a path of 100000 chars? ;)
In other words, we do need a limit. And whatever the limit we will decide, it will be too short for some use case. I think 100 is a fair limit. :)

dachary:
@yohanboniface ah, got it :-) The os.path.exists at https://github.com/ideascube/ideascube/blob/master/ideascube/mediacenter/management/commands/import_medias.py#L69 implicitly enforces the path name limit of the underlying file system, whatever it is. Does that sound like a reasonable upper bound to you ?

yohanboniface:
> Does that sound like a reasonable upper bound to you ?

OK, you have the point :)
But then a migration is needed to update the `max_length` of the already created databases. Command is something like:

    python manage.py makemigration mediacenter

Also, a mention in debian/changelog, please :)

dachary:
@yohanboniface I added a migration file but they are not used / tested when I run make test. Is there another way to verify it ?

yohanboniface:
```python manage.py migrate``` should run it on your database (if created before your change on the fields).
For the tests, they do not need migrations as they can read the latest schema, AFAIK.

dachary:
@yohanboniface I'm willing to add a test that validates migration files (i.e. creates the database using the initial schema file and applies each migration file in turn). I suppose there already exist tests in other django based software from which I could get inspiration. It would be great if you could suggest one. Otherwise I'll research, no worries ;-)

dachary:
@yohanboniface nvm, I misunderstood how the migrations files are used. Is there anything else I should do to improve this patch ?

dachary:
@yohanboniface this should be better, hopefully. Next time I'll read your message more carefully. I added 4602f6b because it looks like a migration file was missing. I don't think there is a matching issue but I may have missed it.

dachary:
@yohanboniface removed the lang commit

yohanboniface:
Thanks! :)

PR #260, opened 2016-01-07
server admin: ensure backup root exists (fix #244)

### LINKED ISSUE #244: backups dir is not created when uploading a backup
It is only created when we create a new backup, so when we want to upload a backup in a new box, it fails.

dachary:
When a Backup() object is instantiated, ensure the backup root directory
exists. It is implicitly created by the create() method, but not by
other methods such as load().

Signed-off-by: Loic Dachary <loic@dachary.org>


yohanboniface:
Thanks :)

PR #263, opened 2016-01-10
A few Wifi fixes

bochecha:
Last commit is a bit of a straw proposal for fixing the issue right now, but it clearly isn't how we should fix it in the long term.

### DIFF
--- a/ideascube/serveradmin/wifi.py
+++ b/ideascube/serveradmin/wifi.py
@@ -17,6 +17,7 @@ try:
     from NetworkManager import (
         NetworkManager,
         Settings as NMSettings,
+        NM_ACTIVE_CONNECTION_STATE_ACTIVATED,
         NM_DEVICE_TYPE_WIFI,
         )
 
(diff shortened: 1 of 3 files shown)


ideascube/serveradmin/wifi.py
```diff
@@ -171,7 +189,8 @@ def connected(self):
         for active_connection in NetworkManager.ActiveConnections:
             if (active_connection.Connection.GetSettings()
                     == self._connection.GetSettings()):
-                return True
+                if active_connection.State == NM_ACTIVE_CONNECTION_STATE_ACTIVATED:
```
yohanboniface: Am I missing something or can we merge those two `if`?

bochecha: There is nothing preventing us from merging them.

I kept them separate as a matter of taste, since each one of them is already long enough to go past the 80-chars limit and needs to be continued on a new line.

I'm happy to merge them if you prefer, it makes no difference to me. :smiley:

PR #268, opened 2016-01-20
Python3 makefile

mbaechler:
These modifications are based on my experiment to run ideascube into docker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,20 +3,20 @@ test:
 testcov:
 	py.test --cov=ideascube/ --cov-report=term-missing
 develop:
-	pip install -r requirements-dev.txt
+	pip3 install -r requirements-dev.txt
 doc:
 	mkdocs serve
 dummydata:
-	python manage.py dummydata
+	python3 manage.py dummydata
 collect_translations:
-	python manage.py makemessages -a --ignore=debian
-	python manage.py makemessages -d djangojs -a --ignore=debian
+	python3 manage.py makemessages -a --ignore=debian
+	python3 manage.py makemessages -d djangojs -a --ignore=debian
 push_translations:
 	tx push -s
 pull_translations:
 	tx pull
 compile_translations:
-	python manage.py compilemessages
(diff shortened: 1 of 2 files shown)


bochecha:
Except for those minor issues, it looks good overall, thanks!

mbaechler:
I submitted a fixup following your comments

bochecha:
All fixed now, thanks!

Makefile
```diff
@@ -3,20 +3,20 @@ test:
 testcov:
 	py.test --cov=ideascube/ --cov-report=term-missing
 develop:
-	pip install -r requirements-dev.txt
```
yohanboniface: Sorry for the lag, but as soon as your venv is made pointing to python3 you should not need that changes.
What was the rationale behind this?

mbaechler: I don't have a strong opinion about that, I did that because in my Dockerfile, I tried not to use a virtualenv. Feel free to revert this part if you think it's not a good idea.

yohanboniface: > I did that because in my Dockerfile, I tried not to use a virtualenv.

And `pip` or `python` was still pointing to python2? Was your venv made with python3 and sourced?

mbaechler: AFAIK, debian python is python 2. Without virtualenv, you have to explicitly call python3 if you want to use it.

I'm not sure to understand your question : 

> Was your venv made with python3 and sourced?

bochecha: @yohanboniface I initially asked the same question: in a virtualenv made for Python3, `bin/python` is `bin/python3` and therefore the change is unneeded.

But if it's unneeded in a virtualenv, it is also entirely harmless: in a virtualenv, `bin/python3` is `bin/python`, and therefore this change makes no difference.

However, this change makes it possible to run ideascube outside a virtualenv.

So I figured, if it makes no difference in a virtualenv, but it makes things better outside of a virtualenv, then why not? :smiley:

yohanboniface: OK, deal :)

PR #271, opened 2016-01-22
systemd: Fix bytes/unicode issue

bochecha:
The problem is that the stdout and stderr of a subprocess.Popen call are bytes, not unicode.

And then in the code, we mix that with string literals... which were bytes in Python2, but unicode in Python3.

The unit tests never found the problem, because we mock the Popen and its stdout/stderr, replacing the latter by StringIO.

And in Python3, StringIO take unicode, not bytes, so everything was just fine for the unit tests.

### DIFF
--- a/ideascube/serveradmin/systemd.py
+++ b/ideascube/serveradmin/systemd.py
@@ -95,6 +95,6 @@ def systemctl(action, unit):
         msg = ["Could not %s %s" % (action, unit)]
 
         if err:
-            msg.append(err.strip())
+            msg.append(err.strip().decode())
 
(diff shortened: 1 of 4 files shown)


ideascube/serveradmin/systemd.py
```diff
@@ -95,6 +95,6 @@ def systemctl(action, unit):
         msg = ["Could not %s %s" % (action, unit)]
 
         if err:
-            msg.append(err.strip())
+            msg.append(err.strip().decode('utf-8'))
```
yohanboniface: 'utf-8' is default ;)

PR #273, opened 2016-01-27
Protect deletion of users authors of blogs (fix #261)

### LINKED ISSUE #261: Content and user administration
When a user is deleted, all files and blogposts generated by the user are removed. We should be able to remove a user without removing any content.

yohanboniface:


almet:
I think the commit has a typo: "project vs protect".

yohanboniface:
> I think the commit has a typo: "project vs protect".

Good catch, fixed :)

PR #279, opened 2016-02-04
More granular reporting for import_medias (fix #275 fix #276)

### LINKED ISSUE #275: The `import_medias` command could be more verbose.
Importing medias should display a summary of the result, something like:

- 3 medias imported flawlessly
- 4 medias import failed (Missing summary 3, path not found 2, etc)

Maybe it should also helps to find the guilty medias by displaying its line number, or the whole line, or something.

### LINKED ISSUE #276: The `import_medias` command should return 1 on errors.
This would help when writing tools around the ideascube commands.

Tested on `ideascube import_medias --dry-run` only, this may apply to other commands, with or without the `--dry-run` option.

yohanboniface:
fix #275
fix #276

### DIFF
--- a/ideascube/mediacenter/management/commands/import_medias.py
+++ b/ideascube/mediacenter/management/commands/import_medias.py
@@ -13,6 +13,7 @@ from ideascube.mediacenter.models import Document
 from ideascube.mediacenter.forms import DocumentForm
 from ideascube.mediacenter.utils import guess_kind_from_content_type
 from ideascube.templatetags.ideascube_tags import smart_truncate
+from ideascube.management.utils import Reporter
 
 
 class Command(BaseCommand):
@@ -34,14 +35,6 @@ class Command(BaseCommand):
         self.stderr.write(msg)
         sys.exit(1)
 
-    def skip(self, msg, metadata=None):
-        self.stderr.write(u'⚠ Skipping. {}.'.format(msg))
-        if metadata:
-            for key, value in metadata.items():
-                value = value if value else ''
-                self.stdout.write(u'- {}: {}'.format(key, value))
-        self.stdout.write('-' * 20)
-
     def load(self, path):
         with codecs.open(path, 'r', encoding=self.encoding) as f:
             content = f.read()
@@ -56,17 +49,21 @@ class Command(BaseCommand):
         self.update = options['update']
         self.encoding = options['encoding']
         self.dry_run = options['dry_run']
+        self.report = Reporter(options['verbosity'])
         if not os.path.exists(path):
             self.abort('Path does not exist: {}'.format(path))
         self.ROOT = os.path.dirname(path)
         rows = self.load(path)
         for row in rows:
             self.add(row)
+        print(self.report)
+        if self.report.has_errors():
+            sys.exit(1)
 
     def add(self, metadata):
         title = metadata.get('title')
         if not title:
-            return self.skip('Missing title', metadata)
+            return self.report.error('Missing title', metadata)
         title = smart_truncate(title)
         metadata['title'] = title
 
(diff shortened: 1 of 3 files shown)


ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -34,13 +38,26 @@ def abort(self, msg):
...
-                self.stdout.write(u'- {}: {}'.format(key, value))
-        self.stdout.write('-' * 20)
+    def report(self, name, item, level):
+        if level == 1:
+            self.errored = True
+        if name not in self._reports:
+            self._reports[name] = []
+        self._reports[name].append((item, level))
+
+    def reporting(self):
```
almet: rename this to `output_report` or `display_report`? In general it's easier to understand methods that have a verb in them.

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -34,13 +38,26 @@ def abort(self, msg):
...
+            self.stdout.write('# Reports:')
+            for name, items in self._reports.items():
+                self.stdout.write('- {}: {}'.format(name, len(items)))
+                if self.verbosity > 0:
+                    for item, level in items:
+                        if self.verbosity >= level:
+                            out = self.stderr if level == 1 else self.stdout
+                            out.write('  . {}'.format(item))
+        self.stdout.write('\n')
+        if self.errored:
```
almet: maybe put this in a `self.handle_exit_code` method?

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -56,17 +73,21 @@ def handle(self, *args, **options):
...
         rows = self.load(path)
         for row in rows:
             self.add(row)
+        self.reporting()
 
     def add(self, metadata):
         title = metadata.get('title')
         if not title:
-            return self.skip('Missing title', metadata)
+            return self.report('Missing title', metadata, ERROR)
```
almet: naming: rename to add_error?

almet:
I've added a few comments. In general, why not relying on loggers?

almet:
You could try:

```py
logger = logging.getLogger('main')
logger.debug("message")
```

almet:
Oh, maybe it's a feature to wait the end of all operations to have an output?

yohanboniface:
> Oh, maybe it's a feature to wait the end of all operations to have an output?

Yes, we want to display some "reporting" at the end of the script, with how many success rows, how many errors, etc. And giving more details when running the script with more verbosity.

almet:
That could be great to have a "Reporter" class that you could initialize and then pass to the commands when needed, on which you have different possible methods: "error", "debug", etc. (like the ones from logging), and additionally a "create_report" one, wich outputs the report somewhere (to stdout for now).

What do you think?

yohanboniface:
Yes, sounds good. My plan was actually at some point to have a BaseCommand class that would handle those methods (report, print_report, etc.). Not sure which pattern is better though, but my plan was to wait for a second use case before trying to make this generic. I'll work on that this week-end :)

almet:
I usually tend to [prefer composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) :-)

I like to decouple things also because it's simpler to test. If you prefer waiting for a second use case to implement this maybe it's wiser. I'll let you decide !

yohanboniface:
@almet pushed a Report util. May need a bit more use cases to be refined, but that should be a first step :)

yohanboniface:
Example of output:
screenshot from 2016-02-08 16-35-34

bochecha:
Reviewed. :+1:

PR #281, opened 2016-02-08
Zim package manager

bochecha:
With this, we can manage optional "packages" on an Ideascube server.

There's a command line tool (the `catalog` Django management command), which uses an API inside Ideascube, preparing the way for having that in the web user interface.

The current status:

* [x] Manage remote sources
  * [x] Add a remote source
  * [x] Remove a remote source
  * [x] List configured remote sources
* [x] Manage the cache
  * [x] Update the local cache (like `apt-get update`, to know the packages offered by each source)
  * [x] Clear the local cache
  * [ ] Allow "incremental" cache updates (right now we just always download the entire catalog for each remote source)
* [x] Download packages
  * [x] Keep a copy in cache
  * [x] Resume incomplete downloads
  * [x] Support additional package caches (e.g on a USB stick)
* [x] Manage packages
  * [x] List installed packages
  * [x] List available packages
  * [x] Support multiple package types, as content for multiple apps
  * [x] Install specified packages (e.g `install wikipedia.fr wikipedia.en`)
  * [x] Remove specified packages (e.g `remove wikipedia.fr wikipedia.en`)
  * [x] Upgrade specified packages (e.g `update wikipedia.fr wikipedia.en`)
  * [x] Upgrade all installed packages if none is specified (i.e `update`)
  * [x] Support globs for list, install, remove and upgrade (e.g `list wikipedia.*`)
  * [x] Allow listing packages in need of an update (i.e `list --upgradable`)
  * [x] Unit tests for list, install, remove and upgrade

There is also absolutely no testing for the command line tool, only the API is tested. Not sure how to properly test the command line though, I guess I'll have to look at other command lines in Ideascube.


ideascube/conf/base.py
```diff
@@ -283,3 +283,6 @@
 
 SESSION_COOKIE_AGE = 60 * 60  # Members must be logged out after one hour
 BACKUP_FORMAT = 'zip'  # One of 'zip', 'tar', 'bztar', 'gztar'
+
+CATALOG_CACHE_BASE_DIR = '/tmp/ideascube-catalog'
```
yohanboniface: Does that mean the cached catalog is lost on reboot?

ideascube/conf/base.py
```diff
@@ -283,3 +283,6 @@
...
+CATALOG_KIWIX_INSTALL_DIR = '/tmp/kiwix'
```
yohanboniface: Are we using it? :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+
+        with zipfile.ZipFile(download_path, "r") as z:
+            names = z.namelist()
+            names = filter(lambda n: n.startswith('data/'), names)
+            z.extractall(install_dir, members=names)
+
+        zimname = os.path.basename(self.url).split('+', 1)[-1][:-4] + '.zim'
+        datadir = os.path.join(install_dir, 'data')
+
+        for zim in glob(os.path.join(datadir, '*', '{}*'.format(zimname))):
```
yohanboniface: IIRC, those are not only zim files (also index and xml), so I would suggest to use `path` instead of `zim` not to mislead future readers ;)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+        for path in glob(os.path.join(datadir, '*', zimname)):
+            try:
+                os.unlink(path)
+
+            except IsADirectoryError:
+                shutil.rmtree(path)
+
+
+# Register the package types
+Package.__TYPES__ = {
```
yohanboniface: Why not using a metaclass for such registry?

ideascube/conf/base.py
yohanboniface: OK, got it!

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+        except KeyError:
+            raise InvalidHandlerType(name)
+
+        handler = object.__new__(cls)
+        handler.__init__(name)
+
+        return handler
+
+    def __init__(self, name):
+        settingname = 'CATALOG_{}_INSTALL_DIR'.format(name.upper())
```
yohanboniface: In the future, some content may need more than one install dir, but let's say that's OK for now :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+                book = books[0]
+                book.set('path', 'data/content/%s' % zimname)
+                book.set('indexPath', 'data/content/%s.idx' % zimname)
+
+                library.append(book)
+
+        with open(self._library, 'wb') as f:
+            f.write(etree.tostring(
+                library, xml_declaration=True, encoding='utf-8'))
+
```
yohanboniface: we need to restart kiwix-server at some point

ideascube/conf/base.py
bochecha: If `/tmp` is mounted as `tmpfs`, and if deployments put their cache in there... then yes.

It's probably not the right location for the cache, for sure.

How about `/var/cache/ideascube` ? (but then, we get into permission-trouble territory :stuck_out_tongue_winking_eye: )

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+        def _progress(*args):
+            self._progress(' {}'.format(package.id), *args)
+
+        filename = '{0.id}-{0.version}'.format(package)
+
+        for cache in self._package_caches:
+            path = os.path.join(cache, filename)
+
+            if os.path.isfile(path):
+                if self._verify_sha256(path, package.sha256sum):
```
yohanboniface: Humm, I'm tempted to say that package.sha256sum should be optional, so the check. What do you think?

ideascube/conf/base.py
yohanboniface: > How about /var/cache/ideascube ? 

Yep

> (but then, we get into permission-trouble territory)

To be managed by debian postinst?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,459 @@
...
+        sha = sha256()
+
+        with open(path, 'rb') as f:
+            while True:
+                data = f.read(8388608)
+
+                if not data:
+                    break
+
+                sha.update(data)
```
yohanboniface: How big that will come for a 40Go files?

ideascube/conf/base.py
bochecha: > To be managed by debian postinst?

Sure, for the package.

Not a solution for the developer workstation, though.

yohanboniface: > Not a solution for the developer workstation, though.

Can be managed by `dev.py` dedicated settings. Also having a default in `STORAGE_ROOT/package_cache` for example.

ideascube/serveradmin/catalog.py
bochecha: Dunno, this felt simple and elegant, since the `Catalog` can just always use the `Package()` constructor, and `Package` itself figures out which package type to return.

How would you do the registry with a metaclass?

yohanboniface: Something like:

```python
class MetaPackage(type):

    __TYPES__ = {}

    def __new__(mcs, name, bases, attrs, **kwargs):
        cls = super().__new__(mcs, name, bases, attrs)
        if hasattr(cls, 'type'):
            mcs.__TYPES__[cls.type] = cls
            # Or we can have the registry on Package itself
        return cls

class Package(metaclass=MetaPackage):
    pass
```

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -0,0 +1,148 @@
...
+            title='Commands', dest='cmd', metavar='',
+            parser_class=argparse.ArgumentParser)
+        subs.required = True
+
+        required_ids = argparse.ArgumentParser('required_ids', add_help=False)
+        required_ids.add_argument(
+            'ids', nargs='+', metavar='ID', help='One or more package ids')
+
+        # -- Manage content ---------------------------------------------------
+        list = subs.add_parser('list', help='List packages')
```
yohanboniface: Would be friendly to say "No remotes yet please add one with..." when calling "list" while not having added any remote :)

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -0,0 +1,148 @@
...
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+from ideascube.serveradmin.catalog import NoSuchPackage, Catalog
+
+
+class Command(BaseCommand):
+    help = 'Manage apps and content'
+
+    def add_arguments(self, parser):
```
yohanboniface: Not sure if it's easy to do, but could be nice to display a nice full help message instead of an error when calling "manage.py catalog" without any command. Maybe adding an "help" command and setting it as default?

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def clear_cache(self, options):
+        self.catalog.clear_cache()
+
+    # -- Manage remote sources ------------------------------------------------
+    def list_remotes(self, options):
+        for remote in self.catalog.list_remotes():
+            print('{0.name:>25} : {0.url}'.format(remote))
+
+    def add_remote(self, options):
```
yohanboniface: I suggest to run a "cache update" after adding a remote catalog.

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -0,0 +1,148 @@
...
+        rm.add_argument('id', help='The id of the remote source to remove')
+        rm.set_defaults(func=self.remove_remote)
+
+    def handle(self, *args, **options):
+        self.catalog = Catalog(settings.CATALOG_CACHE_BASE_DIR)
+        options['func'](options)
+
+    # -- Manage content -------------------------------------------------------
+    def list_packages(self, options):
+        if not any([options[k] for k in ('installed', 'available')]):
```
yohanboniface: Suggestion: instead of having two groups ('installed' and 'available'), I'd suggest to have only one list, but having a 🗸 when the package is installed (and keeping  '--installed' and '--available' options to filters). Thoughts?

ideascube/serveradmin/catalog.py
bochecha: Fixed in the latest push to this branch.

ideascube/serveradmin/catalog.py
bochecha: To be honest, I find this harder to understand. :confused: 

Why the `hasattr(cls, 'type')`? In fact what is `type`?

yohanboniface: > Why the hasattr(cls, 'type')? In fact what is type?

No need for this check, it's only to be able to have abstract children class. (Update the example code.)

> To be honest, I find this harder to understand.

Seems to me that a good use case for metaclass, but I'm fine with your implementation too :)

ideascube/serveradmin/catalog.py
bochecha: Not verifying checksums seems like a giant security hole. How do you ensure you're installing the right content without checking the checksum?

To me, we should at the minimum:

* download **everything** over https
* verify the `catalog.yml` downloaded from the remote source
* verify the downloaded packages with a strong checksum (afaik sha256sum is still ok these days)

Currently we're not doing the first two, so the third one is not very useful. But that seems like an argument for doing the first two, rather than an argument for removing the third one. :smile:

ideascube/serveradmin/catalog.py
bochecha: That's a very good question. I would **think** that the `sha` object itself never grows (after all, a sha256 digest is always the same length, no matter the data it represents).

But that would need to be verified...

ideascube/serveradmin/management/commands/catalog.py
bochecha: Done in the latest push to this branch.

ideascube/serveradmin/management/commands/catalog.py
bochecha: Done in the latest push to this branch.

ideascube/serveradmin/management/commands/catalog.py
bochecha: Done in the latest push to this branch.

ideascube/serveradmin/management/commands/catalog.py
bochecha: What's a "🗸" ? (all I see is a Unicode square, because my font can't display this glyph)

yohanboniface: A unicode check :)
But we could just display "installed" then :)

ideascube/conf/base.py
bochecha: Using STORAGE_ROOT now in the latest push to this branch.

ideascube/serveradmin/management/commands/catalog.py
bochecha: Oh, so you meant two groups in the output!

Here I pretty much did what yum/dnf do, since I'm used to their output for quite a few years. :smile: 

One argument in favor of separating the two, is that we can show that `wikipedia.fr` is installed in its version `2015-08`, and it is also available but in a different, newer version. Oh hey, we need to update it!

ideascube/conf/base.py
```diff
@@ -283,3 +283,6 @@
 
 SESSION_COOKIE_AGE = 60 * 60  # Members must be logged out after one hour
 BACKUP_FORMAT = 'zip'  # One of 'zip', 'tar', 'bztar', 'gztar'
+
+CATALOG_CACHE_BASE_DIR = '/var/cache/ideascube/catalog'
+CATALOG_KIWIX_INSTALL_DIR = os.path.join(STORAGE_ROOT, '/kiwix')
```
yohanboniface: This will ends with '/kiwix' AFAIK, so it should be 'kiwix' only.

ideascube/serveradmin/management/commands/catalog.py
yohanboniface: Good point!
Also we will need a `list --upgradable` option :)

ideascube/conf/base.py
bochecha: Oops! Fixed in the latest push to this branch.

bochecha:
So I've been thinking some more about your suggestion to use a metaclass for the package/handler type registries, because I wasn't entirely happy with the way I did it.

And then I realized that my unhappiness was not really about the registry itself, but about the magic `__new__` method.

I think I might have been a bit too clever on this one. :smile: 

Originally it seemed like a great idea to be able to do this:

```python
pkg = Package('wikipedia.en', {'type': 'zippedzim'})
pkg = Package('kiwix', {'type': 'app'})
...
```

The API was simple, there was only one entry-point (the `Package` constructor) so that the code needing to create a new package didn't need to know about the registry, the various supported package types, etc... But then it's a bit magical and surprising.

And this morning I realized that this API allowed things like this:

```python
pkg = ZippedZim('kiwix', {'type': 'app'})
assert isinstance(pkg, App)
```

And that's just not right. :sweat: 

So I think I'm going to redo this part a bit, to make it simpler, and less surprising.

yohanboniface:
That's why I suggested to test on a "type" property before adding it to the registry, so we would add this property only for the needed class, and not for App for example.

bochecha:
> That's why I suggested to test on a "type" property before adding it to the registry, so we would add this property only for the needed class, and not for App for example.

I don't see why we wouldn't want to add it for App as well, though. An App would be just another type of package, with its own install/remove recipes.

To be honest I'm now leaning more towards a simple, static registry:

```python
PACKAGE_TYPES = {
    'app': App,
    'zippedzim': ZippedZim,
    }
```

And when we have a new package type to handle, we implement the new class and add it to the registry.

I went back and forth on this, and I'm starting to think that it's not complex enough to deserve a dynamic, automatic registry.

bochecha:
Ok, so with the latest push, I added tons of unit tests, which uncovered a few issues in some corner cases, that I've fixed.

I've rebased the branch on master, and splitted/merged a few commits so that it looks more like a logical progression overall.

The API is pretty much entirely covered by tests, there are still some missing ones for the management command: listing, installing, removing and updating packages.

ideascube/conf/base.py
```diff
@@ -302,3 +302,6 @@
 
 SESSION_COOKIE_AGE = 60 * 60  # Members must be logged out after one hour
 BACKUP_FORMAT = 'zip'  # One of 'zip', 'tar', 'bztar', 'gztar'
+
+CATALOG_CACHE_BASE_DIR = '/var/cache/ideascube/catalog'
```
yohanboniface: No need to reflect this in Debian package files?
IIRC, putting a path in Debian `dirs` will create it automatically (we could do so in python too), and in particular will ask the user what to do with the folder content when deinstalling the package.
But you are the packaging expert, not me, so I may have missed something :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,507 @@
...
+
+    def install(self, download_path, install_dir):
+        if not zipfile.is_zipfile(download_path):
+            os.unlink(download_path)
+            raise InvalidFile('{} is not a zip file'.format(download_path))
+
+        with zipfile.ZipFile(download_path, "r") as z:
+            names = z.namelist()
+            names = filter(lambda n: n.startswith('data/'), names)
+            z.extractall(install_dir, members=names)
```
yohanboniface: Real life zips will take a lot to unzip (47Go for English Wikipedia or such), particularly in Koombook. I wonder if we don't need a progress bar here too?</autopromo> :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,507 @@
...
+
+                except NoSuchPackage as e:
+                    if fail_if_no_match:
+                        raise e
+
+        return pkgs
+
+    def _get_handler(self, package):
+        try:
+            handlertype = package.handler
```
yohanboniface: I'd prefer let the Handler register themselves for being able to handle given types. Who handle which kind of content is a discretion of the local environment I'd say, not the remote package, that can be used in any other environment than ideascube. What do you think?

ideascube/serveradmin/catalog.py
yohanboniface: > I would think that the sha object itself never grows (after all, a sha256 digest is always the same length, no matter the data it represents).

I concur :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,507 @@
...
+            if pkg.id in self._catalog['installed']:
+                sys.stderr.write('{0.id} is already installed\n'.format(pkg))
+                sys.stderr.flush()
+                continue
+
+            download_path = self._fetch_package(pkg)
+            handler = self._get_handler(pkg)
+            sys.stdout.write('Installing {0.id}\n'.format(pkg))
+            handler.install(pkg, download_path)
+            used_handlers[handler.__class__.__name__] = handler
```
yohanboniface: Not that a big problem, but we are creating a new handler at each loop even if we were using the same on previous loop. Given that the `commit` is not instance specific, that's fine, but that *may* be a trap in the future if some handler would have some specific bits in their `commit` step.

ideascube/serveradmin/catalog.py
yohanboniface: My use case was hand writing catalog, as we will certainly do in the first time. And while the package manager is only a command line argument we don't really have security warning because we will be the only one to use it, from our office, etc.
But OK, let's keep that check because that's anyway where we want to go in the end. And let's see if it's any problem while doing a lot of thing by hand, and adapt it only in this case. :)

bochecha: > we will be the only one to use it, from our office, etc.

Over an insecure http connection which could very well be tampered with.

ideascube/serveradmin/catalog.py
bochecha: But then how do you know which handler to use for a given package?

Originally we had decided to separate the `type` and `handler` [1] package metadata because a given handler might be able to handle various package types, and a given package type might be handled by various handler, depending on the actual package.

---

[1] which we used to call `app`, but since there's also an `app` package type I found it confusing and renamed it

ideascube/conf/base.py
bochecha: You're right, I forgot that.

This is now done in the latest push to this branch.

ideascube/serveradmin/catalog.py
yohanboniface: Meeeeeh, please, there is not theoretical debate here, we all agree over the need for secure communications. We just need to be *pragmatic* over priorities and adapt on our current workflow.
Anyway, as as said: "let's keep this check", it's coded and that's ok. Let's proceed :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+
+class NoSuchPackage(Exception):
+    pass
+
+
+class InvalidHandlerType(Exception):
+    pass
+
+
+class Remote:
```
almet: Remote what ?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+                raise InvalidFile(
+                    'Remote file is missing a {} key: {}'.format(e, path))
+
+    def to_file(self, path):
+        with open(path, 'w') as f:
+            d = {'id': self.id, 'name': self.name, 'url': self.url}
+            f.write(yaml.safe_dump(d, default_flow_style=False))
+
+
+class Handler:
```
almet: should this be "base" handler?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def __init__(self):
+        default = os.path.join(settings.STORAGE_ROOT, self.name)
```
almet: what about passing "settings" in the constructor?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+        setting = 'CATALOG_{}_INSTALL_DIR'.format(self.name.upper())
+        self._install_dir = getattr(settings, setting, default)
+        self._installed = []
```
almet: nit: installed_packages ?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+        self._removed = []
```
almet: nit: removed_packages ?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def restart_service(self, name):
+        sys.stdout.write('Restarting service "{}"\n'.format(name))
+        try:
+            manager = SystemManager()
+            service = manager.get_service(name)
+
+        except NoSuchUnit:
+            # Service is not installed, give up.
```
almet: give up without any log?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+            pass
+
+        else:
+            manager.restart(service.Id)
+
+
+class Kiwix(Handler):
+
+    def commit(self):
+        sys.stdout.write('Rebuilding the Kiwix library\n')
```
almet: nit: use loggers / reporters instead?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def install(self, package, download_path):
+        package.install(download_path, self._install_dir)
+        self._installed.append(package)
+
+    def remove(self, package):
+        package.remove(self._install_dir)
+        self._removed.append(package)
+
+    def commit(self):
```
almet: not sure to understand what this "commit" thing does.

ideascube/serveradmin/catalog.py
almet: BaseServiceHandler?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+        setting = 'CATALOG_{}_INSTALL_DIR'.format(self.name.upper())
+        self._install_dir = getattr(settings, setting, default)
+        self._installed = []
+        self._removed = []
+
+    @property
+    def name(self):
+        return self.__class__.__name__.lower()
+
+    def install(self, package, download_path):
```
almet: nit: install_package

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+        package.install(download_path, self._install_dir)
+        self._installed.append(package)
+
+    def remove(self, package):
```
almet: nit: remove_package

ideascube/serveradmin/catalog.py
almet: Maybe finding a better name would help :)

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+        enabled = Path(self.root, 'sites-enabled', package.id)
+        server_name = '{subdomain}.{domain}'.format(subdomain=package.id,
+                                                    domain=settings.DOMAIN)
+        root = package.get_root_dir(self._install_dir)
+        with available.open('w') as f:
+            f.write(self.template.format(server_name=server_name, root=root))
+        try:
+            enabled.symlink_to(available)
+        except FileExistsError:
+            pass
```
almet: If the file exists, it doesn't mean that it links to the proper location. Should we check that as well here?

ideascube/serveradmin/catalog.py
```diff
@@ -0,0 +1,576 @@
...
+
+        except AttributeError:
+            # The class doesn't have a 'typename' attribute, don't register it
+            pass
+
+        return cls
+
+
+class Package(metaclass=MetaRegistry):
```
almet: Can you explain what's going on here in a comment?

almet:
Oh, I just realized this was merged already. I will stop commenting now. Let me know if that's helpful and I can continue the (really long) review. Thanks!

ideascube/serveradmin/catalog.py
yohanboniface: Oh, good catch, yeah :)

yohanboniface: Not sure though what to do when the target is not the expected one: in theory it can even be a folder with content we would not want to delete without the user agreement. Or it can be pointing to another site we want to keep (id clash between two catalog for example).
I've update the code to raise when the symlink exist and it's not pointing to the expected target, but that will need refinement for sure.

PR #283, opened 2016-02-11
Add epub type in mediacenter (fix #282)

### LINKED ISSUE #282: new item type in the media center
It would be great to add in the media center a new file type : "epub". Let's not forget ideascube is also used by Library Without Borders. It will allow the ereader to access directly to a list of epub in the media center with a predefined url. We can achieve this with tags though. But it makes sense to add a new file type.

yohanboniface:
Fix #282


yohanboniface:
@bochecha r?

yohanboniface:
erf

bochecha:
Sure, :+1:

PR #300, opened 2016-02-26
Remove STORAGE_ROOT var from kb.py

fheslouin:
I think we can now remove the STORAGE_ROOT variable from the config file, indeed, we are now using a symlink from /var/ideascube pointing to /media/hdd/ideascube (koombook only). This is also fixing the process of installation as I had rights conflicts with the SQLite DB. I have tested it on KoomBook and it's working fine!


yohanboniface:
Cool :)
Let's keep STORAGE_ROOT in default config anyway, it may be used in other scenarios :)

PR #301, opened 2016-02-26
Increase number of attemps on wifi module

fheslouin:
I would suggest to increase the number of attempts to 12. Indeed, the connexion sometime FAILED because the number is too low. After several test it seems that 12 is a good enough value. There is still some random FAILING but it is better.


bochecha:
Just to be sure we all understand the implication: this means we'll wait up to 12 seconds before deciding the connection won't succeed, and get a page back to the user.

Out of curiosity, anybody knows what the timeout for web browsers? I mean, how long do they wait before they decide the server just won't respond?

PR #303, opened 2016-02-26
Add static-folder package type

yohanboniface:
Included in this PR:

- CATALOG_XXX_INSTALL_DIR is no more set in the default settings, but instead defaults to STORAGE_ROOT/xxx; CATALOG_XXX_INSTALL_DIR can still be used to override the default

- the relation Package <=> Handle is now ported directly by the package class; the rationale is that I don't want the catalog to be responsible of defining which local application should handle which content, but instead let the local system know that

- Handler.commit now takes two parameters (that can be None): `removed` and `installed`, that are the lists of package just installed or removed. Rationale is that at commit needs to be idempotent, and at commit time, Nginx handler wasn't able to know which packages to consider (and it has no reference to the Catalog).

- a StaticFolder package class has been added

- a Nginx handler class has been added

I've added only basic tests for now. I'll add more tomorrow, but now I need to run on the camp, then move the box elsewhere, then take the ferry to Lesbos ;)

I'd love to have a review yet though :)

cc @bochecha


ideascube/serveradmin/catalog.py
```diff
@@ -60,6 +60,111 @@ def to_file(self, path):
             f.write(yaml.safe_dump(d, default_flow_style=False))
 
 
+class Handler:
```
yohanboniface: Moved before Package class to be able to reference handlers as class properties.

Issue #326, opened 2016-03-16
Catalog failure when rerunning a multiple install


yohanboniface:
```
Traceback (most recent call last):
  File "/usr/share/python/ideascube/bin/django-admin", line 11, in <module>
    sys.exit(execute_from_command_line())
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/management/commands/catalog.py", line 105, in handle
    options['func'](options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/management/commands/catalog.py", line 151, in install_packages
    self.catalog.install_packages(options['ids'])
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 444, in install_packages
    handler.install(pkg, download_path)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 78, in install
    package.install(download_path, self._install_dir)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 247, in install
    os.rename(path, new)
OSError: [Errno 39] Directory not empty: '/var/ideascube/kiwix/data/index/ubuntudoc_fr_all_2015-12.zim.idx' -> '/var/ideascube/kiwix/data/index/ubuntudoc.fr.zim.idx'
```

[...]

PR #327, opened 2016-03-16
Catalog: more defensive approach

yohanboniface:
This PR does four things:

- when installing a zim, always override any preexisting content (and do not fail)
- persist cache as soon as a package is marked as installed/removed
- download all packages then install all packages instead of doing download/install by package
- catch download/install/remove errors and continue to the next id to be processed (instead of failing)

cf #326

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -20,6 +20,14 @@ from .systemd import Manager as SystemManager, NoSuchUnit
 from ..utils import printerr
 
 
+def rm(path):
+    try:
+        os.unlink(path)
+
+    except IsADirectoryError:
+        shutil.rmtree(path)
+
+
 class InvalidFile(Exception):
     pass
 
@@ -253,6 +261,8 @@ class ZippedZim(Package):
 
         for path in glob(os.path.join(datadir, '*', '{}*'.format(zimname))):
             new = path.replace(zimname, '{0.id}.zim'.format(self))
+            if Path(new).exists():
+                rm(new)
             os.rename(path, new)
 
     def remove(self, install_dir):
@@ -260,11 +270,7 @@ class ZippedZim(Package):
         datadir = os.path.join(install_dir, 'data')
 
         for path in glob(os.path.join(datadir, '*', zimname)):
-            try:
-                os.unlink(path)
-
-            except IsADirectoryError:
-                shutil.rmtree(path)
+            rm(path)
 
 
 class StaticSite(Package):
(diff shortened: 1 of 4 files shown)


yohanboniface:
@bochecha r? :)

ideascube/serveradmin/catalog.py
```diff
@@ -253,6 +261,8 @@ def install(self, download_path, install_dir):
 
         for path in glob(os.path.join(datadir, '*', '{}*'.format(zimname))):
             new = path.replace(zimname, '{0.id}.zim'.format(self))
+            if Path(new).exists():
+                rm(new)
             os.rename(path, new)
```
bochecha: I would do it this way instead:

```python
try:
    os.rename(path, new)

except IsADirectoryError:
    rm(new)
    os.rename(path, new)
```

Because it's easier to ask for forgiveness than permission, and because it removes the need to create a `Path` object just to see if the path exists.

Also, it's not a problem if the path exists, it only is one if it's a directory. (if it's a file, it gets written over)

yohanboniface: os.rename is giving when the dir is not empty in this pattern:
```
OSError: [Errno 39] Directory not empty
```

bochecha: Not sure what you mean.

```
>>> os.listdir()
['CONTRIBUTING.md', 'requirements.txt', 'requirements-rtfd.txt', 'requirements-dev.txt', 'extras', 'storage', 'Makefile', 'README.md', '.travis.yml', 'AUTHORS.md', '.coveragerc', 'mkdocs.yml', 'setup.py', 'docs', 'CHANGELOG', 'LICENSE.md', 'pytest.ini', 'conftest.py', '.cache', '.gitignore', '__pycache__', '.git', 'manage.py', 'MANIFEST.in', 'debian', 'ideascube.egg-info', '.coverage', 'ideascube', '.tx']
>>> 
>>> os.listdir('extras')
['polkit-debian', 'nginx', 'uwsgi', 'bin', 'pam']
>>> 
>>> try:
...     os.rename('LICENSE.md', 'extras')
... except IsADirectoryError:
...     rm('extras')
...     os.rename('LICENSE.md', 'extras')
... 
>>> 
```

Seems to work fine?

bochecha:
Not sure what the last commit has to do with this PR? (the one about migrations)

But other than that, this is pretty good, should make the package manager more robust. :smile:

ideascube/serveradmin/catalog.py
yohanboniface: Weird.
Try to run tests, `test_kiwix_does_not_fail_if_files_already_exist` is failling with the error I paster above. I'll investigate a bit more.

PR #329, opened 2016-03-18
Use print instead of sys.std(out|err).write.

mgautierfr:
Just remove sys.std(out|err).write calls to print and printerr.

No functional changes.

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -18,6 +18,14 @@ import yaml
 
 from .systemd import Manager as SystemManager, NoSuchUnit
 
+# We do not use functool.partial cause we want to mock stderr for unittest
+# If we use partial we keep a ref to the original sys.stderr and output is not
+# captured.
+def printerr(*args, **kwords):
+    kwords['file'] = sys.stderr
+    kwords['flush'] = True
+    return print(*args, **kwords)
+
 
 class InvalidFile(Exception):
     pass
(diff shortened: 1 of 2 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -18,6 +18,14 @@
 
 from .systemd import Manager as SystemManager, NoSuchUnit
 
+# We do not use functool.partial cause we want to mock stderr for unittest
+# If we use partial we keep a ref to the original sys.stderr and output is not
+# captured.
+def printerr(*args, **kwords):
```
yohanboniface: why kwords and not kwargs? :)

yohanboniface: Also I think printerr could be on a util.py module (even in the root one), so we can reuse it.

mgautierfr: Because ! :)
But agreed, kwargs is better.
+1 also for moving printerr into util.py.

ideascube/utils.py
```diff
@@ -6,3 +6,13 @@ class classproperty(property):
     """
     def __get__(self, cls, owner):
         return self.fget.__get__(None, owner)()
+
+
+# We do not use functool.partial cause we want to mock stderr for unittest
+# If we use partial we keep a ref to the original sys.stderr and output is not
+# captured.
+def printerr(*args, **kwargs):
+    import sys
```
yohanboniface: Why not on top? :)

mgautierfr: To not pollute the namespace.
This seems to me "contained enough" to have the import inside the function.
(And in a certain way, utils.py is not intended to import sys (even if it does *if we use printerr*))

But we can move it to top and (eventually) use __all__

PR #330, opened 2016-03-18
Mediacenter catalog

mgautierfr:
There are still opened issues: 

- Only the link to the delete/edit content form is hidden if package_id is not empty.
  The forms are still accessible. I think it's enough and better to keep a way to edit the content even if it comes from a package.
- The document file is still copied by django in CATALOG_MEDIACENTER_ROOT. So there is still a duplication of content but it should be fixed by furthers steps.
- The commit on removed packages remove the document from the mediacenter db but keep it in CATALOG_MEDIACENTER_ROOT. Same as previous point, this should be no relevant when no copy is made.


ideascube/mediacenter/migrations/0007_auto_20160318_1510.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('mediacenter', '0006_auto_20160310_2005'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='document',
+            name='package_id',
+            field=models.CharField(blank=True, max_length=100, verbose_name='package_id'),
```
yohanboniface: Just a reminder: let's check on max_length at some point. 100 sounds OK, but maybe we could have a bit more margin.

yohanboniface: Just a reminder: let's check on max_length at some point. 100 sounds OK, but maybe we could have a bit more margin.

ideascube/mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -19,10 +19,16 @@
 {% endblock twothird %}
 {% block third %}
     {% if user.is_staff %}
-        <ul class="card tinted admin">
-            <li>{% fa 'pencil' 'fa-fw' %}<a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
-            <li>{% fa 'trash' 'fa-fw' %}<a href="{% url 'mediacenter:document_delete' pk=document.pk %}">{% trans 'Delete' %}</a></li>
-        </ul>
+        {% if document.package_id %}
+            <span class="card admin">
+               {% trans 'Document installed by a package. Cannot edit or delete this document.' %}
```
yohanboniface: Let's add the package id too (name would be better actually, but for now let's use id).

ideascube/mediacenter/tests/test_views.py
```diff
@@ -104,6 +104,20 @@ def test_staff_can_edit_document(staffapp, video):
     assert Document.objects.get(pk=video.pk).title == title
 
 
+def test_staff_detail_page_has_no_edit_delete_link_package_id(staffapp):
+    document = DocumentFactory(package_id="foo")
+    response = staffapp.get(reverse('mediacenter:document_detail',
+                                    kwargs={'pk': document.pk}), status=200)
+    assert "Document installed by a package." in response.content.decode()
```
yohanboniface: Also make sure the link itself is not here, just in case :)

ideascube/serveradmin/catalog.py
```diff
@@ -176,6 +185,97 @@ def unregister_site(self, package):
         except FileNotFoundError as e:
             printerr(e)
 
+class MediaCenter(Handler):
+    def commit(self):
+        for pkg in self._removed:
+            self.remove_pkg(pkg)
```
yohanboniface: Handler already have an `install` method that is supposed to install the package, why not using that?

ideascube/serveradmin/catalog.py
```diff
@@ -176,6 +185,97 @@ def unregister_site(self, package):
...
+        for pkg in self._installed:
+            self.install_pkg(pkg)
+        super().commit()
+
+    def _install_media(self, media_info, install_dir, package_id):
+        #ensure that mandatory fields are set
```
yohanboniface: Nit: space after '#', upper first letter, and dot at the end :)

ideascube/serveradmin/catalog.py
```diff
@@ -176,6 +185,97 @@ def unregister_site(self, package):
...
+        for entry in ('title', 'summary', 'path', 'credits'):
```
yohanboniface: Why not relying on DocumentForm to check for mandatory fields?

ideascube/serveradmin/catalog.py
```diff
@@ -176,6 +185,97 @@ def unregister_site(self, package):
...
+            if 'preview' in media_info:
+                preview = media_info['preview']
+                try:
+                    p = Path(install_dir, preview).open('rb')
+                    files['preview'] = File(p, name=preview)
+                except OSError:
+                    pass
+
+            self._save_media(metadata, files)
+        finally:
```
yohanboniface: Not sure we need nested try for that: we could use only one with try/except/else/finally, no?

ideascube/serveradmin/catalog.py
```diff
@@ -176,6 +185,97 @@ def unregister_site(self, package):
...
+        root = pkg.get_root_dir(self._install_dir)
+        manifestfile = Path(root, 'manifest.yml')
+        with manifestfile.open('r') as m:
+            manifest = yaml.safe_load(m.read())
+
+        for media in manifest['medias']:
+            try:
+                self._install_media(media, root, pkg.id)
+            except:
+                # What to do here ? revert/stop/continue ?
```
yohanboniface: I'd say print and continue to the next one for now.

ideascube/serveradmin/catalog.py
```diff
@@ -176,6 +185,97 @@ def unregister_site(self, package):
...
+                raise # easiest solution
+
+    def remove_pkg(self, pkg):
+        #Easy part here. Just delete document from package
+        Document.objects.filter(package_id = pkg.id).delete()
```
yohanboniface: Nit: spaces around the '=' ;)

ideascube/mediacenter/templates/mediacenter/document_detail.html
mgautierfr: Agree.

ideascube/mediacenter/tests/test_views.py
mgautierfr: Agree.

ideascube/serveradmin/catalog.py
mgautierfr: I've made like Nginx/StaticSite handler/package.

That is Package that has an `install` method, who physicaly install files on hd.
The `commit` handler's method update config/db.

yohanboniface: https://github.com/ideascube/ideascube/blob/master/ideascube/serveradmin/catalog.py#L80 ;)

ideascube/serveradmin/catalog.py
mgautierfr: Well, why import_medias command doesn't rely on DocumentForm to check for mandatory fields ? :)
But you're right, it would be better to use DocumentForm.

ideascube/serveradmin/catalog.py
mgautierfr: The first/big try/catch replace the `with open(...) as f:` pattern.  
But I'm not fan of nested `with ...` that end to different calls to `self._save_media` depending of the context (as it is made in import_medias command)

We could somehow merge the first nested `try/except OSError` in the big one. But not sure for the second one.

Btw, we will not keep this code cause we don't what to copy the file but use the already installed one by the catalog. Not sure it worst the change here if we drop the code latter.

ideascube/serveradmin/catalog.py
mgautierfr: At least print a warning.
This could lead to package installed but not all medias accessible from mediacenter in case of errors.

ideascube/mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -19,10 +19,19 @@
 {% endblock twothird %}
 {% block third %}
     {% if user.is_staff %}
-        <ul class="card tinted admin">
-            <li>{% fa 'pencil' 'fa-fw' %}<a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
-            <li>{% fa 'trash' 'fa-fw' %}<a href="{% url 'mediacenter:document_delete' pk=document.pk %}">{% trans 'Delete' %}</a></li>
-        </ul>
+        {% if document.package_id %}
+            <span class="card admin">
+               {% trans 'Cannot edit or delete this document.' %}
```
yohanboniface: Use `blocktrans` instead when you need variable inside :)

ideascube/serveradmin/catalog.py
```diff
@@ -286,6 +293,93 @@ def remove(self, install_dir):
...
+        Document.objects.filter(package_id=self.id).delete()
+        super().remove(install_dir)
+
+    def install(self, download_path, install_dir):
+        super().install(download_path, install_dir)
+        print('Adding medias to mediacenter database.')
+        root = self.get_root_dir(install_dir)
+        manifestfile = Path(root, 'manifest.yml')
+        with manifestfile.open('r') as m:
+            manifest = yaml.safe_load(m.read())
```
yohanboniface: Let's have a unittest to be sure when the yaml is invalid we are not crashing but just skipping current package.

ideascube/serveradmin/catalog.py
```diff
@@ -286,6 +293,93 @@ def remove(self, install_dir):
...
+
+        for media in manifest['medias']:
+            try:
+                self._install_media(media, root)
+            except:
```
yohanboniface: Let's also display the error message, given you are providing it when raising?

ideascube/serveradmin/catalog.py
```diff
@@ -286,6 +293,93 @@ def remove(self, install_dir):
...
+
+    def _install_media(self, media_info, install_dir):
+        files = {'original': None, 'preview': None}
+
+        try:
+            media_info['title'] = title = smart_truncate(media_info['title'])
+        except KeyError:
+            raise InvalidPackageContent('Missing {} in {}'.format(entry, media_info))
+
+        if not 'lang' in media_info:
```
yohanboniface: if xxx not in yyy ;)

(I'm doing the same mistake again and again myself ;) )

ideascube/serveradmin/catalog.py
```diff
@@ -286,6 +293,93 @@ def remove(self, install_dir):
...
+            media_info['lang'] = settings.LANGUAGE_CODE
+
+        kind = media_info.get('kind')
+        if not kind or not hasattr(Document, kind.upper()):
+            content_type, _ = mimetypes.guess_type(original)
+            media_info['kind'] = guess_kind_from_content_type(content_type) or Document.OTHER
+
+        instance = Document.objects.filter(title=title, kind=kind).last()
+        if instance:
+            raise InvalidPackageContent('Document {} already exists'.format(title))
```
yohanboniface: If the package_id is not the same, I think we can tolerate having twice the same title with the same kind.

yohanboniface: BTW, if it already exists with the same package_id, let's override anyway.

mgautierfr:
This PR is replaced by PR #342 .

Issue #333, opened 2016-03-21
Only categories with file should appear


barbayellow:
In the media center, empty categories should be hidden. Example : if there are no pdf files, then the pdf cat should not appear in the right col.

yohanboniface:
cc @mgautierfr this is to be taken into account on the "search based navigation" refactor :)

yohanboniface:
This means live requesting which values are available for the current search. That could work for kind or lang, but that will be a bit harder to achieve this for tags, given that it's a multi value, and SQLite doesn't know about multivalues column.

mgautierfr:
This is partly fix by PR #335.
Only kind and lang are hidden when empty.

For tags we may assume that if a tag is present in the db, it is pointing to at least a content (?)
We could filter tags to show only tags pointing to the displayed documents (if they are already filtered) but as @yohanboniface said, it a bit more complex. This is possible but I think we could delay it.

yohanboniface:
Fixed with #335

PR #335, opened 2016-03-24
Mediacenter search

mgautierfr:
Missing :
 - Unit test.
 - Better view to show user what filters are used.


ideascube/templatetags/ideascube_tags.py
```diff
@@ -130,3 +130,36 @@ def paginate(request, **kwargs):
...
         get[key] = value
     return '?{}'.format(get.urlencode())
+
+
+class AddGetParameter(template.Node):
+    def __init__(self, values):
+        self.values = values
+
+    def render(self, context):
+        req = template.resolve_variable('request',context)
```
yohanboniface: Missing space. Flake 8? :)

ideascube/templatetags/ideascube_tags.py
```diff
@@ -130,3 +130,36 @@ def paginate(request, **kwargs):
...
+            if key.endswith('s'):
+                for v in value:
+                    v = template.Variable(v).resolve(context)
+                    params.appendlist(key, v)
+            else:
+                params[key] = template.Variable(value).resolve(context)
+        return '?%s' %  params.urlencode()
+
+
+@register.tag()
```
yohanboniface: Can't we go with a simple_tag here that would accept *args?

yohanboniface: About naming, I'd go instead for `add_query_string` or `add_qs`.
`get` is a bit generic imho to have a clear meaning ;)

ideascube/templates/ideascube/includes/tag_cloud.html
```diff
@@ -1,10 +1,15 @@
-{% load i18n %}
+{% load i18n ideascube_tags %}
 {% spaceless %}
 <div class="card tinted">
     <h4><span class="theme learn">{% trans "browse" %}</span> {% trans "Themes" %}</h4>
     {% for tag in tags %}
         {% if tag.slug %}
-            <a href="{% url url tag=tag.slug %}">{{ tag }}</a>{% if not forloop.last %},{% endif %}
+            {% if use_get %}
```
yohanboniface: Can't we simply always use get now?

ideascube/search/templates/search/box.html
```diff
@@ -4,6 +4,11 @@
 <div class="card tinted search">
     <form action="{% if action %}{{ action }}{% else %}{% url "search:search" %}{% endif %}" id="search">
         <input name="q" type="text" placeholder="{% trans 'search' %}" value="{{ q }}" />
+        {% if default_values %}
```
yohanboniface: Seems like the if is not needed.

ideascube/mediacenter/views.py
```diff
@@ -16,49 +16,33 @@
...
 
     def get_queryset(self):
         qs = super(Index, self).get_queryset()
         query = self.request.GET.get('q')
-        if query:
-            return qs.search(query)
+        kind = self.request.GET.get('kind')
+        lang = self.request.GET.get('lang')
+        tags = self.request.GET.getlist('tags')
+        if query or kind or lang or tags:
```
yohanboniface: `if any(x, y, z)`? :)

ideascube/mediacenter/urls.py
```diff
@@ -4,7 +4,6 @@
 
 urlpatterns = [
     url(r'^$', views.index, name='index'),
-    url(r'^tag/(?P<tag>[\w\-_]+)/$', views.by_tag, name='by_tag'),
```
yohanboniface: Open question: do we need to take care of retrocompatibility of this URL?
I'd say no, but let's keep that in mind.

ideascube/templates/ideascube/includes/tag_cloud.html
mgautierfr: It would be preferable, but the "old" way is still use by other apps.
We should adapt all other apps and then always use get.

ideascube/templatetags/ideascube_tags.py
mgautierfr: Yes, I will try to use simple_tag instead.
Agreed on the tag name.

yohanboniface:
Just thought we'll need a `reindex` on every server where we will update the version, once this merged.

Maybe we should add the `reindex` command on debian post_install for every update? cc @bochecha

bochecha:
> Maybe we should add the `reindex` command on debian post_install for every update?

Is it interactive?

Can it fail, leaving the system in "who knows what state"?

yohanboniface:
> Is it interactive?

Good question, but no :)

> Can it fail, leaving the system in "who knows what state"?

I'd say not really. That's "only" the search after all.

bochecha:
So, if it's not interactive and failures wouldn't be dramatic, then it certainly **could** go in the `post_install` script, yes.

I guess if we're doing the `migrate` and `collectstatic` in `postinst`, then there's no reason not to do this one as well.

yohanboniface:
Also do you feel #333 could be added here or should we keep it for a following iteration?

ideascube/mediacenter/models.py
```diff
@@ -38,16 +38,16 @@ class Document(SearchMixin, TimeStampedModel):
...
-        (IMAGE, _('image')),
-        (AUDIO, _('sound')),
-        (VIDEO, _('video')),
-        (PDF, _('pdf')),
-        (TEXT, _('text')),
-        (EPUB, _('epub')),
-        (APP, _('app')),
-        (OTHER, _('other')),
-    )
+    KIND_CHOICES = {
```
yohanboniface: I think we'd better keep a tuple as `KIND_CHOICES` and have a dict generated from it:

`KIND_DICT = dict(KIND_CHOICES)`

This for two reasons:
- be convention, we expect something named `XXX_CHOICES` to be a tuple
- more important: when using a dict, the order of the generated select will be random, which is not user friendly ;)

ideascube/mediacenter/templates/mediacenter/index.html
```diff
@@ -21,13 +21,43 @@
         </div>
     {% endif %}
     {% url 'mediacenter:index' as index %}
+    {% if q or kind or lang or tags %}
```
yohanboniface: All this will need a bit of CSS love ;)
I think the classic tag with coloured background and a 'x' to remove it will be ok.

ideascube/templates/ideascube/includes/tag_cloud.html
yohanboniface: >  We should adapt all other apps and then always use get.

Let's do that unless there is a good reason not to :)

ideascube/templatetags/ideascube_tags.py
```diff
@@ -65,7 +67,7 @@ def tag_cloud(url, model=None, limit=20, tags=None):
             qs = qs.filter(taggit_taggeditem_items__content_type=content_type)
         qs = qs.annotate(count=Count('taggit_taggeditem_items__id'))
         tags = qs.order_by('-count', 'slug')[:limit]
-    return {'tags': tags, 'url': url}
+    return {'tags': tags, 'url': url, 'use_get': use_get, 'request':context.request}
```
yohanboniface: pep8 ;)

ideascube/templatetags/ideascube_tags.py
```diff
@@ -130,3 +132,37 @@ def paginate(request, **kwargs):
...
         get[key] = value
     return '?{}'.format(get.urlencode())
+
+
+@register.simple_tag(takes_context=True)
+def add_qs(context, **kwargs):
+    req = template.resolve_variable('request', context)
+    params = req.GET.copy()
+    for key, value in kwargs.items():
+        if key.endswith('s'):
```
yohanboniface: Why not using a `list` instead?

yohanboniface:
For the record, we just discussed about the need of "reindexing" when installing this new app version: maybe the best option is to create a dedicated migration file that just run the `reindex` command.

ideascube/mediacenter/views.py
```diff
@@ -10,55 +10,48 @@
...
-        context['q'] = self.request.GET.get('q', '')
+        for key in ('q', 'kind', 'lang'):
+            context[key] = self.request.GET.get(key)
+        context['tags'] = self.request.GET.getlist('tags')
+        default_values = {k: context[k] for k in ('kind', 'lang', 'tags')
+                          if context[k]}
+        context['default_values'] = default_values
+        context['not_empty_kinds'] = [
+            (kind, label) for kind, label in Document.KIND_CHOICES
+            if len(Document.objects.filter(kind=kind))
```
yohanboniface: I think the idea here is to have this dynamic according to current request: so if I'm filtering on, say, tag "plane", I want to see only kinds that have at list one Document with "plane" tag, and so on.
The subtle part is that if we are for example already filtering by some kind, we want to see all other kinds available for the other filters (tags, lang).
So we need to do something like (example for kind, but should be the same with lang):
- get the filters
- remove the "kind" filter if any
- issue a group by on the search table keeping the other filters
- return those kinds

Also, at this point, we certainly want to have a `FilterableView` mixin to factorize this between blog, mediacenter and library.

yohanboniface:
Note: merged with master and pushed some very small changes while reviewing.

yohanboniface:
Also, we need a migration to reindex content.

ideascube/mediacenter/views.py
yohanboniface: @mgautierfr this point does not seem to be adressed, does it?

ideascube/templatetags/ideascube_tags.py
yohanboniface: Is that question answered somewhere else?

mgautierfr: We use add_qs and remove_qs in django template.
But there is no easy way to create a list from a single element in django template.

However, we do not use this functionality (we provide a single element) so we may simply remove the functionality and assume is always a single element.
What do you think ?

ideascube/mediacenter/views.py
mgautierfr: Done :)

ideascube/templatetags/ideascube_tags.py
yohanboniface: Yes, remove for now :)

ideascube/views.py
```diff
@@ -70,8 +70,14 @@ class ByTag(ListView):
...
     paginate_by = 20
 
+    def get_context_data(self, **kwargs):
+        context = super(ByTag, self).get_context_data(**kwargs)
+        context['tags'] = self.request.GET.getlist('tags')
+        return context
+
     def get_queryset(self):
-        return TaggedItem.objects.filter(tag__slug=self.kwargs['tag'])
+        tags = self.request.GET.getlist('tags')
```
yohanboniface: Why are we using names and not `slug` anymore?

ideascube/mixins.py
```diff
@@ -9,19 +9,43 @@
 from taggit.models import Tag
 
 
-class ByTagListView(ListView):
+class FilterableViewMixin:
+    def _search_for_attr_from_context(self, attr, context):
+        search_context = {k:context[k] for k in ('q', 'kind', 'lang', 'tags')
+                            if k != attr}
```
yohanboniface: Shouldn't we filter out None values, something like `if k!= attr and context[k]`?

ideascube/mediacenter/views.py
```diff
@@ -9,56 +9,40 @@
...
+            if kind in available_kinds
+            ]
+
+        # Do the same for langs
+        available_langs = self._search_for_attr_from_context('lang', context)
+        context['not_empty_langs'] = [
+            (lang, label) for lang, label in settings.LANGUAGES
+            if lang in available_langs
+            ]
+        return context
```
yohanboniface: I guess something prevented to do the same for tags?

ideascube/mediacenter/templates/mediacenter/index.html
```diff
@@ -21,13 +21,25 @@
...
-        {% for kind, label in view.model.KIND_CHOICES %}
-            <a href="?kind={{ kind }}">{{ label }}</a>
+        {% for kind, label in not_empty_kinds %}
+            <a href="{% add_qs kind=kind %}">{{ label }}</a>
             {% if not forloop.last %}⚫{% endif  %}
         {% endfor %}
     </div>
-    {% tag_cloud url='mediacenter:by_tag' model=view.model %}
+    {% endif %}
+    {% if not_empty_langs|length > 1 %}
```
yohanboniface: nit: not a big fan a double negated naming (not + empty), I'd go for "available_xxx" instead :)

yohanboniface:
Weird enough, seems that when I create new objects they are not indexed, but when running `reindex` it's OK. I'll investigate this.

On your side, can you make sure the migration command is OK? When I've switched to the branch and run the migrate command, it seemed to me that the objects were indexed correctly. I remember missing the kind/lang filters on the search page. I ran a `reindex` and it was OK.

yohanboniface:
For the record, diff of the POC we were discussing in video:

```
diff --git a/ideascube/mediacenter/views.py b/ideascube/mediacenter/views.py
index 4276c5a..91366bc 100644
--- a/ideascube/mediacenter/views.py
+++ b/ideascube/mediacenter/views.py
@@ -36,6 +36,14 @@ class Index(FilterableViewMixin, ListView):
             (lang, label) for lang, label in settings.LANGUAGES
             if lang in available_langs]
 
+        # Do the same for tags
+        available_tags = self._search_for_attr_from_context('lang', context)
+        ['|history|france|', '|plane|pouet|']
+        for slugs in available_tags:
+            for slug in slugs.split('|'):
+                context['available_tags'].append(Tag.objects.get(slug=slug).name)
+                # do a request to get the tag name
+
         return context
 
 index = Index.as_view()
diff --git a/ideascube/mixins.py b/ideascube/mixins.py
index c9a4394..029640a 100644
--- a/ideascube/mixins.py
+++ b/ideascube/mixins.py
@@ -18,6 +18,7 @@ class FilterableViewMixin:
         existing_attrs = set(self.model.objects
                                  .search(**search_context)
                                  .values_list(attr, flat=True))
+        return Search.objects.filter(**search_context).values_list(attr, flat=True)
         return existing_attrs
```

yohanboniface:
Yay!

PR #341, opened 2016-04-06
Packaging improvements

bochecha:
So I sat down with @didrocks on Sunday^w^wa few weeks ago, during the [Journées du Logiciel Libre](http://www.jdll.org/) (awesome event by the way, highly recommended) and he had a look at our Debian packaging.

Of course, he found "a few" problems. :stuck_out_tongue_winking_eye: 

This pull request fixes them.

Note that it is now ready for review, it isn't WIP any more. Hurray for hackfests giving you the time to actually finish stuff. \o/


debian/ideascube.postinst
```diff
@@ -31,9 +31,6 @@ case "$1" in
         ln -fs /etc/nginx/sites-available/$DPKG_MAINTSCRIPT_PACKAGE /etc/nginx/sites-enabled/$DPKG_MAINTSCRIPT_PACKAGE
         echo "Linking uWSGI app"
         ln -fs /etc/uwsgi/apps-available/$DPKG_MAINTSCRIPT_PACKAGE.ini /etc/uwsgi/apps-enabled/$DPKG_MAINTSCRIPT_PACKAGE.ini
-        echo "Restarting Nginx and uWSGI"
-        systemctl restart nginx.service uwsgi.service
-        systemctl enable uwsgi.service
```
fheslouin: All this can be done within Ansiblecube

extras/uwsgi/uwsgi.service
```diff
@@ -1,13 +0,0 @@
...
-[Service]
-ExecStart=/usr/bin/uwsgi --emperor /etc/uwsgi/apps-enabled --logto /var/log/uwsgi/access.log
-RuntimeDirectory=uwsgi
-Restart=always
-KillSignal=SIGQUIT
-Type=notify
-NotifyAccess=all
-
-[Install]
-WantedBy=multi-user.target
```
fheslouin: What will start the service then ? Is there any sysvinit script that we can rely on ? 

I'm using this unit to start other services working with uwsgi, plus I use the "after" options to start Uwsgi service after MongoDB, this is pretty efficient.

fheslouin:
Great job !
I created an issue on [Ansiblecube](https://github.com/ideascube/ansiblecube/issues/24) to adapt ansible to the new packaging

extras/uwsgi/uwsgi.service
bochecha: Yes, Debian already provides a sysvinit script.

You can have services use `After=uwsgi.service` even if uwsgi is in fact started by a sysvinit script.

And if you want uwsgi to start after mongodb, then you can continue doing either of these:

1. drop in a `mongodb.service.d/whatever.conf` file that simply adds the required `Before=uwsgi.service` line
2. drop in a `uwsgi.service.d/whatever.conf` file that simply adds the required `After=mongodb.service` line

And it will just work.

I'm not sure why you would want any ordering between mongodb and uwsgi, since AFAIK ideascube doesn't use mongodb, but you can, there's no issue with that. However, this isn't ideascube responsibility IMHO, so you just have to do it yourself.

debian/ideascube.postinst
bochecha: But I don't think it should.

You should **reload** nginx and uwsgi, there's no reason to **restart** them.

And it should only be needed on first install, I'm pretty sure subsequent installs don't need it. (see the `touch-reload` I'm adding to the uwsgi config file in another commit)

extras/uwsgi/uwsgi.service
fheslouin: Ok I can choose the first option to make sure that uwsgi is starting after MongoDB. 
Indeed Ideascube does not use MongoDB, but I was sharing this unit with an other web app, I will take care of this in Ansiblecube

bochecha: Do note that you don't need the whole `uwsgi.service` unit file.

You only need to add a `/etc/systemd/system/uwsgi.service.d/after-mongodb.conf` file, containing:

```ini
[Unit]
After=mongodb.service
```

And that's it!

extras/uwsgi/ideascube.conf
```diff
@@ -1,5 +0,0 @@
-[Service]
-User=ideascube
-Group=ideascube
-PAMName=ideascube
-Environment=SHELL=/bin/bash
```
bochecha: @fheslouin In another comment you mention that you were sharing that unit with another webapp.

That means your other webapp was running as the `ideascube` user, because all of uWSGI was running as that user.

But with the removal of the file above, now uWSGI will run normally as root (only the ideascube app runs as the `ideascube` user).

So if you want your other webapp to run as a specific non-root user, you'll need to adapt its configuration.

See the `ideascube.ini` configuration file, it includes the `uid` and `gid` parameters necessary for that.

But please don't add the above removed file back into Ansiblecube, that never was necessary, and we shouldn't have done it.

fheslouin: Good point ! The other web app are already using the uid and gid in their *.ini, so it should work as good as before

extras/uwsgi/uwsgi.service
fheslouin: Understood

debian/ideascube.postinst
fheslouin: Ok, I can lunch a reload at first install only

extras/uwsgi/uwsgi.service
letompouce: Since we don't need to deal with uwsgi anymore, I picked the second option and bloated the mongodb roles instead.
https://github.com/ideascube/ansiblecube/tree/issue-25-uwsgi-cleanup

debian/changelog
```diff
@@ -46,7 +46,7 @@ ideascube (0.9.3) unstable; urgency=low
 
   * Missing import in TV5 Monde conf
 
- -- Yohan Boniface <hi@yohanboniface.me>  Tue, 30 Mar 2016 18:10:00 +0100
+ -- Yohan Boniface <hi@yohanboniface.me>  Wed, 30 Mar 2016 18:10:00 +0100
```
yohanboniface: :p

debian/control
```diff
@@ -1,19 +1,45 @@
...
+  ${python3:Depends},
+  ${misc:Depends},
+  ${shlibs:Depends},
+  python3,
+Description: Library server
+ Ideascube is a complete library management server. It allows librarians to
+ manage their content, the users, and more. It also allows installing more
+ content and applications from various repositories, for example offline
+ copies of Wikipedia.
+ .
```
yohanboniface: Is that a dot?

debian/ideascube.postinst
```diff
@@ -0,0 +1,27 @@
...
+
+set -e
+
+case "$1" in
+    configure)
+        if ! getent passwd "$DPKG_MAINTSCRIPT_PACKAGE" > /dev/null 2>&1 ; then
+            adduser --quiet \
+                    --system \
+                    --no-create-home \
+                    --group \
```
yohanboniface: No need to give a name to `--group`?

extras/bin/ideascube
```diff
@@ -17,4 +17,11 @@ export DEBUG=$DEBUG
 export STORAGE_ROOT=$STORAGE_ROOT
 export IDEASCUBE_ID=$IDEASCUBE_ID
 
+
+if [ "$1" == "start" ]; then
```
yohanboniface: Does that mean that one needs to run `ideascube start`?

debian/control
bochecha: Yes.

https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Description

> Those containing a single space followed by a single full stop character. These are rendered as blank lines. This is the only way to get a blank line.

yohanboniface:
Are we in sync on the ansiblecube side? I.e. can we merge/release yet? :)

debian/ideascube.postinst
bochecha: Quoting the `adduser(8)` : 

> **--group**
> When  combined  with --system, a group with the same name and ID as the system user is created.  If not combined with --system, a group  with  the  given  name  is  created.  This is the default action if the program is invoked as addgroup.

We are running `adduser` with the `--system` option, therefore no, we do not need to specify the group name.

Note that I did test the command. :smiley:

extras/bin/ideascube
bochecha: Isn't that how it is started already?

yohanboniface: Nope, the service is started by uwsgi.

bochecha:
> Are we in sync on the ansiblecube side? I.e. can we merge/release yet? :)

No, we're not.

The needed changes are identified and I opened issues for them.

@letompouce started working on some of them, but I'm not sure that's finished?

As I was discussing with @letompouce this week-end, the problem is that (some of) the changes need to be done in sync between ideascube and ansiblecube (but some can be done right now without any problem). That means:

* if we do (some of) those changes in ansiblecube, then the current ideascube package won't be installed correctly
* if we release a new package with these changes merged, then the current ansiblecube playbooks won't do the right thing

So we need to coordinate this a minimum.

extras/bin/ideascube
bochecha: Ah, well, yes, of course, hmm, I mean, hmm, I knew that!

...

...

...

\*frenetically fixes this, hoping nobody realizes anything\*

yohanboniface: Not sure what is the best entry point, then: wsgi.py will not be loaded when running locally with manage.py, and vice versa.
Maybe hooking in wsgi.py is better, guessing that in dev process the migrate command will be run by hand, and the collectstatic is useless.

bochecha: Exactly, collectstatic is useless for development and migrate should really be run exclusively manually by the developer (especially if you're working on migrations themselves)

Is there any other use case than development for running locally with `manage.py`?

If not, then let's not do those automatically at all for this case.

Hooking in uwsgi.py is where I should have hooked in the first place. I know that, since this very pull-request is changing some of the uwsgi setup. But somehow, I still went with `ideascube start`. Don't ask me where that came from, I don't know.

I'll fix this.

bochecha: Fixed in the latest push to this branch.

bochecha:
Hey, look at that! 20 less commits in this branch! :-°

bochecha:
Ideascube moved to Gitlab.

https://framagit.org/ideascube/ideascube/merge_requests/341

PR #342, opened 2016-04-13
Multi db

mgautierfr:
This PR is based on (and replace) PR #330.
This PR is rebased on master.

The catalog install medias "in" the mediacenter without extra files copy.
So the backup now do not save medias installed from the catalog and should be smaller (and so usable)


conftest.py
```diff
@@ -17,7 +26,47 @@ def staffuser():
...
+
+    # Update paths replacing STORAGE_ROOT by tmpdir.
+    for k, v in original_paths.items():
+        rel_path = os.path.relpath(v, STORAGE_ROOT)
+        new_temp_path = os.path.join(str(tmpdir), rel_path)
+        try:
+            os.makedirs(new_temp_path)
+        except FileExistsError:
+            pass
+        setattr(settings, k, new_temp_path)
```
yohanboniface: Seems you should be using the `settings` fixture.

db_router.py
```diff
@@ -0,0 +1,34 @@
+
+
+class TmpBaseRouter(object):
```
yohanboniface: I think we should find a better name here. `tmp` is connoted to something like `data that should be deleted at some point`, while we try to express something like `data that is not backuped`.
Maybe something around `volatil` or just `secondary`.
Let's brainstorm ;)

db_router.py
```diff
@@ -0,0 +1,34 @@
...
+    def allow_relation(self, obj1, obj2, **hints):
+        """
+        Allow relations if a model in the auth app is involved.
+        """
+        if obj1._meta.model_name == 'search' or \
+           obj2._meta.model_name == 'search':
+            return True
+        return None
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
```
yohanboniface: I think we should keep migrations up: when Search models (and later catalog) will change, we need a way to regenerate the db. We decided in #335 to specifically use a migration for that instead of adding a step in the Debian install/update process. I think we should follow that way.

db_router.py
```diff
@@ -0,0 +1,34 @@
+
+
+class TmpBaseRouter(object):
+    """
+    A router to control all database operations on models in the
```
yohanboniface: Docstring needs moar love ;)

db_router.py
```diff
@@ -0,0 +1,34 @@
...
+    auth application.
+    """
+    def db_for_read(self, model, **hints):
+        """
+        Attempts to read auth models go to auth_db.
+        """
+        if model._meta.model_name == 'search':
```
yohanboniface: What about letting Model register them-selves for this router? I'm thinking about adding a `Meta` property like `router = "xxxx"`? Or maybe `read_router/write_router` but that'd be a bit overkill for now.

ideascube/mediacenter/forms.py
```diff
@@ -27,3 +30,30 @@ def save(self, commit=True):
...
+
+
+def CreateDocumentForm(path):
+    class NotFileDocumentForm(forms.ModelForm):
+
+        class Meta:
+            model = Document
+            exclude = ['original', 'preview']
+
+        original = forms.FilePathField(path=path)
```
yohanboniface: path can be a callable IIRC, so no need for a wrapper function I think.

ideascube/mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -19,10 +19,19 @@
 {% endblock twothird %}
 {% block third %}
     {% if user.is_staff %}
-        <ul class="card tinted admin">
-            <li>{% fa 'pencil' 'fa-fw' %}<a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
-            <li>{% fa 'trash' 'fa-fw' %}<a href="{% url 'mediacenter:document_delete' pk=document.pk %}">{% trans 'Delete' %}</a></li>
-        </ul>
+        {% if document.package_id %}
+            <span class="card admin">
+               {% trans 'Cannot edit or delete this document.' %}
```
yohanboniface: Use `blocktrans` for multiline or complex translate string :)

ideascube/search/tests/test_models.py
```diff
@@ -3,23 +3,28 @@
 
 from ideascube.blog.tests.factories import ContentFactory
 from ideascube.blog.models import Content
+from ideascube.mediacenter.models import Document
```
yohanboniface: Unused imports?

ideascube/blog/tests/test_search.py
```diff
@@ -12,16 +12,19 @@ def test_nothing_is_indexed_without_any_fixture():
     assert Content.objects.count() == 0
 
 
+@pytest.mark.usefixtures('cleansearch')
```
yohanboniface: Maybe move that in a `pytest_setup` conftest.py hook not to repeat it again and again?
Not sure it's good idea because: it's less explicit, it will consume a bit more time (because will be run at every test).
But let's think about it.

ideascube/search/utils.py
```diff
@@ -1,17 +1,24 @@
...
-from django.db import connection
+from django.db import connections
 
 
 def create_index_table(force=True):
-    cursor = connection.cursor()
-    cursor.execute("SELECT count(*) FROM sqlite_master "
-                   "WHERE type='table' AND name='idx';")
-    count = cursor.fetchone()[0]
+    cursor_default = connections['default'].cursor()
```
yohanboniface: Shouldn't this only be in a migration?

ideascube/serveradmin/backup.py
```diff
@@ -75,12 +75,37 @@ def format(self):
...
-            base_name=os.path.join(Backup.ROOT, self.basename),  # W/o ext.
-            format=self.format,
-            root_dir=settings.BACKUPED_ROOT,
-            base_dir='./'
-        )
+        base_name = os.path.join(Backup.ROOT, self.basename)
+        base_name = "{}{}".format(base_name,
+                                  self.FORMAT_TO_EXTENSION[self.format])
+        try:
+            if self.format == 'zip':
```
yohanboniface: I guess we cannot exclude links from while using `shutil.make_archive`.
And thus I wonder if it's not an occasion to drop `zip` support for backup? Trying to keep things simple.

ideascube/serveradmin/catalog.py
```diff
@@ -296,6 +303,91 @@ def remove(self, install_dir):
...
+                printerr("Cannot install media {} from package {}".format(
+                    media['title'], self.id))
+                continue
+
+    def _install_media(self, media_info, pseudo_install_dir):
+        try:
+            media_info['title'] = title = smart_truncate(media_info['title'])
+        except KeyError:
+            raise InvalidPackageContent('Missing {} in {}'.format(
+                entry, media_info))
```
yohanboniface: `entry` does not seem to be defined ;)

ideascube/serveradmin/catalog.py
```diff
@@ -296,6 +303,91 @@ def remove(self, install_dir):
...
+        if 'lang' not in media_info:
+            media_info['lang'] = settings.LANGUAGE_CODE
+
+        kind = media_info.get('kind')
+        if not kind or not hasattr(Document, kind.upper()):
+            content_type, _ = mimetypes.guess_type(original)
+            media_info['kind'] = guess_kind_from_content_type(content_type) \
+                or Document.OTHER
+
+        instance = Document.objects.filter(title=title, kind=kind).last()
```
yohanboniface: I think this is not needed here. It was in the import_medias command because it was the only way to rerun the command without reimporting everything, but here we should not be reimporting without first having deleted all the medias with the package_id.

ideascube/serveradmin/tests/conftest.py
```diff
@@ -1,8 +1,23 @@
 import pytest
+import os
 
 from ..backup import Backup
 
 DATA_ROOT = 'ideascube/serveradmin/tests/data/backup'
+BACKUPS_ROOT = 'ideascube/serveradmin/tests/backups'
```
yohanboniface: I'm a bit confused about those two paths ending with `backup`.

conftest.py
mgautierfr: Yes, and monkeypatch :smile:

db_router.py
mgautierfr: `return None` means "we do not care" not "no migration" (False)

db_router.py
mgautierfr: I've tried this, but we cannot add custom attribute to the Meta properties :cry:

ideascube/mediacenter/forms.py
mgautierfr: I don't get it here, the path is use a the creation of the class and it come from outside.
We could use a callable but we still need to provide this custom callable at class definition.

ideascube/blog/tests/test_search.py
mgautierfr: I prefer to have it explicit.
On top of the performance point, it let remind future developer that this test especially need
a clean search database

ideascube/serveradmin/tests/conftest.py
mgautierfr: Agree with you.
BACKUPS_ROOT come from test_backup.py.
I copy it just to be able to create the backup_root fixture.

However, this should not be necessary with the storage_dir fixture who include also the creation of a new BACKUP_ROOT.

There is lot to do on this point; a lots of tests in test_view.py and test_backup.py still monkeypath themselves the Backup.ROOT stuff. We should cleanup all those tests with better fixture.
But this PR is big enough I think. We should clean this is a second step.

db_router.py
yohanboniface: Can be a first level class property at this point

db_router.py
yohanboniface: Why not True, then? I'm not sure what the meaning of "we do not care" is, in this case ;)

db_router.py
mgautierfr: That was too easy for me to think about it :)
I'll do it.

db_router.py
mgautierfr: "we do not care" is "use the default behavior".
It can depend on a lot of things (other Database Routers) but, it is probably True in our case.
However we can indeed set it to True be sure that migration will be done.

barbayellow:
you make my day !

bochecha:
Can you rebase on master? The branch can't be merged as it is, due to conflicts.

mgautierfr:
Arf, yes, I'll do it.

conftest.py
```diff
@@ -1,9 +1,17 @@
...
 from django.core.urlresolvers import reverse
+from django.core.files.storage import default_storage
 
 from ideascube.tests.factories import UserFactory
+from ideascube.search.utils import create_index_table
+
+
+@pytest.fixture()
+def cleansearch():
+    create_index_table(True)
```
bochecha: `True`?

Please use the kwarg notation here. Boolean args have this horrible characteristics that they are extremely unexplicit about what they do.

conftest.py
```diff
@@ -17,7 +25,45 @@ def staffuser():
...
+def storage_dir(monkeypatch, tmpdir, settings):
+    """Return a new temporary storage dir.
+    Settings are also modified to point to the directory."""
+    STORAGE_ROOT = settings.STORAGE_ROOT
+
+    # Get the paths to changes and their original values.
+    original_paths = (k for k in dir(settings.wrapped_object)
+                      if k.isupper() and not k.startswith('__'))
+    original_paths = (k for k in original_paths
+                      if k.endswith('_ROOT') or k.endswith('_DIR'))
```
bochecha: This seems pretty fragile, as its based on an implicit, unwritten naming convention for settings.

Why isn't it enough to replace `STORAGE_ROOT`? I thought all other paths were relative to it?

conftest.py
```diff
@@ -17,7 +25,45 @@ def staffuser():
...
+    original_paths = {k: getattr(settings, k) for k in original_paths}
+    original_paths = {k: v for k, v in original_paths.items() if v}
+
+    # Update paths replacing STORAGE_ROOT by tmpdir.
+    for k, v in original_paths.items():
+        rel_path = os.path.relpath(v, STORAGE_ROOT)
+        new_temp_path = os.path.join(str(tmpdir), rel_path)
+        try:
+            os.makedirs(new_temp_path)
```
bochecha: Maybe add an `exist_ok=True`, just to be safe? (in case two settings use the same value)

db_router.py
```diff
@@ -0,0 +1,32 @@
+
+
+class DatabaseRouter(object):
+    """
+    A Database router.
+    Some models do not need (and must not be backuped).
+    Those models will (must) have a _dbname class attribute set to 'transiant'.
+    This attribute will be use as database name.
```
bochecha: I don't understand the last line.

db_router.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """
+        Allow relations if both object are in the same database.
+        """
+        return getattr(obj1, '_dbname', None) == getattr(obj2, '_dbname', None)
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        """
+        All database need to be migrated.
```
bochecha: That's not what returning `None` achieves, though.

> Return True if the operation should run, False if it shouldn’t run, or None if the router has no opinion.

ideascube/blog/tests/test_search.py
bochecha: +1 on being explicit here.

ideascube/conf/base.py
```diff
@@ -315,6 +315,6 @@
 
 
 SESSION_COOKIE_AGE = 60 * 60  # Members must be logged out after one hour
-BACKUP_FORMAT = 'zip'  # One of 'zip', 'tar', 'bztar', 'gztar'
+BACKUP_FORMAT = 'gztar'  # One of 'tar', 'bztar', 'gztar'
```
bochecha: Since we have Python 3.4, we should also be able to support xz compression.

I know, this is entirely unrelated to your pull-request, just figured I'd mention it. :stuck_out_tongue_winking_eye:

ideascube/mediacenter/forms.py
bochecha: So it seems the Django recommanded way of doing this is:

```python
class NotFileDocumentForm(forms.ModelForm):
    class Meta:
        model = Document
        exclude = ['original', 'preview']

    def __init__(self, path, *args, **kwargs):
        super(TicketForm, self).__init__(*args, **kwargs)
        self.fields['original'] = forms.FilePathField(path=path)
        self.fields['preview'] = forms.FilePathField(path=path, required=False)

    def save(self, commit=True):
        ...
```

At least that's what StackOverflow tells me. :stuck_out_tongue: 

It does feel a bit better/simpler than a wrapping function, but it's not as nice since the fields are not defined in their usual place (as class attributes).

ideascube/mediacenter/forms.py
```diff
@@ -27,3 +30,30 @@ def save(self, commit=True):
...
+        preview = forms.FilePathField(path=path, required=False)
+
+        def save(self, commit=True):
```
bochecha: You don't seem to use this commit attribute?

ideascube/mediacenter/forms.py
```diff
@@ -27,3 +30,30 @@ def save(self, commit=True):
...
+            document = super(NotFileDocumentForm, self).save(commit=False)
+            original = self.cleaned_data['original']
+            original = os.path.relpath(original, settings.MEDIA_ROOT)
```
bochecha: Why do you prefer storing the path relative to the media root?

I assume later, when opening the file, we have to `os.path.join(settings.MEDIA_ROOT, document.original)`, which defeats the purpose, doesn't it?

mgautierfr: Yes, we do a os.path.join later. But it is django storagesystem who do it, so we have no choice, we need to store a relative path.

ideascube/mediacenter/forms.py
mgautierfr: Hum... indeed.

But I not sure of what we must not do if commit==False.
Do not do the save calls ?

ideascube/mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -19,10 +19,19 @@
...
-        <ul class="card tinted admin">
-            <li>{% fa 'pencil' 'fa-fw' %}<a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
-            <li>{% fa 'trash' 'fa-fw' %}<a href="{% url 'mediacenter:document_delete' pk=document.pk %}">{% trans 'Delete' %}</a></li>
-        </ul>
+        {% if document.package_id %}
+            <span class="card admin">
+               {% blocktrans %}
+                   Cannot edit or delete this document.
+                   It is installed by the {{ document.package_id }} package.
+               {% endblocktrans %}
```
bochecha: So I have to wonder... We're slowly going to a future where all content in Ideascube is managed by packages, including content created and uploaded by users.

Which pretty much means that the above block will be always visible, and no content will be modifiable.

That doesn't feel right, wouldn't it be better to let users modify the content, improve it, adapt it and share it freely?

Or is this just temporary, and eventually the content will be modifiable again, with modifications simply becoming package updates?

ideascube/mediacenter/tests/test_views.py
```diff
@@ -104,6 +104,21 @@ def test_staff_can_edit_document(staffapp, video):
     assert Document.objects.get(pk=video.pk).title == title
 
 
+def test_staff_detail_page_has_no_edit_delete_link_package_id(staffapp):
+    document = DocumentFactory(package_id="foo")
+    response = staffapp.get(reverse('mediacenter:document_detail',
+                                    kwargs={'pk': document.pk}), status=200)
+    assert "Cannot edit or delete this document." in response.content.decode()
+    assert '<a href="/en/mediacenter/document/{}/edit/">Edit</a>'.format(document.pk) not in response.content.decode()
```
bochecha: Maybe also check the delete link isn't there?

ideascube/mediacenter/tests/test_views.py
```diff
@@ -104,6 +104,21 @@ def test_staff_can_edit_document(staffapp, video):
...
+
+
+def test_staff_detail_page_has_edit_delete_link_no_package_id(staffapp):
+    document = DocumentFactory(package_id="")
+    response = staffapp.get(reverse('mediacenter:document_detail',
+                                    kwargs={'pk': document.pk}), status=200)
+    assert "Cannot edit or delete this document." not in response.content.decode()
+    assert '<a href="/en/mediacenter/document/{}/edit/">Edit</a>'.format(document.pk) in response.content.decode()
```
bochecha: Same as above.

ideascube/search/tests/test_models.py
```diff
@@ -3,23 +3,27 @@
 
 from ideascube.blog.tests.factories import ContentFactory
 from ideascube.blog.models import Content
+from ideascube.mediacenter.tests.factories import DocumentFactory
```
bochecha: I think you're not using this import anywhere in the file?

ideascube/search/utils.py
```diff
@@ -1,18 +1,29 @@
...
+    cursor_transient.execute("SELECT count(*) FROM sqlite_master "
+                             "WHERE type='table' AND name='idx';")
+    count = cursor_transient.fetchone()[0]
     if not count or force:
-        cursor.execute("DROP TABLE IF EXISTS idx")
-        cursor.execute("CREATE VIRTUAL TABLE idx using "
-                       "FTS4(id, model, model_id, public, text)")
+        cursor_transient.execute("DROP TABLE IF EXISTS idx")
+        cursor_transient.execute("CREATE VIRTUAL TABLE idx using "
+                                 "FTS4(id, model, model_id, public, text)")
```
bochecha: Is there a reason we can't use Django here, to create the table, since it maps to the `Search` model?

ideascube/serveradmin/backup.py
```diff
@@ -113,6 +128,11 @@ def delete(self):
 
     @classmethod
     def create(cls, format=None):
+        format = format or Backup.FORMAT
+        if format not in Backup.SUPPORTED_FORMATS_AT_CREATION:
+            raise ValueError(
+                _("Format {} is not supported at creation").format(format)
+            )
         name = make_name(format)
         backup = Backup(name)
```
bochecha: So we create a `Backup` passing it just a `name`, even though we already know its format.

And then, we have a `guess_file_format` method which has a mapping of extensions to format, and from the the `name` deducts the format.

Why not just store the format as an attribute of the `Backup` instance?

I know, that's not something you touched here. :smile:

ideascube/serveradmin/catalog.py
```diff
@@ -296,6 +303,86 @@ def remove(self, install_dir):
...
+        media_info['original'] = os.path.join(pseudo_install_dir,
+                                              media_info['path'])
+
+        if 'preview' in media_info:
+            media_info['preview'] = os.path.join(pseudo_install_dir,
+                                                 media_info['preview'])
+
+        self._save_media(media_info, pseudo_install_dir)
+
+    def _save_media(self, metadata, install_dir):
```
bochecha: I wonder if all this interaction with the database shouldn't be the responsibility of the handler, rather than the package?

ideascube/serveradmin/catalog.py
```diff
@@ -296,6 +303,86 @@ def remove(self, install_dir):
...
+        form = CreateDocumentForm(install_dir)(data=metadata, instance=None)
+
+        if form.is_valid():
+            doc = form.save()
+            print("Media", doc, "has been added to database.")
```
bochecha: I'm not sure it's interesting to know the document has been inserted into the database.

We already report that the package has been installed, and inserting it into the database is merely part of that installation process.

ideascube/serveradmin/tests/test_backup.py
```diff
@@ -19,7 +18,7 @@ def test_backup_init_should_raise_with_malformatted_string():
 
 
 def test_underscore_as_separator_should_still_be_supported():
-    Backup('edoardo-0.0.0-201501231405.zip')
+    Backup('edoardo-0.0.0-201501231405.tar.gz')
```
bochecha: Huh?

The method talks about underscores as separator, but I can't see any underscore here?

ideascube/serveradmin/tests/conftest.py
```diff
@@ -1,12 +1,27 @@
...
+
+
+@pytest.fixture
+def backup_root(request):
+    try:
+        os.makedirs(BACKUPS_ROOT)
+    except OSError:
+        pass  # Already exists.
```
bochecha: We have Python >= 3.4:

```python
os.makedirs(BACKUPS_ROOT, exist_ok=True)
```

ideascube/serveradmin/tests/test_backup.py
```diff
@@ -32,61 +31,29 @@ def test_guess_format_from_name(input, expected):
...
     open(bad, mode='w')
     backups = list(Backup.list())
     assert len(backups) == 1
     assert backups[0].name == filename
     os.remove(good)
     os.remove(bad)
-    os.rmdir(BACKUPS_ROOT)
-
-
-def test_create_zipfile(monkeypatch, settings):
```
bochecha: Should you not replace that by a test that ensures we can't create zip files any more?

ideascube/serveradmin/tests/test_backup.py
```diff
@@ -107,15 +74,26 @@ def test_create_tarfile(monkeypatch, settings, format, extension):
...
     archive.close()
     os.remove(filepath)
     os.remove(proof_file)
+    os.remove(symlink_file)
 
+def test_create_zipfile_must_fail(monkeypatch, tmpdir):
+    monkeypatch.setattr('ideascube.serveradmin.backup.Backup.ROOT', str(tmpdir))
+    monkeypatch.setattr('ideascube.serveradmin.backup.Backup.FORMAT', 'zip')
+    assert len(list(Backup.list())) == 0
+    with pytest.raises(ValueError) as excinfo:
```
bochecha: Do we want a more specific error type? Something like `UnsupportedBackupFormat` ?

ideascube/serveradmin/tests/test_backup.py
```diff
@@ -154,10 +133,17 @@ def test_load(monkeypatch, extension):
     os.remove(backup_path)
 
 
-def test_delete(monkeypatch):
+@pytest.mark.parametrize('extension', [
+    ('.zip'),
+    ('.tar'),
+    ('.tar.gz'),
+    ('.tar.bz2'),
+])
```
bochecha: Just a big +1 for having `parametrize` here to test multiple formats. :smiley:

ideascube/mediacenter/templates/mediacenter/document_detail.html
mgautierfr: We are going to a situation where we have two separate application : 
- One displaying the content (so, without modification)
- One allowing user to create/modify content (by creating packages)

It could be interesting on the architecture point of view, to think on how the two applications will interact and how the user will use them. (but it's out of scope for now :) )

ideascube/mediacenter/tests/test_views.py
mgautierfr: This is a (wrong) developer assumption. I know that both links (de)activated by the same if, so I've tested only one link.

But yes, we should test also the delete link.

ideascube/search/utils.py
mgautierfr: Hum, I don't know.
@yohanboniface will probably know why.

ideascube/serveradmin/backup.py
mgautierfr: Yes, we should do some refactoring.

In another pr maybe ? To avoid a too big pr here.

ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -319,6 +326,32 @@ def test_remove_staticsite(staticsite_path, install_dir):
     assert root.check(exists=False)
 
 
+def test_install_zippedmedia(zippedmedia_path, install_dir, storage_dir):
```
bochecha: You don't use `storage_dir` here, but I assume it's required to setup the settings etc.

Isn't there a way to say this test function requires this fixture, without passing it as a parameter?

As it is, I'm worried that in a few months, someone sees the parameter is unused, and decides to remove it. :worried:

ideascube/search/utils.py
yohanboniface: `FST4` is a sqlite extension that Django does not cover. It adds indices dedicated to full text search, but AFAIK Django does not know how to map to it.

ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -546,6 +579,64 @@ def test_nginx_commits_after_remove(tmpdir, settings, staticsite_path,
     assert manager().get_service.call_count == 2
     manager().restart.assert_not_called()
 
+pytestmark = pytest.mark.django_db
```
bochecha: Doesn't that mark all the tests in this module?

Since most don't require the db to be set up, wouldn't it be preferrable to only explicitly mark the functions requiring it?

ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -546,6 +579,64 @@ def test_nginx_commits_after_remove(tmpdir, settings, staticsite_path,
...
+    assert Document.objects.count() == 3
+    video = Document.objects.get(title='my video')
+    assert video.summary == 'my video summary'
+    assert video.kind == Document.VIDEO
+    assert Document.objects.search('summary').count() == 3
+
+    packaged_documents = list(Document.objects.filter(package_id='test-media'))
+    assert len(packaged_documents) == 3
+
+    # Be sure that referenced documents are the ones install by the package
```
bochecha: ... are the ones install**ed** by...

ideascube/serveradmin/tests/test_views.py
```diff
@@ -231,20 +231,20 @@ def test_backup_should_list_available_backups(staffapp, backup):
...
     radio_options = form.get('backup').options
     assert len(radio_options) == 4
-    assert radio_options[3][0] == backup.name
+    assert radio_options[2][0] == backup.name
 
 
 def test_backup_button_should_save_a_new_backup(staffapp, monkeypatch,
                                                 settings):
-    monkeypatch.setattr('ideascube.serveradmin.backup.Backup.FORMAT', 'zip')
+    monkeypatch.setattr('ideascube.serveradmin.backup.Backup.FORMAT', 'gztar')
```
bochecha: Having to monkeypatch this makes me think it probably shouldn't be a class attribute in the first place...

pytest.ini
```diff
@@ -3,3 +3,4 @@ addopts = -x
 DJANGO_SETTINGS_MODULE=ideascube.settings
 norecursedirs = debian
                 .pybuild
+usefixtures = storage_dir
```
bochecha: Does this mean it gets used for every single unit tests?

If so, why did you pass it to some test functions as a parameter?

bochecha:
One more thing: why did you drop support for creating zip backups?

Please explain it somewhere. (the commit message is a perfectly fine place to do so)

pytest.ini
mgautierfr: Cause I add the usefixtures after and haven't removed the fixture as parameter.

ideascube/mediacenter/forms.py
bochecha: I don't think we want that.

I'd just use `**_` here, to catch all undesired kwargs and show we don't do anything with them.

ideascube/mediacenter/forms.py
bochecha: Well that's too bad. :disappointed:

db_router.py
mgautierfr: Fix in last push --force

ideascube/serveradmin/tests/test_backup.py
mgautierfr: Hum, well.. I don't know.

db_router.py
bochecha: Oh, so now you're returning True? I have to admit, I thought returning None was better.

ideascube/serveradmin/catalog.py
bochecha: Anybody wants to chime in on this?

ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -547,6 +586,61 @@ def test_nginx_commits_after_remove(tmpdir, settings, staticsite_path,
...
+    assert Document.objects.count() == 3
+    video = Document.objects.get(title='my video')
+    assert video.summary == 'my video summary'
+    assert video.kind == Document.VIDEO
+    assert Document.objects.search('summary').count() == 3
+
+    packaged_documents = list(Document.objects.filter(package_id='test-media'))
+    assert len(packaged_documents) == 3
+
+    # Be sure that referenced documents are the ones installed by the package
```
bochecha: I think you mean "Make sure that" or "Ensure that" here, not "Be sure that".

bochecha:
> One more thing: why did you drop support for creating zip backups?

You don't seem to have addressed this.

bochecha:
This is getting big, we've discussed about a lot of things, you've addressed most of them I think, but there are still a few things left here and there, lost in the sea of changes. :smile: 

So to make it easier, here's the list of things I'd like to see addressed before I'd be comfortable merging this. (of course, someone else might have a different opinion on what still needs to be addressed)

* [ ] Why was the zip support dropped for creating backups?

    This really needs to be explained somewhere.

    Whether it's an issue (marked fixed by the commit), the commit message itself, or another way, doesn't matter much to me, but I find it important to write it down.

* [ ] I still think the way you identify the path settings to mangle by finding those ending with `_ROOT` or `_DIR` is fragile, as it relies on an unwritten convention.

    Wouldn't it be better to loop over all settings, and modify all those that are 1. strings and 2. are under `STORAGE_ROOT`?

* [ ] Should our router really decide to force migrations?

    Django has complex rules to manage the database models, and especially how to run migrations.

    Since our router is only in the business of choosing which database to use for which model, I don't think it should have any impact on whether migrations are run, unless that impact is specifically about which database is used (e.g we don't want to ever run migrations in one of the database, but run it in the other, ...)

    So I think our router's `allow_migration` method should return `None`.

    @yohanboniface you seemed to have reservations on this, can I convince you that letting Django decide is the right thing to do, here? :smile: 

* [ ] The more I think about it, the more I think that the db interaction as part of installing media should be done by the handler, not by the package itself

    Eventually, we might get to a place where we unpack the zips all at once, in parallel, and we'd want to bulk-insert the media in the database.

    I'm not saying you should implement that in this PR, but moving this code to the handler would be the first step towards that.

Now, the above are points that I think should be addressed before merging. And of course, one valid way to address them is to convince me that I'm wrong on those. :wink:

mgautierfr:
I agree that all the points you're mentioned should be addressed.
However, this PR is pretty big and I prefer to not continue make it grow even more.

- For point 2 and 4, the code is not perfect but it's working. This is code improvement so I suggest to not address theses points for the merge and move them in different issues (to be addressed later). 
- Point 1 should be fixed, it a missing explanation on a code allready in the PR.
- Point 3 should also be fixed as it is just a line of code. Important point is discussion about force migration or let django decide.

ideascube/serveradmin/tests/test_catalog.py
mgautierfr: In this test I do nothing but a check that documents are installed correctly.
I do not take any action to "make sure" or "ensure" that every thing is correct, I just want to "be sure" that this is correct.

However, maybe my English is not good enough and "make sure" is the right expression.
And it is in the comment so, I don't especially care :)

bochecha:
About the `STORAGE_ROOT` fixture (my point 2 above), there is something else that could cause problems.

You're not ensuring that the path you mangle is under `STORAGE_ROOT` in the first place.

For example, we have `CATALOG_CACHE_BASE_DIR` which is set to `/var/cache/ideascube/catalog`

```
>>> import os
>>> 
>>> STORAGE_ROOT = '/var/ideascube'
>>> v = '/var/cache/ideascube/catalog'
>>> 
>>> os.path.relpath(v, STORAGE_ROOT)
'../cache/ideascube/catalog'
```

Do we really want to set `CATALOG_CACHE_BASE_DIR` to `os.path.join(str(tmpdir), '../cache/ideascube/catalog')` ? Won't that make it shared by all tests?

Honestly, I'm starting to think this fixture should just set all the paths explicitly, one by one. :unamused:

mgautierfr:
Yes, you're right, there is a bug here, CATALOG_CACHE_BASE_DIR should be a subpath of tmpdir.
There is more to do if a dir is not originally a subpath of STORAGE_ROOT.

However, this is not worth than now. For now, there is no fixture at all to avoid CATALOG_CACHE_BASE_DIR to be shared by test and, when needed, CATALOG_CACHE_BASE_DIR is changed in the test itself (test_catalog_reinstall_package).

As I said, I agree that this fixture is far from being perfect and must be fixed. But the solution is not so easy. Do we change all path subdir of STORAGE_ROOT ? all variable ending with (_DIR|_ROOT) ? set all path one by one ?
There is no regression*, and we need to reduce the scope here if we want this PR to be merge in the next release.

* There is a regression here, that create a directory by side of tmpdir. But travis create a virtual machine for all tests (with which config ?) and on dev config it is a subdir of STORAGE_ROOT.

bochecha:
> For now, there is no fixture at all to avoid CATALOG_CACHE_BASE_DIR to be shared by test

There also wasn't any fixture for storage root, but you added one in this PR. :wink: 

> we need to reduce the scope here if we want this PR to be merge in the next release.

In general I don't think merging something is so important that we shouldn't try and fix obvious issues. Whether these specific issues are bad enough to block the merge or not is something we can discuss, of course.

I do agree that this PR has moved beyond its original scope of handling multiple database: it also heavily touches on the backup code in multiple ways, and adds a new content package type. Should we consider splitting it?

---

On a more general note, I think we are falling victims of something I'm going to call the "Github workflow trap". That is, the idea that since it's so easy to merge all the commits in a branch at once, then we should do so.

If a PR gets too big, we could certainly decide to merge some of its commits that we've already reviewed. We don't need to split it into multiple PR for that, we just need to `git merge` or `git cherry-pick` in our consoles.

In this specific case, we could easily merge a few commits into master, for example:

* 5311ce12ea1e61a9b4870e1b3f4a99aa41f21ddd
* c04245a26279911eadc0aff5bb5e6664e0514fe5
* 14255ab9103363a5ef1e133d8c1388cdd80a4f85
* 5e848fbf71b4b42e015c09da99983c7972f66e77
* 4eb676be1f734b0eb80a57b42a7b5519e8bf92aa

In fact, these should really be squashed as one commit, since they all just fix what the first one did. And if we merge them into master, then the PR suddenly becomes 19% smaller. :smiley:

bochecha:
Also, if you rebase this PR on master, then the tests are actually broken:

```
E           PermissionError: [Errno 13] Permission denied: '/tmp/pytest-of-mathieu/pytest-5/test_it_should_create_a_defaul0/../../../../../../var/cache/ideascube'
```

bochecha:
So I've talked with @mgautierfr today, and we both agreed that this PR has become too big and unwieldy, conflating various changes that came up as he was working on this. (he started working on installing media from packages, which evolved into backup and multi-db territory)

Since he's on holiday this week, I'm going to take this PR and split it up into smaller ones.

I'll also fix the things we agreed upon in this review, but which he hasn't had the time to fix yet. (see previous comments). Those are:

* explaining why we dropped zip for new backups
* fixing the issue with the cache folder due to the `storage_root` fixture

I'll leave out moving the code from the package to the handler, though.

bochecha:
I rebased @mgautierfr's branch onto master, reordered, squashed and split commits, and made a few tiny cosmetic changes.

Here is the complete diff between his previous branch and the one I just force-pushed:

```diff
diff --git a/db_router.py b/db_router.py
index a5a9ce0..b3b0ca1 100644
--- a/db_router.py
+++ b/db_router.py
@@ -1,5 +1,3 @@
-
-
 class DatabaseRouter(object):
     """
     A Database router.
diff --git a/debian/changelog b/debian/changelog
index 9cb5022..d27929a 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,21 @@
+ideascube (0.10.3) unstable; urgency=low
+
+  * Fix Dusseldorf conf
+  * Do not fallback on dev settings anymore (#380)
+
+ -- Yohan Boniface <hi@yohanboniface.me>  Tue, 11 Jul 2016 22:12:00 +0100
+
+
+ideascube (0.10.2) unstable; urgency=low
+
+  * Add conf for Sarcelles and Dusseldorf
+  * Added an order_by selector in list pages (#363 #368)
+  * Fixed page not being reset when changing filters in list pages (#364)
+  * Added Spanish in AVAILABLE_LANGUAGES
+
+ -- Yohan Boniface <hi@yohanboniface.me>  Tue, 22 Jun 2016 12:12:00 +0100
+
+
 ideascube (0.10.1) unstable; urgency=low
 
   * Use our own fork of resumable-urlretrieve
diff --git a/ideascube/__init__.py b/ideascube/__init__.py
index 7db4f68..f8cb779 100644
--- a/ideascube/__init__.py
+++ b/ideascube/__init__.py
@@ -1,3 +1,3 @@
-VERSION = (0, 10, 1)
+VERSION = (0, 10, 3)
 
 __version__ = ".".join(map(str, VERSION))
diff --git a/ideascube/conf/idb_de_dusseldorf.py b/ideascube/conf/idb_de_dusseldorf.py
deleted file mode 100644
index 2e942b5..0000000
--- a/ideascube/conf/idb_de_dusseldorf.py
+++ /dev/null
@@ -1,40 +0,0 @@
-# -*- coding: utf-8 -*-
-"""Ideaxbox for Welcome Point, Dusseldorf"""
-from .idb import *  # noqa
-from django.utils.translation import ugettext_lazy as _
-
-IDEASCUBE_NAME = u"Welcome Point Dusseldorf"
-IDEASCUBE_PLACE_NAME = _("city")
-COUNTRIES_FIRST = ['DE']
-TIME_ZONE = None
-LANGUAGE_CODE = 'de'
-LOAN_DURATION = 14
-MONITORING_ENTRY_EXPORT_FIELDS = ['serial', 'user_id', 'birth_year', 'gender']
-USER_FORM_FIELDS = (
-    ('Ideasbox', ['serial', 'box_awareness']),
-    (_('Personal informations'), ['short_name', 'full_name', 'birth_year', 'gender', 'id_card_number']),  # noqa
-    (_('Family'), ['marital_status', 'family_status', 'children_under_12', 'children_under_18', 'children_above_18']),  # noqa
-    (_('In the town'), ['current_occupation', 'school_level']),
-    (_('Language skills'), ['en_level']),
-)
-HOME_CARDS = STAFF_HOME_CARDS + [
-    {
-        'id': 'blog',
-    },
-    {
-        'id': 'library',
-    },
-    {
-        'id': 'mediacenter',
-    },
-    {
-        'id': 'khanacademy',
-    },
-    {
-        'id': 'wikipedia',
-        'languages': ['de', 'ar', 'fa', 'en']
-    },
-    {
-        'id': 'w2eu',
-    }
-]
diff --git a/ideascube/conf/idb_deu_dusseldorf.py b/ideascube/conf/idb_deu_dusseldorf.py
new file mode 100644
index 0000000..e35d03c
--- /dev/null
+++ b/ideascube/conf/idb_deu_dusseldorf.py
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""Ideaxbox for Welcome Point, Dusseldorf"""
+from .idb import *  # noqa
+from django.utils.translation import ugettext_lazy as _
+
+IDEASCUBE_NAME = u"Welcome Point Dusseldorf"
+IDEASCUBE_PLACE_NAME = _("city")
+COUNTRIES_FIRST = ['DE']
+TIME_ZONE = None
+LANGUAGE_CODE = 'de'
+LOAN_DURATION = 14
+MONITORING_ENTRY_EXPORT_FIELDS = ['serial', 'user_id', 'birth_year', 'gender']
+USER_FORM_FIELDS = (
+    ('Ideasbox', ['serial', 'box_awareness']),
+    (_('Personal informations'), ['short_name', 'full_name', 'birth_year', 'gender', 'id_card_number']),  # noqa
+    (_('Family'), ['marital_status', 'family_status', 'children_under_12', 'children_under_18', 'children_above_18']),  # noqa
+    (_('In the town'), ['current_occupation', 'school_level']),
+    (_('Language skills'), ['en_level']),
+)
+HOME_CARDS = STAFF_HOME_CARDS + [
+    {
+        'id': 'blog',
+    },
+    {
+        'id': 'library',
+    },
+    {
+        'id': 'mediacenter',
+    },
+    {
+        'id': 'khanacademy',
+    },
+    {
+        'id': 'gutenberg',
+        'lang': 'de'
+    },
+    {
+        'id': 'wikipedia',
+        'languages': ['de', 'ar', 'fa', 'en']
+    },
+    {
+        'id': 'w2eu',
+    }
+]
diff --git a/ideascube/conf/idb_fra_sarcelles.py b/ideascube/conf/idb_fra_sarcelles.py
index 292c8fd..4b4b78e 100644
--- a/ideascube/conf/idb_fra_sarcelles.py
+++ b/ideascube/conf/idb_fra_sarcelles.py
@@ -36,10 +36,10 @@ HOME_CARDS = STAFF_HOME_CARDS + [
     },
     {
         'id': 'gutenberg',
-        'languages': ['fr']
+        'lang': 'fr'
     },
     {
-        'id' : 'dirtybiology',
+        'id': 'dirtybiology',
         'languages': ['fr']
     },
     {
diff --git a/ideascube/search/migrations/0002_reindex.py b/ideascube/search/migrations/0002_reindex.py
index c27142e..83dd735 100644
--- a/ideascube/search/migrations/0002_reindex.py
+++ b/ideascube/search/migrations/0002_reindex.py
@@ -4,11 +4,13 @@ from __future__ import unicode_literals
 from django.db import migrations, connections
 from ideascube.search.utils import reindex_content
 
+
 def update_reindex(*args):
     cursor_default = connections['default'].cursor()
     cursor_default.execute("DROP TABLE IF EXISTS idx")
     reindex_content()
 
+
 class Migration(migrations.Migration):
 
     dependencies = [('search', '0001_initial')]
diff --git a/ideascube/search/utils.py b/ideascube/search/utils.py
index c2cb52e..b06957b 100644
--- a/ideascube/search/utils.py
+++ b/ideascube/search/utils.py
@@ -14,9 +14,10 @@ def create_index_table(force=True):
                                  "FTS4(id, model, model_id, public, text, "
                                  "lang, kind, tags)")
 
+
 def reindex_content(*args, **kwargs):
     from ideascube.search.models import SEARCHABLE
-    create_index_table(True)
+    create_index_table(force=True)
     indexed = {}
     for model in SEARCHABLE.values():
         count = 0
@@ -26,6 +27,7 @@ def reindex_content(*args, **kwargs):
         indexed[model.__name__] = count
     return indexed
 
+
 def rank(match_info):
     # Handle match_info called w/default args 'pcx' - based on the example
     # rank function http://sqlite.org/fts3.html#appendix_a
diff --git a/ideascube/serveradmin/backup.py b/ideascube/serveradmin/backup.py
index bf83726..71faad5 100644
--- a/ideascube/serveradmin/backup.py
+++ b/ideascube/serveradmin/backup.py
@@ -77,6 +77,7 @@ class Backup(object):
         """Make a backup of the server data."""
         if self.format == 'zip':
             raise ValueError(_("Zip is no more supported to create backup"))
+
         base_name = os.path.join(Backup.ROOT, self.basename)
         base_name = "{}{}".format(base_name,
                                   self.FORMAT_TO_EXTENSION[self.format])
diff --git a/ideascube/serveradmin/tests/test_backup.py b/ideascube/serveradmin/tests/test_backup.py
index b3c6b71..1e1a986 100644
--- a/ideascube/serveradmin/tests/test_backup.py
+++ b/ideascube/serveradmin/tests/test_backup.py
@@ -56,10 +56,7 @@ def test_list(monkeypatch, settings):
 @pytest.mark.usefixtures('backup_root')
 def test_create_tarfile(monkeypatch, settings, format, extension):
     settings.BACKUPED_ROOT = BACKUPED_ROOT
-    try:
-        os.makedirs(BACKUPED_ROOT)
-    except OSError:
-        pass
+    os.makedirs(BACKUPED_ROOT, exist_ok=True)
     filename = 'edoardo-0.0.0-201501231405' + extension
     filepath = os.path.join(BACKUPS_ROOT, filename)
     try:
@@ -87,6 +84,7 @@ def test_create_tarfile(monkeypatch, settings, format, extension):
     os.remove(proof_file)
     os.remove(symlink_file)
 
+
 def test_create_zipfile_must_fail(monkeypatch, tmpdir):
     monkeypatch.setattr('ideascube.serveradmin.backup.Backup.ROOT', str(tmpdir))
     monkeypatch.setattr('ideascube.serveradmin.backup.Backup.FORMAT', 'zip')
@@ -95,6 +93,7 @@ def test_create_zipfile_must_fail(monkeypatch, tmpdir):
         Backup.create()
     assert len(list(Backup.list())) == 0
 
+
 @pytest.mark.parametrize('extension', [
     ('.zip'),
     ('.tar'),
diff --git a/ideascube/serveradmin/tests/test_catalog.py b/ideascube/serveradmin/tests/test_catalog.py
index 7190417..c5c8914 100644
--- a/ideascube/serveradmin/tests/test_catalog.py
+++ b/ideascube/serveradmin/tests/test_catalog.py
@@ -610,8 +610,8 @@ def test_mediacenter_installs_zippedmedia(tmpdir, settings, zippedmedia_path):
     assert video.kind == Document.VIDEO
     assert Document.objects.search('summary').count() == 3
 
-    packaged_documents = list(Document.objects.filter(package_id='test-media'))
-    assert len(packaged_documents) == 3
+    packaged_documents = Document.objects.filter(package_id='test-media')
+    assert packaged_documents.count() == 3
 
     # Be sure that referenced documents are the ones installed by the package
     # and are not copied somewhere by the django media system.
diff --git a/ideascube/settings.py b/ideascube/settings.py
index 0d9996e..58c10f4 100644
--- a/ideascube/settings.py
+++ b/ideascube/settings.py
@@ -24,10 +24,11 @@ log('IDEASCUBE_ID={}'.format(IDEASCUBE_ID))
 # local config, and easier to manage code upgrade.
 try:
     sub = importlib.import_module(".conf." + IDEASCUBE_ID, package="ideascube")
-except Exception as e:
-    if not isinstance(e, ImportError):
-        sys.stderr.write(e)
-    from .conf import dev as sub
+
+except ImportError:
+    # No specific config for this box
+    from .conf import base as sub
+
 finally:
     # Make it available as a settings, to be able to display it in the admin.
     SETTINGS_MODULE = sub.__name__
diff --git a/manage.py b/manage.py
index dd30842..ffa5cf0 100755
--- a/manage.py
+++ b/manage.py
@@ -4,6 +4,7 @@ import sys
 
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ideascube.settings")
+    os.environ.setdefault("IDEASCUBE_ID", "dev")
 
     from django.core.management import execute_from_command_line
 
```

Most differences are actually changes that happened on master, outside of this pull request.

The things I actually changed are adding some empty lines to appease the pep8 gods, ued a kwarg somewhere to make the code a bit more explicit, used the `exist_ok=True` thing from Python 3.4, and explained why we dropped Zip in the commit message.

The tests fail now, as I explained in a previous comment, but I'm going to submit individual pull requests for each change now, and I'll fix the tests at the same time.

bochecha:
The 5 small pull requests that were split out of this one are now all merged.

Hurray!

Closing this.

PR #352, opened 2016-05-29
catalog: Allow reinstalling packages

bochecha:
Eventually, we're going to a place where we backup the list of installed content, but not the actual installed content.

Thus, when restoring from a backup, we will need a way to reinstall the content that the catalog thinks is installed.

More generally, this could also be useful if for some reason a package had been badly installed.


yohanboniface:
No tests? :)

bochecha:
> No tests? :)

The commit adds test on the new API.

:whistle: :hopes_nobody_saw_anything:

PR #355, opened 2016-05-30
Use a fresh database for every test

bochecha:
Our database is in memory, so it doesn't have any impact on execution time.


yohanboniface:
Excellent!

PR #358, opened 2016-06-07
Do not fail if we try to add the same remote twice. (fix #357)

### LINKED ISSUE #357: ideascube should not issue a fatal error when a remote already exists
### Actual result

    # ideascube catalog remotes add "Kiwix" "Kiwix ZIM content" http://catalog.ideascube.org/kiwix.yml
    IDEASCUBE_ID=idb_jor_alhashmi
    Importing settings from ideascube.conf.idb_jor_alhashmi
    
    CommandError: There already is a "Kiwix" remote
    # echo $?
    1

This should not raise a fatal error. If the remote already exists, fine, let's move forward then!

### Expected result

    # ideascube catalog remotes add "Kiwix" "Kiwix ZIM content" http://catalog.ideascube.org/kiwix.yml
    IDEASCUBE_ID=idb_jor_alhashmi
    Importing settings from ideascube.conf.idb_jor_alhashmi
    
    CommandWarning: "Kiwix" remote: been there, done that. Let's move forward.
    # echo $?
    0

mgautierfr:
If we try to add a remote already existing (same id) with same url,
print a warning.

If we try to add a remote already exsting but with different url,
raise as usual.

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -48,6 +48,12 @@ class NoSuchPackage(Exception):
 class InvalidHandlerType(Exception):
     pass
 
+class ExistingRemoteError(Exception):
+    def __init__(self, remote):
+        self.remote = remote
+
+    def __str__(self):
+        return "Remote {} already exists".format(self.remote.id)
 
 class Remote:
     def __init__(self, id, name, url):
@@ -611,8 +617,9 @@ class Catalog:
         return sorted(self._remotes.values(), key=attrgetter('id'))
 
     def add_remote(self, id, name, url):
-        if id in self._remotes:
-            raise ValueError('There already is a "{}" remote'.format(id))
+        existing_remote = self._remotes.get(id)
+        if existing_remote:
+            raise ExistingRemoteError(existing_remote)
 
         r = Remote(id, name, url)
         r.to_file(os.path.join(self._cache_remote_dir, '{}.yml'.format(id)))
--- a/ideascube/serveradmin/management/commands/catalog.py
+++ b/ideascube/serveradmin/management/commands/catalog.py
@@ -2,7 +2,9 @@ import argparse
 
 from django.core.management.base import BaseCommand, CommandError
 
-from ideascube.serveradmin.catalog import Catalog, NoSuchPackage
+from ideascube.serveradmin.catalog import (Catalog,
+                                           NoSuchPackage,
+                                           ExistingRemoteError)
 
 
 class Command(BaseCommand):
@@ -201,10 +203,18 @@ class Command(BaseCommand):
             self.catalog.add_remote(
                 options['id'], options['name'], options['url'])
 
-        except ValueError as e:
-            raise CommandError(e)
+            self.catalog.update_cache()
 
-        self.catalog.update_cache()
+        except ExistingRemoteError as e:
+            if e.remote.url != options['url']:
+                raise CommandError(
+                         ('There already is a "{0.id}" remote'
+                          'and urls differ ({0.url} and {1})'
+                         ).format(e.remote, options['url'])
+                      )
+            # Just print a warning if we try to add the exact same remote twice.
+            print(('Warning: "{}" remote already exists. '
+                   "They have same urls, so let's move forward.").format(id))
 
     def remove_remote(self, options):
         try:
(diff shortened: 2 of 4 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -611,8 +611,20 @@ def list_remotes(self):
         return sorted(self._remotes.values(), key=attrgetter('id'))
 
     def add_remote(self, id, name, url):
-        if id in self._remotes:
-            raise ValueError('There already is a "{}" remote'.format(id))
+        try:
+            existing_remote = self._remotes[id]
```
yohanboniface: I'd do `self._remotes.get(id)` to avoid the try/catch and indent.

ideascube/serveradmin/tests/test_catalog_command.py
```diff
@@ -90,10 +90,13 @@ def test_cannot_add_duplicate_remote(tmpdir, settings, monkeypatch):
 
     old_mtime = tmpdir.join('remotes', 'foo.yml').mtime()
 
+    call_command(
```
yohanboniface: Isn't this already added in line 84?

ideascube/serveradmin/tests/test_catalog_command.py
```diff
@@ -90,10 +90,13 @@ def test_cannot_add_duplicate_remote(tmpdir, settings, monkeypatch):
...
+        'catalog', 'remotes', 'add', remote['id'], remote['name'],
+        remote['url'])
     with pytest.raises(CommandError):
         call_command(
             'catalog', 'remotes', 'add', remote['id'], remote['name'],
-            remote['url'])
+            remote['url']+"bad")
```
yohanboniface: Spaces around `+` ;)

ideascube/serveradmin/catalog.py
```diff
@@ -611,8 +611,20 @@ def list_remotes(self):
...
+            if url != existing_remote.url:
+                raise ValueError(
+                         ('There already is a "{0.id}" remote'
+                          'and urls differ ({0.url} and {1}'
+                         ).format(existing_remote, url)
+                      )
+
+            print(('CommandWarning: "{}" remote already exists.\n'
```
yohanboniface: Not sure about "CommandWarning" prefix. I'd say simply "Warning". Or we should a consistent usage in all messages, no?

ideascube/serveradmin/catalog.py
mgautierfr: If you want, but it will not save a lot of indent cause of the
```
if existing_remote:
    …
```
However, it is maybe better to use get(). try/catch may imply that is it the normal code path and, here, it is the integrity check/warning.

ideascube/serveradmin/tests/test_catalog_command.py
mgautierfr: Yes, this the goal : testing that 2 add of the same repository do not fail :)

mgautierfr:
Hum, maybe it is better to raise in `Catalog.add_remote` all the time and let's the command call transform the exception to a warning if needed.
`add_remote` doesn't know the context of the call and should not print the warning (and even less **Command**Warning).
Or return a bool to advert that insert really happen.

What do you think ?

bochecha:
> add_remote doesn't know the context of the call and should not print the warning (and even less CommandWarning).

This.

yohanboniface:
Humm, good point, I guess that means moving the check on the command (so the module always raise, and the command do the check on the URL to know what to do: print or reraise).
In the future we'll call the module from an UI, but we will certainly want another kind of user feedback.

ideascube/serveradmin/tests/test_catalog_command.py
yohanboniface: Ah, got it :)
Can you add a quick comment for future me that will forget that again? :)
Something like: `# Trying to add the same remote with the same URL should not fail`.

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -201,10 +203,18 @@ def add_remote(self, options):
             self.catalog.add_remote(
                 options['id'], options['name'], options['url'])
 
-        except ValueError as e:
-            raise CommandError(e)
-
-        self.catalog.update_cache()
+            self.catalog.update_cache()
```
bochecha: Just a nit: I'd put this line in an `else` block following the `except one`.

The reason is that only the `add_remote` line may raise an `ExistingRemoteError`, which is made clearer by the `else` block I suggest.

I'd consider it more important if you weren't specifically catching an `ExistingRemoteError` which makes no sense in the context of the `update_cache` method, though, so I wouldn't block the review on this.

yohanboniface:
Rebased and merged directly: https://github.com/ideascube/ideascube/commit/5eaeeffdc4f5af07456e3cf42c043f583912238a :)

PR #366, opened 2016-06-24
Reset page parameter when changing search filters (fix #364)

### LINKED ISSUE #364: Media center bug on tag and page
On the media center, while being on page 2 or 3, selecting a tag raise the following error
```
Page not found (404)
Request Method: 	GET
Request URL: 	http://koombook.lan/es/mediacenter/?tags=orinoco&page=2
Raised by: 	ideascube.mediacenter.views.Index
```
This error does not occur on page 1

yohanboniface:
Fix #364

### DIFF
--- a/ideascube/mediacenter/tests/test_views.py
+++ b/ideascube/mediacenter/tests/test_views.py
@@ -413,6 +413,27 @@ def test_by_tag_page_is_paginated(app, monkeypatch):
     app.get(url + '?page=3', status=200)
 
 
+def test_changing_filter_reset_page_parameter(app, monkeypatch):
+    monkeypatch.setattr(Index, 'paginate_by', 2)
+    DocumentFactory.create_batch(size=4, tags=['plane', 'car'],
+                                 kind=Document.PDF)
+    DocumentFactory.create_batch(size=4, tags=['plane'], kind=Document.IMAGE)
+    DocumentFactory.create_batch(size=2)
+    url = reverse('mediacenter:index')
+    response = app.get(url, {'page': 2, 'tags': 'plane', 'kind': 'pdf'})
+    find = response.pyquery.find
+    # Removing 'tags' filter.
+    assert 'page' not in find("a.flatlist:contains('plane')")[0].attrib['href']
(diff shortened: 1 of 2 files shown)


yohanboniface:
@mgautierfr @bochecha r?

bochecha:
I'm a bit worried by the way you `pop` the `page` querystring parameter, but that's probably only because I lack enough context to judge.

What I'm seeing are 3 functions to change the querystring: add some key/value params, replace some key/value params, and remove some key/value params.

They are fairly generic functions operating on (sort of) dictionaries.

And then, in each of them, there is one key/value param that you single out, and handle specifically: `page`.

This raises a few questions:

1. Are we always doing pagination with a `page` parameter?

2. Are we always using the `page` parameter for pagination?

3. Do we really want to drop pagination every time we call one of these functions?

4. I wonder if it wouldn't be better to instead leave the implementation of these functions as they are, and instead when calling them explicitly reset the `page` parameter when needed?

Again, I might be worrying over nothing because I don't know the context well and don't grasp all the implications of this. So feel free to reassure me that this is the right thing to do. :smile:

yohanboniface:
> Are we always doing pagination with a page parameter?

That should be, given it's factorized in the `pagination.html` template.

> Are we always using the page parameter for pagination?

Humm, good question. As for now yes, I can certainly imagine a use case where `page` would mean something else, but honestly I don't think we'll hit it soon.

> Do we really want to drop pagination every time we call one of these functions?

Good question again ;) Changing the query string means changing the page content, I don't think it makes a lot of sense to change the page content while staying on the same page number (different than 1). So while the question is legit, I'm confident it's quite safe.

> I wonder if it wouldn't be better to instead leave the implementation of these functions as they are, and instead when calling them explicitly reset the page parameter when needed?

It's not really convenient in Django to try to combine two templatetags, and I haven't found any good option myself. Do you have an idea in mind?

bochecha:
> Do you have an idea in mind?

None. :)

Overall, this looks good, the questions above were really just questions I had, but you seem to have thought about it carefully, so let's merge this.

PR #371, opened 2016-06-27
Add "order by" selector in list views (fix #363 fix #368)

### LINKED ISSUE #363: Order medias in mediacenter by date or name
Medias in the mediacenter are ordered by last creation date. This is not always relevant. It could be useful to allow the user to choose to order medias by name or by date, in the list pages.

### LINKED ISSUE #368: blog post order
Blog application : blog posts are ordered by first creation date. Meaning : the first created blogpost will always stick to the top of the blog home page and the latest created blopost will be burried and not really accessible.

capture du 2016-06-26 11-20-48

yohanboniface:
Fix #363 fix #368.

- adds a "order by" box in the mediacenter, blog and library pages, which shows the current ordering, and the other available
- when clicking on the active ordering, the sort is reversed
- the possibility to define for each home page which orders are possible, including the default sort for each one

Preview:
screenshot from 2016-06-27 23-23-03

### DIFF
--- a/ideascube/mixins.py
+++ b/ideascube/mixins.py
@@ -11,6 +11,35 @@ from taggit.models import Tag
 from ideascube.search.models import Search
 
 
+class OrderableViewMixin:
+
+    ORDERS = []
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self.ORDERS:
+            default = self.ORDERS[0]
+            context['order_by'] = (self.request.GET.get('order_by') or
+                                   default['key'])
+            context['sort'] = self.request.GET.get('sort') or default['sort']
+        return context
+
+    def get_ordering(self):
+        if not self.ORDERS:
+            return None
+        key = self.request.GET.get('order_by')
+        for order in self.ORDERS:
+            if order['key'] == key:
+                break
+        else:
+            order = self.ORDERS[0]
+        order_by = order['expression']
+        sort = self.request.GET.get('sort', order['sort'])
+        if sort == 'desc':
+            order_by = order_by.desc()
+        return [order_by]
+
+
 class FilterableViewMixin:
 
     def _search_for_attr_from_context(self, attr, context):
@@ -60,7 +89,6 @@ class FilterableViewMixin:
         kind = self.request.GET.get('kind')
         lang = self.request.GET.get('lang')
         tags = self.request.GET.getlist('tags')
-
         if any((query, kind, lang, tags)):
             qs = qs.search(query=query, lang=lang, kind=kind, tags=tags)
         return qs
--- a/ideascube/blog/templates/blog/index.html
+++ b/ideascube/blog/templates/blog/index.html
@@ -20,6 +20,7 @@
     {% url 'blog:index' as index %}
     {% include 'search/filter_cloud.html' %}
     {% include 'search/box.html' with action=index %}
+    {% include 'search/order_by.html' %}
     {% include 'search/filter_by_lang.html' %}
     {% tag_cloud url="blog:index" model=view.model tags=available_tags %}
 {% endblock third %}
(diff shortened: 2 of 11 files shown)


yohanboniface:
Anyone for a review? :)

ideascube/mixins.py
```diff
@@ -11,6 +11,36 @@
...
 
 
+class OrderableViewMixin:
+
+    ORDERS = []
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self.ORDERS:
+            context['orders'] = self.ORDERS
```
bochecha: Did you know? http://reinout.vanrees.org/weblog/2014/05/19/context.html

That is, you could just remove this line, and in the template you could use `{{ view.ORDERS }}` directly.

ideascube/mixins.py
```diff
@@ -11,6 +11,36 @@
...
+            default = self.ORDERS[0]
+            context['order_by'] = (self.request.GET.get('order_by') or
+                                   default['key'])
+            context['sort'] = self.request.GET.get('sort') or default['sort']
+        return context
+
+    def get_ordering(self):
+        key = self.request.GET.get('order_by')
+        orders = OrderedDict((o['key'], o) for o in self.ORDERS)
```
bochecha: You don't need an OrderedDict here, a simple dict would be just enough:

```python
        orders = {o['key']: o for o in self.ORDERS}
```

But going further, you don't really need the dict either:

```python
    def get_ordering(self):
        key = self.request.GET.get('order_by')

        for o in self.ORDERS:
            if o['key'] == key:
                order = o
                break
        elif self.ORDERS:
            order = self.ORDERS[0]
        else:
            return None

        order_by = order['expression']
        ...
```

I'd personally also get rid of the case where `self.ORDERS` evaluates to False as early as possible, right at the beginning of this function, because there's no point even looking at the `order_by` passed as `GET` if we can't order by anything anyway:

```python
    def get_ordering(self):
        if not self.ORDERS:
            return

        key = self.request.GET.get('order_by')

        for o in self.ORDERS:
            if o['key'] == key:
                order = o
                break
        else:
            order = self.ORDERS[0]

        order_by = order['expression']
        ...
```

But that's mere personal preference, not something critical.

bochecha:
Any reason the commit message says `[WIP]` ?

bochecha:
Looks pretty good overall, except for three very minor things.

Nothing critical though, I don't think I'd block merging this pull request until you "fixed" them, they are just something I felt were worth mentioning.

So as far as I'm concerned, feel free to merge this, with or without acting on my review. :smiley:

ideascube/mixins.py
yohanboniface: Oh, right, good point! :)

yohanboniface:
> You don't need an OrderedDict here

Good catch :) I've used an OrderedDict in my first iteration because I was then passing it to the context.
Anyway, refactored a bit and pushed a new version :)

mgautierfr:
If the 'sort' key is used to only store a binary value ('asc' or 'desc') it may be better to use a boolean ?

GET paramaters has no concept of boolean so it will be a bit tricky to parse argument in the mixin but elsewhere it makes more sense to use a boolean instead of a string.

yohanboniface:
I'm not sure it's explicit enough: will 1 be asc or desc? :)

mgautierfr:
You can rename the key to sort_descending or descending_order

bochecha:
> GET paramaters has no concept of boolean so it will be a bit tricky to parse argument in the mixin but elsewhere it makes more sense to use a boolean instead of a string.

Or, since we have Python 3.4, an [enum](https://docs.python.org/3/library/enum.html)

bochecha:
Discussing this, we agreed that even though the string parameter wasn't ideal, we weren't entirely sure whether making it a boolean or an enum would really improve things much.

So let's merge this, we can always improve it later on.

PR #378, opened 2016-07-07
Catalog do not fail when listing with a unknown type package.

### LINKED ISSUE #359: Catalog should not fail if there is package type not handle by current version.
If there are packages in a remote and the catalog doesn't handle this kind of package, catalog should not fail but ignore those packages.

For example with a version not handling the `zipped-media` type and a remote with `zipped-media` packages :

## Actual result

```
$ python manage.py catalog list
IDEASCUBE_ID=matbook
Importing settings from ideascube.conf.dev

  File "/home/mgautier/Project/BSF/ideascube/ideascube/serveradmin/catalog.py", line 347, in _get_package
    return Package.registered_types[type](id, metadata)
KeyError: 'zipped-media'

During handling of the above exception, another exception occurred:

  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/mgautier/.virtualenvs/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/home/mgautier/.virtualenvs/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/mgautier/.virtualenvs/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/mgautier/.virtualenvs/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/home/mgautier/Project/BSF/ideascube/ideascube/serveradmin/management/commands/catalog.py", line 112, in handle
    options['func'](options)
  File "/home/mgautier/Project/BSF/ideascube/ideascube/serveradmin/management/commands/catalog.py", line 126, in list_packages
    pkgs = self.catalog.list_installed(options['ids'])
  File "/home/mgautier/Project/BSF/ideascube/ideascube/serveradmin/catalog.py", line 428, in list_installed
    ids, self._catalog['installed'], fail_if_no_match=False)
  File "/home/mgautier/Project/BSF/ideascube/ideascube/serveradmin/catalog.py", line 359, in _get_packages
    pkgs.append(self._get_package(id, source))
  File "/home/mgautier/Project/BSF/ideascube/ideascube/serveradmin/catalog.py", line 350, in _get_package
    raise InvalidPackageType(type)
ideascube.serveradmin.catalog.InvalidPackageType: zipped-media
```

## Expected result

```
$ python manage.py catalog list
IDEASCUBE_ID=matbook
Importing settings from ideascube.conf.dev

Not handled packages
 media.test (zipped-media)      2016-03-18    12.4 MB   Base test content
```

mgautierfr:
Fix #359

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -467,18 +467,31 @@ class Catalog:
         except KeyError:
             raise InvalidPackageType(type)
 
-    def _get_packages(self, id_patterns, source, fail_if_no_match=True):
+    def _get_packages(
+        self, id_patterns, source,
+        fail_if_no_match=True, fail_if_invalid=True):
+
         pkgs = []
 
         for id_pattern in id_patterns:
             if '*' in id_pattern:
                 for id in source:
                     if fnmatch(id, id_pattern):
-                        pkgs.append(self._get_package(id, source))
+                        try:
+                            pkgs.append(self._get_package(id, source))
+                        except InvalidPackageType as e:
+                            if fail_if_invalid:
+                                raise e
 
             else:
                 try:
-                    pkgs.append(self._get_package(id_pattern, source))
+                    try:
+                        pkgs.append(self._get_package(id_pattern, source))
+                    except InvalidPackageType as e:
+                        if fail_if_invalid:
+                            raise e
+                        print("Package {} is present but not handled"
+                              .format(id_pattern))
 
                 except NoSuchPackage as e:
                     if fail_if_no_match:
@@ -543,23 +556,38 @@ class Catalog:
 
     def list_installed(self, ids):
         pkgs = self._get_packages(
-            ids, self._installed, fail_if_no_match=False)
+            ids, self._installed,
+            fail_if_no_match=False, fail_if_invalid=False)
         return sorted(pkgs, key=attrgetter('id'))
 
     def list_available(self, ids):
         pkgs = self._get_packages(
-            ids, self._available, fail_if_no_match=False)
+            ids, self._available,
+            fail_if_no_match=False, fail_if_invalid=False)
         return sorted(pkgs, key=attrgetter('id'))
 
     def list_upgradable(self, ids):
         pkgs = []
 
         for ipkg in self._get_packages(
-                ids, self._installed, fail_if_no_match=False):
+                ids, self._installed,
+                fail_if_no_match=False, fail_if_invalid=False):
             upkg = self._get_package(ipkg.id, self._available)
 
             if ipkg != upkg:
                 pkgs.append(upkg)
+
+        return sorted(pkgs, key=attrgetter('id'))
+
+    def list_nothandled(self, ids):
+        pkgs = []
+        source = dict(self._available)
+        source.update(self._installed)
+        for pkgid, metadata in source.items():
+            try:
+                self._get_package(pkgid, source)
+            except InvalidPackageType:
+                pkgs.append(Package(pkgid, metadata))
 
         return sorted(pkgs, key=attrgetter('id'))
 
--- a/ideascube/serveradmin/management/commands/catalog.py
+++ b/ideascube/serveradmin/management/commands/catalog.py
@@ -49,6 +49,10 @@ class Command(BaseCommand):
         group.add_argument(
             '--all', action='store_const', dest='filter', const='all',
             help='List available and installed packages (default)')
+        group.add_argument(
+            '--unhandled', action='store_const', dest='filter',
+            const='unhandled',
+            help='List unhandled packages.')
         list.set_defaults(filter='all', func=self.list_packages)
 
         install = subs.add_parser(
@@ -120,7 +124,20 @@ class Command(BaseCommand):
             print("Add a remote source with:\n")
             print("    {} remotes add ID NAME URL\n".format(self.parser.prog))
 
-        fmt = ' {0.id:20}  {0.version!s:12}  {0.filesize:8}  {0.name}'
+        fmt = (' {0.id:20}  {0.version!s:12}  {0.filesize:8}'
+               ' {0.type:15} {0.name}')
+
+        unhandled_pkgs = self.catalog.list_nothandled('*')
+        if unhandled_pkgs:
+            if options['filter'] == 'unhandled':
+                print('Not handled packages')
+
+                for pkg in unhandled_pkgs:
+                    print(fmt.format(pkg))
+            else:
+                print(("Warning: Ignoring {} unsupported package(s)\n"
+                       "Use '--unhandled' for details.\n\n"
+                      ).format(len(unhandled_pkgs)))
 
         if options['filter'] in ('all', 'installed'):
             pkgs = self.catalog.list_installed(options['ids'])
@@ -169,7 +186,7 @@ class Command(BaseCommand):
 
     def reinstall_packages(self, options):
         try:
-            self.catalog.reinstall_packages(oprions['ids'])
+            self.catalog.reinstall_packages(options['ids'])
 
         except NoSuchPackage as e:
             raise CommandError('No such package: {}'.format(e))
--- a/ideascube/serveradmin/tests/test_catalog.py
+++ b/ideascube/serveradmin/tests/test_catalog.py
@@ -1894,3 +1894,59 @@ def test_catalog_list_upgradable_packages(
     assert pkg.version == '2015-09'
     assert pkg.size == '200KB'
     assert isinstance(pkg, ZippedZim)
+
+
+def test_catalog_list_nothandled_packages(
+        tmpdir, settings, testdatadir, mocker):
+    from ideascube.serveradmin.catalog import Catalog
+
+    zippedzim = testdatadir.join('catalog', 'wikipedia.tum-2015-08')
+    path = tmpdir.mkdir('source').join('wikipedia_tum_all_nopic_2015-08.zim')
+    zippedzim.copy(path)
+
+    remote_catalog_file = tmpdir.join('source').join('catalog.yml')
+    with remote_catalog_file.open(mode='w') as f:
+        f.write('all:\n')
+        f.write('  wikipedia.tum:\n')
+        f.write('    version: 2015-08\n')
+        f.write('    size: 200KB\n')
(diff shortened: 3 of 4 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -351,18 +351,31 @@ def _get_package(self, id, source):
         except KeyError:
             raise InvalidPackageType(type)
 
-    def _get_packages(self, id_patterns, source, fail_if_no_match=True):
+    def _get_packages(
+        self, id_patterns, source,
+        fail_if_no_match=True, fail_if_invalid_package=True):
```
yohanboniface: `fail_if_invalid_package` => `fail_if_invalid` given that we are talking about packages?

BTW, I usually think that having two boolean kwargs on the same method is a refactor needed alert.

ideascube/serveradmin/catalog.py
```diff
@@ -427,26 +440,41 @@ def _progress(*args):
...
+                ids, self._catalog['installed'],
+                fail_if_no_match=False, fail_if_invalid_package=False):
             upkg = self._get_package(ipkg.id, self._catalog['available'])
 
             if ipkg != upkg:
                 pkgs.append(upkg)
 
         return sorted(pkgs, key=attrgetter('id'))
 
+    def list_nothandled(self, ids):
```
yohanboniface: I'm not sure we should go to the point of listing not handled. Maybe not listing them at all is enough?
What would do a box admin with the detailed info of non handled packages?
Sure they can update the server, but in the target I'm not sure those will same people (installing packages and upgrading system).

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -149,6 +150,14 @@ def list_packages(self, options):
                 for pkg in pkgs:
                     print(fmt.format(pkg))
 
+        pkgs = self.catalog.list_nothandled(options['ids'])
+
+        if pkgs:
+            print('Not handled packages')
```
yohanboniface: If we decide to display the list of not handled, puting them at the end make them the more visible, given that the terminal is autoscrolling.
Maybe we could just have a warning before the table like "Skipping unsupported content: id1, id2, di3" ?

ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -1826,3 +1827,65 @@ def test_catalog_list_upgradable_packages(
...
+        tmpdir, settings, testdatadir, mocker):
+    from ideascube.serveradmin.catalog import Catalog
+
+    zippedzim = testdatadir.join('catalog', 'wikipedia.tum-2015-08')
+    path = tmpdir.mkdir('source').join('wikipedia_tum_all_nopic_2015-08.zim')
+    zippedzim.copy(path)
+
+    remote_catalog_file = tmpdir.join('source').join('catalog.yml')
+    with remote_catalog_file.open(mode='w') as f:
+        f.write('all:\n')
```
yohanboniface: Not big fan of calling write for each line. yaml.safe_dumps should take a dict and make the code a bit more elegant :)

ideascube/serveradmin/catalog.py
bochecha: Having the first `fail_if_no_match` boolean already was something I found pretty bad... when I introduced it.

I agree that this code needs refactoring, but maybe let's hold off on it a bit, because lots of things will need to change when we start using the catalog in the web app anyway. (multiple simultaneous accesses anyone?)

ideascube/serveradmin/catalog.py
bochecha: In general I have reservations about silently ignoring anything.

Finding unhandled packages in a catalog you're consuming is a clear sign that something isn't right and ought to be fixed, among others:

* the catalog provider made a typo
* the client is out of date

Maybe listing the unhandled packages isn't the right thing to do, I don't know. Should we instead just display a warning, something like `Warning: Ignoring unhandled package 'foo' ('unhandled-type')` just as the problematic package is encountered? (that is, before reporting the list)

But I would argue that (at least for the moment, while things are still being fleshed out), we really want to know that some packages in the catalog can't be installed, as it's a sign we made a mistake and we'll want to fix it.

ideascube/serveradmin/catalog.py
bochecha: For the record, though: at the time I introduced this boolean I was still growing the features organically. (I was writing this package manager, after all)

In hindsight, I'm starting to think that validation of the packages should not be done when listing/installing/... as we do now, but instead when refreshing the cache.

That is, when downloading the remote's `yml` file, instead of just storing it as it is, this is when we could try and parse it, look at all the packages, drop the invalid/unsupported ones (warning about them), and then locally keep only the "good" packages.

I guess I just should open a ticket about that, though. :smile:

ideascube/serveradmin/tests/test_catalog.py
bochecha: Also, we're pretty much cargo-culting all those tests through copy-pasta at this point, but when I wrote the first few ones, I found `with file.open(...)` was mighty cool and `py.path` was great for offering this API.

Since then, I discovered it also offered a `file.write(...)` API, making all those `with` blocks look silly. :smile:

ideascube/serveradmin/catalog.py
bochecha: An alternative would be to [validate the data when updating the cache](https://github.com/ideascube/ideascube/issues/381).

This way, unsupported package types just wouldn't land in the local cache.

... which renders this whole pull request moot. :disappointed: 

Sorry, I should have thought about that earlier.

bochecha: FTR: I have been convinced by the others that even if we had pre-validation of the catalog cache, we'd still need this code to not crash on unhandled package types.

The obvious use case is a server that gets updated, the new code has **dropped** support for an old package type, but the local cache still has packages available with that old type.

yohanboniface:
can you rebase this PR on master?
Also, it seems some comments needs action, whether to refute them or do the suggestions :)

ideascube/serveradmin/tests/test_catalog.py
bochecha: I also remember specifically not wanting to use `yaml.safe_dumps` here, because... Well I don't remember why.

Maybe I thought that since the API being tested uses `yaml`, it was better to have the tests use something else? (to avoid making the same mistake in the tests as in the API, which would then cover up that mistake, getting the tests to pass)

In hindsight, that would have been a poor reason in this specific case.

But I can't remember if that was it, or if I had a really good reason. :disappointed:

ideascube/serveradmin/catalog.py
mgautierfr: I'm also not fan of silently ignoring content.

- This can be useful for debug/analysis ("hello devs, I can't install content on my box, and I know that this content is existing")
- This is a clue that something is not up to date.

ideascube/serveradmin/management/commands/catalog.py
mgautierfr: Agree, we can put the warning on top instead of the end of the screen.

ideascube/serveradmin/tests/test_catalog.py
mgautierfr: Not fan neither, but they were their before :)

I suggest that we open a new bug to fix this f.write in all tests and keep this specific test coherent with others for this pr.

(I'm not against fixing this, even for the next release. But this is two separated things, so two separated pr/bugs)

ideascube/serveradmin/management/commands/catalog.py
mgautierfr: Done on last "push --force"

By default, only display that x packages have been ignored.
With option --unhandled, the full list is displayed.

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -120,7 +124,22 @@ def list_packages(self, options):
...
+            if options['filter'] == 'unhandled':
+                print('Not handled packages')
+
+                for pkg in unhandled_pkgs:
+                    print(fmt.format(pkg))
+            else:
+                print(("Warning: There is {} packages ignored by ideascube "
+                       "cause it doesn't know how to handle them.\n"
+                       "Use the command 'list --unhandled' to get the full "
+                       "list of unhandled packages.\n\n"
```
bochecha: "There is" followed by plural form "packages", is incorrect.

"cause" is oral slang, not correct English.

Lots of repetitions of "list" and "unhandled".

How about `Warning: Ignoring {} unsupported packages\nUse '--unhandled' for details.\n\n` ?

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -169,7 +188,7 @@ def remove_packages(self, options):
 
     def reinstall_packages(self, options):
         try:
-            self.catalog.reinstall_packages(oprions['ids'])
+            self.catalog.reinstall_packages(options['ids'])
```
bochecha: This one hurts. :disappointed:

ideascube/serveradmin/tests/test_catalog.py
bochecha: Opened the issue: https://github.com/ideascube/ideascube/issues/406

ideascube/serveradmin/catalog.py
yohanboniface: I'd still rename in `fail_if_invalid` in the meantime.

ideascube/serveradmin/management/commands/catalog.py
mgautierfr: Agree.
Used "Warning: Ignoring {} unsupported package(s)\nUse '--unhandled' for details.\n\n" in last push --force

ideascube/serveradmin/catalog.py
mgautierfr: Agree.
Changed in last 'push --force'

bochecha:
All the comments raised during review are addressed as far as I can see, merging. :+1:

PR #379, opened 2016-07-07
Split catalog cache

### LINKED ISSUE #376: Split the catalog cache
Today the catalog caches a few things:

1. the list of configured remotes
2. the list of packages available in each remote
3. the downloaded package files
4. the list of installed packages

Items 2 and 3 are clearly cache: if they get dropped, then they can be obtained again by updating the cache of the catalog or redownloading the package(s).

However, items 1 and 4 are really local data, and as such they would need to be backed up. This way, we could only backup the fact that a package is installed and not the actual contents, and when restoring from a backup we'd just have to reinstall the package.

So we need to split the current catalog cache, and move some of it to `BACKUPED_ROOT`.

bochecha:
Fixes #376 

I strongly suggest you review this one commit at a time, as it will make it much easier to grasp the changes. :wink:

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -415,12 +415,14 @@ class Catalog:
         self._cache_root = settings.CATALOG_CACHE_ROOT
         os.makedirs(self._cache_root, exist_ok=True)
 
-        self._cache_remote_dir = os.path.join(self._cache_root, 'remotes')
-        self._load_remotes()
-
         self._catalog_cache = os.path.join(self._cache_root, 'catalog.yml')
         self._local_package_cache = os.path.join(self._cache_root, 'packages')
+        os.makedirs(self._local_package_cache, exist_ok=True)
 
+        self._cache_remote_dir = os.path.join(self._cache_base_dir, 'remotes')
+        os.makedirs(self._cache_remote_dir, exist_ok=True)
+
+        self._load_remotes()
         self._load_catalog()
 
         self._bar = Bar()
@@ -663,7 +665,6 @@ class Catalog:
             self._persist_catalog()
 
         self._package_caches = [self._local_package_cache]
-        os.makedirs(self._local_package_cache, exist_ok=True)
 
     def _persist_catalog(self):
         with open(self._catalog_cache, 'w') as f:
@@ -704,8 +705,6 @@ class Catalog:
 
     # -- Manage remote sources ------------------------------------------------
     def _load_remotes(self):
-        os.makedirs(self._cache_remote_dir, exist_ok=True)
-
         self._remotes = {}
 
         for path in glob(os.path.join(self._cache_remote_dir, '*.yml')):
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -422,7 +422,8 @@ class Catalog:
         self._local_package_cache = os.path.join(
             self._cache_base_dir, 'packages')
 
-        self._load_cache()
+        self._load_catalog()
+
         self._bar = Bar()
 
     def _progress(self, msg, i, chunk_size, remote_size):
@@ -573,7 +574,7 @@ class Catalog:
             used_handlers[handler.__class__.__name__] = handler
             self._catalog['installed'][pkg.id] = (
                 self._catalog['available'][pkg.id])
-            self._persist_cache()
+            self._persist_catalog()
 
         for handler in used_handlers.values():
             handler.commit()
@@ -592,7 +593,7 @@ class Catalog:
                 continue
             used_handlers[handler.__class__.__name__] = handler
             del(self._catalog['installed'][pkg.id])
-            self._persist_cache()
+            self._persist_catalog()
 
         if not commit:
             return
@@ -646,13 +647,13 @@ class Catalog:
 
             self._catalog['installed'][ipkg.id] = (
                 self._catalog['available'][upkg.id])
-            self._persist_cache()
+            self._persist_catalog()
 
         for handler in used_handlers.values():
             handler.commit()
 
     # -- Manage local cache ---------------------------------------------------
-    def _load_cache(self):
+    def _load_catalog(self):
         if os.path.exists(self._cache_catalog):
             with open(self._cache_catalog, 'r') as f:
                 self._catalog = yaml.safe_load(f.read())
@@ -660,12 +661,12 @@ class Catalog:
         # yaml.safe_load returns None for an empty file.
         if not getattr(self, '_catalog', None):
             self._catalog = {'installed': {}, 'available': {}}
-            self._persist_cache()
+            self._persist_catalog()
 
         self._package_caches = [self._local_package_cache]
         os.makedirs(self._local_package_cache, exist_ok=True)
 
-    def _persist_cache(self):
+    def _persist_catalog(self):
         with open(self._cache_catalog, 'w') as f:
             f.write(yaml.safe_dump(self._catalog, default_flow_style=False))
 
@@ -693,14 +694,14 @@ class Catalog:
 
             os.unlink(tmppath)
 
-        self._persist_cache()
+        self._persist_catalog()
 
     def clear_cache(self):
         shutil.rmtree(self._local_package_cache)
         os.mkdir(self._local_package_cache)
 
         self._catalog['available'] = {}
-        self._persist_cache()
+        self._persist_catalog()
 
     # -- Manage remote sources ------------------------------------------------
     def _load_remotes(self):
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -412,15 +412,14 @@ class Bar(ProgressBar):
 
 class Catalog:
     def __init__(self):
-        self._cache_base_dir = settings.CATALOG_CACHE_BASE_DIR
-        os.makedirs(self._cache_base_dir, exist_ok=True)
+        self._cache_root = settings.CATALOG_CACHE_ROOT
+        os.makedirs(self._cache_root, exist_ok=True)
 
-        self._cache_remote_dir = os.path.join(self._cache_base_dir, 'remotes')
+        self._cache_remote_dir = os.path.join(self._cache_root, 'remotes')
         self._load_remotes()
 
-        self._catalog_cache = os.path.join(self._cache_base_dir, 'catalog.yml')
-        self._local_package_cache = os.path.join(
-            self._cache_base_dir, 'packages')
+        self._catalog_cache = os.path.join(self._cache_root, 'catalog.yml')
+        self._local_package_cache = os.path.join(self._cache_root, 'packages')
 
         self._load_catalog()
 
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -419,8 +419,13 @@ class Catalog:
         self._local_package_cache = os.path.join(self._cache_root, 'packages')
         os.makedirs(self._local_package_cache, exist_ok=True)
 
-        self._cache_remote_dir = os.path.join(self._cache_base_dir, 'remotes')
-        os.makedirs(self._cache_remote_dir, exist_ok=True)
+        self._storage_root = settings.CATALOG_STORAGE_ROOT
+        os.makedirs(self._storage_root, exist_ok=True)
+
+        self._installed_storage = os.path.join(
+            self._storage_root, 'installed.yml')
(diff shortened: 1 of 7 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -559,14 +559,21 @@ def upgrade_packages(self, ids):
...
+        try:
             with open(self._catalog_cache, 'r') as f:
-                self._catalog = yaml.safe_load(f.read())
+                catalog = yaml.safe_load(f.read())
 
-        # yaml.safe_load returns None for an empty file.
-        if not getattr(self, '_catalog', None):
-            self._catalog = {'installed': {}, 'available': {}}
-            self._persist_catalog()
+            # yaml.safe_load returns None for an empty file
```
yohanboniface: Move in `else` block?

ideascube/serveradmin/catalog.py
```diff
@@ -567,7 +572,24 @@ def _load_catalog(self):
 
             # yaml.safe_load returns None for an empty file
             if catalog is not None:
-                self._catalog = catalog
+                if 'available' in catalog and 'installed' in catalog:
+                    # The cache on file hasn't been splitted yet
```
yohanboniface: `yet` here seems abiguous to me: yet in stack or yet in time? I know what you mean, but if someone else than you and me read this it may not understand. I'll comment something like "we are loading a catalog in a former format" or so.

ideascube/serveradmin/catalog.py
```diff
@@ -579,7 +601,12 @@ def _load_catalog(self):
 
     def _persist_catalog(self):
         with open(self._catalog_cache, 'w') as f:
-            f.write(yaml.safe_dump(self._catalog, default_flow_style=False))
+            f.write(yaml.safe_dump(
+                self._catalog['available'], default_flow_style=False))
+
+        with open(self._installed_storage, 'w') as f:
+            f.write(yaml.safe_dump(
```
yohanboniface: Tempted to have a local helper like

```
def dump(filepath, content):
    with open(filepath, 'w') as f:
         f.write(yaml.safe_dump(content, default_flow_style=False))
```

ideascube/serveradmin/catalog.py
bochecha: Which `else` block?

ideascube/serveradmin/catalog.py
bochecha: I see what you mean, let me improve this comment.

ideascube/serveradmin/catalog.py
yohanboniface: The one after the except I'm suggesting to add :)

bochecha: And what would you put in it? :confused:

ideascube/serveradmin/catalog.py
bochecha: Fixed in the latest push to this branch.

ideascube/serveradmin/catalog.py
bochecha: Done in the latest push to this branch.

ideascube/serveradmin/catalog.py
yohanboniface: Something like that:
```
         try:
             with open(self._catalog_cache, 'r') as f:
                catalog = yaml.safe_load(f.read()) 
        except FileNotFoundError:
            # That's ok
            pass
        else:
            # yaml.safe_load returns None for an empty file
            if catalog is not None:
                self._catalog = catalog
```

bochecha: Ooooooooh! **That** `else` block!

I thought you meant a new one for the `if`, not one for the `try`!

I see what you mean, yes. And indeed, that's probably better.

bochecha: Fixed in the latest push to this branch.

ideascube/serveradmin/catalog.py
```diff
@@ -549,28 +557,58 @@ def upgrade_packages(self, ids):
...
+        self._persist_catalog()
 
         self._package_caches = [self._local_package_cache]
-        os.makedirs(self._local_package_cache, exist_ok=True)
 
-    def _persist_cache(self):
-        with open(self._cache_catalog, 'w') as f:
-            f.write(yaml.safe_dump(self._catalog, default_flow_style=False))
+    def _persist_catalog(self):
+        def persist(path, data):
```
yohanboniface: Given that it's used also somwhere else on te module, I would have made the `persist` method available as a global.

yohanboniface:
Just a thought (a bit late) about splitting the catalog yml itself: maybe we come to a simpler solution by keeping only one file, but moving it all to the backed up path?

ideascube/serveradmin/catalog.py
bochecha: Where else is it used?

bochecha:
> Just a thought (a bit late) about splitting the catalog yml itself: maybe we come to a simpler solution by keeping only one file, but moving it all to the backed up path?

I do think they should be split.

What's available is really cache: you can `catalog cache clear` and `catalog cache update` at will. Remove the file with `rm -f`, rerun `catalog cache update` and you're good to go.

But what's installed is application data, representing past user actions, and it needs to be backed up.

Conceptually, they really are different, and to me it makes sense to split them.

Plus, it's not very complex anyway. I mean look at the commit which splits them: I pretty much didn't change anything, except loading the `self._catalog` dict from two files, and then saving it to two files. (at first I had started splitting it into a `self._catalog_available` and `self._catalog_installed`, but then I realized it was too complex for no reason, and it would make the change much harder to review.)

If you truly think we should keep them together, I'm not going to fight you. :smiley:

But before I redo this, why do you think it would make more sense to keep them together? What makes you think the split is too complex?

yohanboniface:
> But before I redo this, why do you think it would make more sense to keep them together? What makes you think the split is too complex?

No strong argument. Just thinking it's simpler to manage one file than managing two files.
Let's continue anyway, now the work is done :)

ideascube/serveradmin/catalog.py
yohanboniface: Here https://github.com/ideascube/ideascube/blob/7964f4dec7a1f61039b6d9f75d89c94a605e622f/ideascube/serveradmin/catalog.py#L79 ?

bochecha: Ah, right, I missed that one. Done in the latest push to this branch.

(and now that makes me want to do the same for loading yaml from a file...)

ideascube/serveradmin/catalog.py
```diff
@@ -35,6 +35,32 @@ def rm(path):
         shutil.rmtree(path)
 
 
+def load_from_file(path):
+    """Load catalog data from a local file
```
yohanboniface: Honnestly, I think those docstrings are useless: function name and argument names should be self explanatory (and they are in this given case).
Those are usefeful only when we want to expose an API documentation, which is not the case here.

ideascube/serveradmin/catalog.py
```diff
@@ -415,14 +439,22 @@ def __init__(self):
...
-        self._load_remotes()
-
-        self._cache_catalog = os.path.join(self._cache_base_dir, 'catalog.yml')
+        self._catalog_cache = os.path.join(self._cache_base_dir, 'catalog.yml')
         self._local_package_cache = os.path.join(
             self._cache_base_dir, 'packages')
+        os.makedirs(self._local_package_cache, exist_ok=True)
+
+        self._storage_base_dir = settings.CATALOG_STORAGE_BASE_DIR
+        os.makedirs(self._storage_base_dir, exist_ok=True)
```
yohanboniface: What about renaming all those `xxx_base_dir` by `xxx_root` ?

ideascube/serveradmin/catalog.py
bochecha: The fact it returns None is the important bit.

I removed the comment in the code where that function is used (https://github.com/ideascube/ideascube/pull/379/files/6b30174a5b4c57ac3405d1a1bd34b36cdffe284f#diff-e680b7b2bed4f6d8c15c2653df1eebdfL660), because I found it much better to have it documented once in the function.

Also, if we don't want these to be exposed as API, should I make them private?

ideascube/serveradmin/catalog.py
```diff
@@ -646,28 +678,50 @@ def upgrade_packages(self, ids):
...
+        self._persist_catalog()
 
         self._package_caches = [self._local_package_cache]
-        os.makedirs(self._local_package_cache, exist_ok=True)
 
-    def _persist_cache(self):
-        with open(self._cache_catalog, 'w') as f:
-            f.write(yaml.safe_dump(self._catalog, default_flow_style=False))
+    def _persist_catalog(self):
+        persist_to_file(self._catalog_cache, self._catalog['available'])
```
yohanboniface: Shouldn't we have `self._available` and `self._installed` now instead of two keys of the same dict?

ideascube/serveradmin/catalog.py
bochecha: Then we'd get `self._storage_root`, which would be extremely confusing as it's not the same as `settings.STORAGE_ROOT` (it is below it).

ideascube/serveradmin/catalog.py
yohanboniface: > The fact it returns None is the important bit.

Agreed. This should definitely be in the docstring or comment.

> Also, if we don't want these to be exposed as API, should I make them private?

I'm talking about creating an API *documentation*, which we don't want nor need to do ;) That's fine to let the importable, even if we don't use them elsewhere.

ideascube/serveradmin/catalog.py
bochecha: We certainly could. In fact that's what I had done at first. But somehow I went back on it, preferring to keep things as they are.

What would separating them achieve?

Keeping them as is has the merit of making for a much simpler diff to review. :stuck_out_tongue_winking_eye:

ideascube/serveradmin/catalog.py
yohanboniface: Good point. But still I think `base_dir` is verbose and still a bit confusing as it's synonym of `root` anyway.
Maybe we can call that one `self._root`?

ideascube/serveradmin/catalog.py
yohanboniface: > Keeping them as is has the merit of making for a much simpler diff to review. :stuck_out_tongue_winking_eye: 

Meeeh! ;)

> What would separating them achieve?

Something useless being removed? :)

ideascube/serveradmin/catalog.py
bochecha: > Maybe we can call that one self._root?

I think the distinction between "cache" and "storage" is very important to keep.

A few months down the road, we'll have completely forgotten which is which.

ideascube/serveradmin/catalog.py
bochecha: >> Keeping them as is has the merit of making for a much simpler diff to review. :stuck_out_tongue_winking_eye:

> Meeeh! ;)

I agree that's not a very strong argument. :wink:

>> What would separating them achieve?

> Something useless being removed? :)

I don't understand. Right now we have one attribute, you want to split it in two. That's not removing anything, that's adding something.

ideascube/serveradmin/catalog.py
yohanboniface: humm, then I'd vote for `self._storage_root`.
I think it's OK that `Catalog._storage_root` is CATALOG_STORAGE_ROOT, and not STORAGE_ROOT itself.
I think that one being named `base_dir` while the other is `root` does not really change this anyway.

ideascube/serveradmin/catalog.py
yohanboniface: > I don't understand. Right now we have one attribute, you want to split it in two. That's not removing anything, that's adding something.

That's removing a useless dict ;)

ideascube/serveradmin/catalog.py
bochecha: Feels like busywork to me, but that's nevertheless done in the latest push to this branch.

ideascube/serveradmin/catalog.py
bochecha: Ok... done.

yohanboniface: Thanks :)

ideascube/serveradmin/catalog.py
yohanboniface: I meant that only this comment worth it, the rest is noise and should be removed IMHO. :s

PR #383, opened 2016-07-12
Add missing migrations

bochecha:
I think this is due to commit e4d96eee8867d2ee89a9eb8053b730a1dd322af8 which added Spanish to the list of available languages.

Since that list is used in the models, that causes a migration.


mgautierfr:
I don't like the fact that you repeat yourself twice.

It would be ideal to use settings.LANGUAGE in the migration script, but as the settings evolve, the migration script may be broken.

But, maybe, at least use the same list of language in the 3 applications to avoid repetition.
(Saying that, I don't know where to put this common language list)

bochecha:
> I don't like the fact that you repeat yourself twice.

What do you mean?

mgautierfr:
There is 3 times the same list of languages.

Le 12 juillet 2016 17:19:53 GMT+02:00, Mathieu Bridon <notifications@github.com> a écrit :
>> I don't like the fact that you repeat yourself twice.
>
>What do you mean?
>
>---
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/ideascube/ideascube/pull/383#issuecomment-232081736

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

bochecha:
> There is 3 times the same list of languages.

Right, that's because the migrations are automatically generated, so Django uses the value of the setting, not the setting itself.

In addition, and as you mentioned it in your original comment, we can't use the setting anyway here, because then past migrations using future settings won't be correct any more.

The only different way I could see of doing this would be to not make this field have a set of possible values (with the `choices`) but do validation ourselves. That would also remove the need for any future migration when we add new languages to the list. However, it also makes us lose automatic validation by the database.

Not sure what's better.

In any case, the 0.10.3 release is broken because of these missing migrations.

Issue #354, opened 2016-05-29
Create media center package content from CSV file


barbayellow:
To import medias in the ideascube medias center, we use an ideascube command to import medias from a csv file. 
Since the media center will  - in a very short future (after the code review :) - be integrated with the catalog feature, it would be great to replace the old csv batch process by a new oneallowing us to create a catalog from file in a folder to integrate it in the media center.

PR #386, opened 2016-07-12
Add a django command to create a package from a csv.

mgautierfr:
The csv has the same format as the csv used by import_media.
The created package should be added to a repository to be installed later.

This fix the issue #354. This PR is base on pr #342 as the later one introduce MediaCenter package.


bochecha:
What? Why did I close this? o_O

PR #387, opened 2016-07-12
Drop support of creating zip backup.

bochecha:
Zip format is generally a pain to deal with (hello non-ascii characters).

In addition, we're about to add filtering of which files to backup and which files to ignore, which is trivial to do with the tarfile module, but much harder with the zipfile one.

To keep things simpler, we've decided to remove Zip support.

Zip backups are still supported for restore, though.

[CONFIG_CHANGE]
- zip is no more supported for BACKUP_FORMAT.
- Default value for BACKUP_FORMAT as change from 'zip' to 'gztar'

This is part of #342

### DIFF
--- a/ideascube/serveradmin/backup.py
+++ b/ideascube/serveradmin/backup.py
@@ -11,9 +11,8 @@ from django.utils.translation import ugettext as _
 from ideascube import __version__
 
 
-def make_name(format=None):
+def make_name(format):
     """Return backup formatted file name."""
-    format = format or Backup.FORMAT
     basename = '-'.join([
         settings.IDEASCUBE_ID,
         __version__,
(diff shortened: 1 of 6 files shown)


ideascube/serveradmin/backup.py
```diff
@@ -25,6 +24,7 @@ def make_name(format=None):
 class Backup(object):
 
     FORMAT = settings.BACKUP_FORMAT
+    SUPPORTED_FORMATS_AT_CREATION = ('bztar', 'gztar', 'tar')
```
yohanboniface: Rereviewing this, I'm thinking we could also drop zip restore support: very very fiew backups have been made (I think only by my or so), I think no body is going to try to restore them (it does a backup before restoring, which backup takes ages); and if we really need to restore a zip file, it's at the end only an unzip and put the directory in the right place, which we can do by hand for once.

bochecha: Since the code is there, it causes no difficulty to keep it, and there might be existing backups we don't know about, I'd say no, let's keep it, at least for some time, to provide a safer transition path.

How about we remove it in a month or two, when we can feel confident that people either have made new tar backups, or they just don't do backups at all?

yohanboniface: > How about we remove it in a month or two, when we can feel confident that people either have made new tar backups, or they just don't do backups at all?

In a month, nobody but the new deployed boxes will have the new code, but OK ;)

PR #390, opened 2016-07-12
One db for you, and one db for you too

bochecha:
This is part of #342

### DIFF
--- a/db_router.py
+++ b/db_router.py
@@ -1,0 +1,30 @@
+class DatabaseRouter(object):
+    """
+    A Database router.
+    Some models do not need (and must not be backuped).
+    Those models will (must) have a _dbname class attribute set to 'transiant'.
+    This attribute will be use as database name.
+    """
+    def db_for_read(self, model, **hints):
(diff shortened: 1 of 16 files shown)


db_router.py
```diff
@@ -0,0 +1,30 @@
...
+        """
+        Allow relations if both object are in the same database.
+        """
+        return getattr(obj1, '_dbname', None) == getattr(obj2, '_dbname', None)
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        """
+        All database need to be migrated.
+        """
+        return True
```
bochecha: As discussed in the other PR, I strongly think we should return `None` here, to let Django decide with its own rules whether the migration should be applied or not.

I don't think our database router should be in the business of forcing (or forbidding) migrations, its job is to decide which database to connect to depending on the model.

bochecha: We agreed on IRC to return `None`.

yohanboniface:
Already merged, but Github bug.

PR #391, opened 2016-07-12
Packaged content for the media center

bochecha:
This is part of #342


db_router.py
```diff
@@ -0,0 +1,30 @@
...
+        """
+        Allow relations if both object are in the same database.
+        """
+        return getattr(obj1, '_dbname', None) == getattr(obj2, '_dbname', None)
+
+    def allow_migrate(self, db, app_label, model_name=None, **hints):
+        """
+        All database need to be migrated.
+        """
+        return True
```
bochecha: As discussed in the other PR, I strongly think we should return `None` here, to let Django decide with its own rules whether the migration should be applied or not.

I don't think our database router should be in the business of forcing (or forbidding) migrations, its job is to decide which database to connect to depending on the model.

ideascube/mediacenter/forms.py
```diff
@@ -24,9 +27,38 @@ def save(self, commit=True):
...
+        if commit:
+            if not document.id:
+                # m2m need the document to have an id.
+                document.save()
+            self.save_m2m()
+            document.save()  # Search index needs m2m to be saved.
+        return document
+
+
+class NotFileDocumentForm(forms.ModelForm):
```
yohanboniface: Not sure about the naming.
`DocumentWithoutFileForm`?

ideascube/mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -19,10 +19,19 @@
...
     {% if user.is_staff %}
-        <ul class="card tinted admin">
-            <li>{% fa 'pencil' 'fa-fw' %}<a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
-            <li>{% fa 'trash' 'fa-fw' %}<a href="{% url 'mediacenter:document_delete' pk=document.pk %}">{% trans 'Delete' %}</a></li>
-        </ul>
+        {% if document.package_id %}
+            <span class="card admin">
+               {% blocktrans %}
+                   Cannot edit or delete this document.
+                   It is installed by the {{ document.package_id }} package.
```
yohanboniface: Add `"` around the package id?

ideascube/mediacenter/templates/mediacenter/document_detail.html
```diff
@@ -19,10 +19,19 @@
 {% endblock twothird %}
 {% block third %}
     {% if user.is_staff %}
-        <ul class="card tinted admin">
-            <li>{% fa 'pencil' 'fa-fw' %}<a href="{% url 'mediacenter:document_update' pk=document.pk %}">{% trans 'Edit' %}</a></li>
-            <li>{% fa 'trash' 'fa-fw' %}<a href="{% url 'mediacenter:document_delete' pk=document.pk %}">{% trans 'Delete' %}</a></li>
-        </ul>
+        {% if document.package_id %}
+            <span class="card admin">
```
yohanboniface: I think all the cards in the right column needs `tinted` class.

ideascube/serveradmin/catalog.py
```diff
@@ -306,6 +317,88 @@ def remove(self, install_dir):
             printerr(e)
 
 
+class StaticSite(SimpleZipPackage):
+    typename = 'static-site'
+    handler = Nginx
+
+
+class ZippedMedia(SimpleZipPackage):
```
yohanboniface: `ZippedMedias` and `zipped-medias`?

ideascube/serveradmin/catalog.py
```diff
@@ -306,6 +317,88 @@ def remove(self, install_dir):
...
+    def remove(self, install_dir):
+        # Easy part here. Just delete documents from the package.
+        Document.objects.filter(package_id=self.id).delete()
+        super().remove(install_dir)
+
+    def install(self, download_path, install_dir):
+        super().install(download_path, install_dir)
+        print('Adding medias to mediacenter database.')
+        root = self.get_root_dir(install_dir)
+        manifestfile = Path(root, 'manifest.yml')
```
yohanboniface: Catch missing `manifest.yml` ?

bochecha:
In his commit message, @mgautierfr says this:

> The link is just hidden. This is still possible to change the document metadatas if a user directly knows and uses the edit form.
>
> This probably should be totally removed in the futur, but for now, I think it is better to keep it.

It certainly needs to be removed, indeed. Shall I create a ticket so we don't forget to do it? (maybe in the next sprint)

yohanboniface:
> It certainly needs to be removed, indeed. Shall I create a ticket so we don't forget to do it? 

Yes

> (maybe in the next sprint)

Yes

ideascube/mediacenter/forms.py
bochecha: Except that document does have a file.

It's just that the file is not uploaded through normal means, it already is in the right place and we just pass the file **path** to the form.

`PreuploadedDocumentForm`?

ideascube/serveradmin/catalog.py
bochecha: Nope, "media" is already the plural form.

https://en.wiktionary.org/wiki/media#Etymology_2

bochecha:
>> It certainly needs to be removed, indeed. Shall I create a ticket so we don't forget to do it?

> Yes

https://github.com/ideascube/ideascube/issues/396

ideascube/serveradmin/catalog.py
yohanboniface: I've been heard we are not speeking latin anymore since some years now…

ideascube/mediacenter/forms.py
bochecha: Renamed to PackagedDocumentForm in the latest push to this branch.

ideascube/mediacenter/templates/mediacenter/document_detail.html
bochecha: Done in the latest push to this branch.

ideascube/mediacenter/templates/mediacenter/document_detail.html
bochecha: Done in the latest push to this branch.

ideascube/serveradmin/catalog.py
bochecha: Done in the latest push to this branch.

ideascube/serveradmin/catalog.py
bochecha: Done in the latest push to this branch.

:mask:

PR #397, opened 2016-07-13
Add a django command to create a package from a csv

bochecha:
The csv has the same format as the csv used by import_media.
The created package should be added to a repository to be installed later.

This fix the issue #354.

Apologies to @mgautierfr for closing his original #386 . I have no clue what happened.


yohanboniface:
I'm tempted to say that in a first iteration we should stay in `serveradmin` module, and move to something else when we are a bit more mature.

ideascube/catalog_utils/package.py
```diff
@@ -0,0 +1,55 @@
...
+
+import yaml
+import os.path as op
+import zipfile
+import hashlib
+import codecs
+import csv
+
+
+class MediaCenterPackage:
```
bochecha: So we have a class somewhere that represents media content in a zip that we can install on the system and import in the media center.

This here is a class that lets us zip media content into a package from a CSV description.

The two seem awfully related. Should they be merged?

yohanboniface:
We need to also add preview files when they are defined.

yohanboniface:
Use `default_flow_style=False` when dumping yaml, so we have a more readable file for debugging.

ideascube/catalog_utils/package.py
bochecha: We agreed not to merge these, at least for now, since this code is probably going to evolve quite a bit.

ideascube/serveradmin/package.py
```diff
@@ -22,6 +22,12 @@ def check_media(self, media):
                 return False
         if not op.exists(op.join(self.working_dir, media['path'])):
             return False
+        preview_path = media.get('preview')
+        if (preview_path
+         and not op.exists(op.join(self.working_dir, preview_path))):
+            print("Warning : preview_path {} defined but file do not exists"
```
bochecha: s/do not/does not/

ideascube/serveradmin/package.py
```diff
@@ -0,0 +1,66 @@
...
+import hashlib
+import codecs
+import csv
+
+
+class MediaCenterPackage:
+    def __init__(self, working_dir, medias=None):
+        self.working_dir = working_dir
+        self.medias = medias or []
+        self.medias = list(filter(self.check_media, medias or []))
```
bochecha: Why twice the `medias or []` ?

ideascube/serveradmin/tests/data/a-video.mp4
```diff
@@ -0,0 +1 @@
+aaaaaaaaaaaaaaaaaaaaaaa
```
bochecha: I've seen this movie, loved it!

bochecha:
Except for these 2 trivial details, this is mergeable as far as I'm concerned.

ideascube/serveradmin/package.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def check_media(self, media):
+        for require_metadata in ('title', 'summary', 'credits', 'path'):
+            if require_metadata not in media:
+                return False
+            if not media[require_metadata]:
```
yohanboniface: `if not media.get(required_medata)`?

ideascube/serveradmin/package.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class MediaCenterPackage:
+    def __init__(self, working_dir, medias=None):
+        self.working_dir = working_dir
+        self.medias = medias or []
+        self.medias = list(filter(self.check_media, medias or []))
+
+    def check_media(self, media):
+        for require_metadata in ('title', 'summary', 'credits', 'path'):
```
yohanboniface: requireD?

ideascube/serveradmin/package.py
```diff
@@ -0,0 +1,66 @@
...
+            if require_metadata not in media:
+                return False
+            if not media[require_metadata]:
+                return False
+        if not op.exists(op.join(self.working_dir, media['path'])):
+            return False
+        preview_path = media.get('preview')
+        if (preview_path
+         and not op.exists(op.join(self.working_dir, preview_path))):
+            print("Warning : preview_path {} defined but file do not exists"
```
yohanboniface: does not exist

ideascube/serveradmin/tests/data/a-video.mp4
yohanboniface: Commenting about the images: don't we already have them for the tests somewhere?

ideascube/serveradmin/tests/test_csv2package_command.py
```diff
@@ -0,0 +1,37 @@
...
+    csv_path = csv_writer(metadata)
+    call_command('csv2package', csv_path, package_path)
+    with zipfile.ZipFile(package_path) as package:
+        assert set(package.namelist()) == set(['manifest.yml',
+                                               'a-video.mp4',
+                                               'a-pdf.pdf',
+                                               'an-image.jpg'])
+        with package.open('manifest.yml') as m:
+            manifest = yaml.load(m)
+        assert manifest == {'medias': [{'title': 'my video',
```
yohanboniface: For readability, I'd return after the first `[`

ideascube/serveradmin/package.py
mgautierfr: Why twice self.media = … ?
Only one is useful, you never seen the first line. Look the red light, you never seen this line. :)

ideascube/serveradmin/package.py
mgautierfr: Definitively.

ideascube/serveradmin/management/commands/csv2package.py
```diff
@@ -0,0 +1,33 @@
+
```
yohanboniface: tempted to name the command `csv2pkg` to gain some characters to type, and `pkg` is still readable imho :)

ideascube/serveradmin/package.py
```diff
@@ -3,29 +3,25 @@
...
+        for required_metadata in ('title', 'summary', 'credits', 'path'):
+            if not media.get(required_metadata):
                 return False
         if not op.exists(op.join(self.working_dir, media['path'])):
             return False
         preview_path = media.get('preview')
         if (preview_path
          and not op.exists(op.join(self.working_dir, preview_path))):
-            print("Warning : preview_path {} defined but file do not exists"
+            print("Warning : preview_path {} defined but file does not exists"
```
yohanboniface: does not exisT :p

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+import sys
+
+from ideascube.serveradmin.package import MediaCenterPackage
+
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = ('Batch import medias from CSV metadata. CSV file must contain '
```
yohanboniface: help needs to be adapted.

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -0,0 +1,36 @@
...
+    def abort(self, msg):
+        self.stderr.write(msg)
+        sys.exit(1)
+
+    def handle(self, *args, **options):
+        csv_path = os.path.abspath(options['csv_path'])
+        package_path = os.path.abspath(options['package_path'])
+
+        if not os.path.exists(csv_path):
+            self.abort('Path does not exist: {}'.format(csv_path))
```
yohanboniface: I wonder if when in dry-run mode we should abort, or just print the error and continue.

yohanboniface: Erf, forget, I was reading media_path instead of csv_path…

ideascube/serveradmin/package.py
```diff
@@ -0,0 +1,57 @@
...
+            if not media.get(required_metadata):
+                return False
+        if not op.exists(op.join(self.working_dir, media['path'])):
+            return False
+        preview_path = media.get('preview')
+        if (preview_path
+         and not op.exists(op.join(self.working_dir, preview_path))):
+            print("Warning : preview_path {} defined but file does not exists"
+                  .format(preview_path))
+            media['preview'] = None
```
yohanboniface: I'm not sure `check_media` should alter the metadatas. I'd move the print when actually using the preview path (when creating the zip).

mgautierfr: Hum, you will get no warning if you use --dry-run then.

Checking that preview path is valid or not is part of the check. It should not be in the zip creation.
(And do we need to keep an optional metadata if we know it is invalid ?)

ideascube/serveradmin/tests/data/a-video.mp4
mgautierfr: The image/video/pdf come from tests for mediacenter import_medias command.
They are indeed duplicated but import_medias will be removed in the future (maybe now ?)

yohanboniface: Good point.
Let's keep the command for now.

ideascube/serveradmin/package.py
yohanboniface: ok

PR #398, opened 2016-07-18
Configuration draft for Bardarash, Kurdistan.

yohanboniface:


ideascube/conf/idb_kur_bardarash.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+"""Bardarash in Kurdistan"""
+from .azraq import *  # noqa
+from django.utils.translation import ugettext_lazy as _
+
+IDEASCUBE_NAME = u"Bardarash"
+USER_FORM_FIELDS = (
```
yohanboniface: I think we can keep Azraq ones until we have better knowledge of wanted fields.

ideascube/conf/idb_kur_bardarash.py
```diff
@@ -0,0 +1,36 @@
...
+    {
+        'id': 'khanacademy',
+    },
+    {
+        'id': 'wikipedia',
+        'languages': ['en','ar']
+    },
+    {
+        'id': 'gutenberg',
+        'languages': ['en','ar']
```
yohanboniface: `lang: mul`

barbayellow:
I didn't even know we could add or remove fields in the user handler app. Can we add any kind of field (image, blob, etc.) ? Any number of fields ? Ore are there any restrictions ?

yohanboniface:
> Can we add any kind of field (image, blob, etc.) ? Any number of fields ? Ore are there any restrictions ?

No strong restriction. The trick is that every field is added for every box, but then each box decides to show only some of them.

Issue #374, opened 2016-07-07
Use Site name for Ideasbox name


yohanboniface:
And add a form in the admin for changing the site name.

PR #403, opened 2016-07-21
Allow setting the server name in the admin

bochecha:
Implements #374


ideascube/serveradmin/templates/serveradmin/index.html
```diff
@@ -16,11 +16,12 @@
         </ul>
         <hr />
         <ul>
-            <li>{% fa 'cogs' 'fa-fw' %} <a href="{% url 'server:services' %}">{% trans "Manage services" %}</a></li>
-            <li>{% fa 'power-off' 'fa-fw' %} <a href="{% url 'server:power' %}">{% trans "Restart server" %}</a></li>
-            <li>{% fa 'life-ring' 'fa-fw' %} <a href="{% url 'server:backup' %}">{% trans "Manage backups" %}</a></li>
-            <li>{% fa 'plug' 'fa-fw' %} <a href="{% url 'server:battery' %}">{% trans "Monitor battery" %}</a></li>
-            <li>{% fa 'wifi' 'fa-fw' %} <a href="{% url 'server:wifi' %}">{% trans "Manage Wi-Fi" %}</a></li>
+            <li>{% fa 'info-circle' 'fa-fw' %} <a href="{% url 'server:name' %}">{% trans "Server name" %}</a></li>
+            <li>{% fa 'cogs' 'fa-fw' %} <a href="{% url 'server:services' %}">{% trans "Services" %}</a></li>
```
yohanboniface: Changing labels means that translations need to be updated. We need to be sure we want those change then.

ideascube/utils.py
```diff
@@ -9,6 +13,19 @@ def __get__(self, cls, owner):
         return self.fget.__get__(None, owner)()
 
 
+def get_server_name():
```
yohanboniface: cache me?

yohanboniface: Given it's in the header, and thus requested on every page.

ideascube/serveradmin/tests/test_models.py
```diff
@@ -0,0 +1,45 @@
...
+        'string',
+    ])
+@pytest.mark.usefixtures('db')
+def test_settings(value, type_method, user):
+    from ideascube.serveradmin.models import Setting
+
+    fakenow = datetime.now(tz=timezone.utc)
+    assert Setting.objects.count() == 0
+
+    with freezegun.freeze_time(fakenow):
```
yohanboniface: I think you can simply use `Setting.objects.update(date=mydate)` to avoid this new dep.

ideascube/serveradmin/tests/test_models.py
```diff
@@ -0,0 +1,45 @@
...
+    setting = Setting.objects.first()
+    assert setting.namespace == 'tests'
+    assert setting.key == 'setting1'
+    assert setting.value == value
+    assert setting.actor == user
+    assert setting.date == fakenow
+    assert str(setting) == 'tests.setting1=%r' % value
+
+    if type_method is not None:
+        for method in filter(lambda m: m.startswith('as_'), dir(setting)):
```
yohanboniface: Why not using `hasattr`?

ideascube/serveradmin/models.py
```diff
@@ -0,0 +1,27 @@
...
+    class Meta:
+        ordering = ['-date']
+
+    namespace = models.CharField(max_length=40)
+    key = models.CharField(max_length=40)
+    value = JSONField()
+    actor = models.ForeignKey(settings.AUTH_USER_MODEL)
+    date = models.DateTimeField(auto_now=True)
+
+    def as_boolean(self):
```
yohanboniface: What about adding a `default` to those methods signature?

    # We'd use the default is the value is not on the DB
    # or if casting to boolean fails.
    setting.get_as_boolean(default=False)

ideascube/serveradmin/tests/test_models.py
bochecha: I don't see how that would help here?

yohanboniface: prevent looping?

ideascube/serveradmin/models.py
bochecha: So at first here, I had those methods that weren't casting.

Instead, they were ensuring that the value was of the given type, and raising `TypeError` if not.

That is, I had:

```python
    def as_boolean(self):
        if not isinstance(self, bool):
            raise TypeError('%s is not a boolean' % self)

        return self.value

    def as_integer(self):
        if not isinstance(self, int):
            raise TypeError('%s is not an integer' % self)

        return self.value

    def as_string(self):
        if not isinstance(self, str):
            raise TypeError('%s is not a string' % self)

        return self.value
```

I found it more interesting, because it served as validation of the setting values, and I was happy. The `default` parameters you suggest would have made **lots** of sense here.

Then went on to implement the tests, and was surprised when they didn't do what I expect.

I was expecting that a string setting to raise for all the `as_*` methods but `as_string`, an integer setting to raise for all the `as_*` method but `as_integer`, etc...

But what happened is that a boolean setting raised only for `as_string`, it successfully passed through for `as_boolean` (of course) and `as_integer`. Because of course, `bool` is a subclass of `int`!

That made me realize that the validation provided by these methods wasn't as good as I initially thought.

Then I started questioning whether it even was useful in the first place.

Then I decided that instead of validation, I'd just cast, which is another way of ensuring I'm getting a value in the right type (although it might be a completely bogus value).

But to be honest, I'm now less convinced that these methods are useful at all.

And I don't think the `default` you suggest makes sense in this case, either.

---

Fast forward to today, I've slept on it, and I'm glad you didn't merge it. :smile: 

This morning I'm even less convinced by those `as_*` methods than I was yesterday.

I mean, they ensure that if I want a string setting I'm going to get a string setting, sure.

But I still have to do something like:

```python
try:
    setting = Setting.objects.get(namespace='foo', key='bar')

except Setting.DoesNotExist:
    value = default_value

else:
    try:
        value = setting.as_<something>()

    except TypeError:
        value = default_value
```

That's a pretty horrendous API!

With the `default` you suggest, it gets better:

```python
try:
    setting = Setting.objects.get(namespace='foo', key='bar')

except Setting.DoesNotExist:
    value = default_value

else:
    value = setting.as_<something>(default=default_value)
```

But it's still pretty bad API. :smiley: 

This would be a much better API:

```python
value = Setting.get_<something>('foo', 'bar', default=default_value)
```

That would mean making those current `as_*` methods into `get_*` class methods, implemented something like this:

```python
    @classmethod
    def get_string(cls, namespace, key, default_value=None):
        try:
            setting = Setting.objects.get(namespace=namespace, key=key)

        except Setting.DoesNotExist:
            if default_value is not None:
                return default_value

            raise

        else:
            if insinstance(setting.value, str):
                return setting.value

            if default_value is not None:
                return default_value

            raise TypeError('%s is not a string setting' % setting)
```

Yes, I went back to checking the type instead of casting. I think this is better, and the problem I had with boolean settings is... well I don't really think it's a problem after a good night's sleep.

Of course, rather than class methods, maybe manager methods would be more idiomatic Django, even though they'd make for worse API:

```python
value = Setting.objects.get_<something>('foo', 'bar', default=default_value)
```

(and for the record, this is what I had actually started with yesterday, but then changed my mind)

What do you think?

ideascube/utils.py
bochecha: How would you cache it?

ideascube/serveradmin/tests/test_models.py
bochecha: I don't see it as a problem to have this new dep.

In fact, I see it as a basic need in order to write unit tests properly, just like pytest.

And in fact, I was thinking of slowly going back over all unit tests, and making them use freezegun where appropriate, one by one. (but not in this branch, of course)

What you suggest is a different test: it allows ensuring that if we explicitly set a date for settings, then it is stored and retrieved properly.

freezegun allows testing something else: it allows ensuring that the date is correctly set by `auto_now`.

ideascube/serveradmin/templates/serveradmin/index.html
bochecha: Ok, dropped.

ideascube/serveradmin/tests/test_models.py
bochecha: Huh? I just don't get what you mean, can you make an example?

ideascube/serveradmin/models.py
yohanboniface: I agree on class methods over manager, and on the API you suggest, raising only when no default is given.
My only concern is that we *may* need to have `None` as default value.
I don't like it that much, but an alternative option would be to have `**kwargs` and to check for `default` being actually passed.

yohanboniface: I agree on class methods over manager, and on the API you suggest, raising only when no default is given.
My only concern is that we *may* need to have `None` as default value.
I don't like it that much, but an alternative option would be to have `**kwargs` and to check for `default` being actually passed.

ideascube/serveradmin/tests/test_models.py
yohanboniface: OK

ideascube/serveradmin/tests/test_models.py
yohanboniface:
```
if hasattr(setting, type_method):
    assert getattr(setting, method)() == value
```
Sounds so simple that I'm certainly missing something (maybe hours of sleep? :p )

bochecha: Well yeah, all settings have all type methods.

So to avoid looping, the solution is:

```python
assert getattr(setting, type_method)() == value
```

The looping is indeed entirely unnecessary here, I had it because at first I was testing that calling the corresponding `type_method` would succeed **and** calling the other methods would raise (see the other discussion about how I went from checking `isinstance` to casting).

But when the methods stopped raising, I removed the calls to the other `as_methods`, and left that loop, without realizing it was entirely pointless.

But in any case, this will completely change as I redo the API.

ideascube/serveradmin/models.py
bochecha: I really wouldn't like using `**kwargs` here.

`**kwargs` generally implies something like « pass in what you want, I'll make sense of it »

Here, we expect a single optional argument, `default`.

If `None` as a default value ever becomes a concern, then we could just introduce a new « empty » value, and check for that instead of None.

But I feel that's a concern better kept for later, if that ever happens.

ideascube/utils.py
bochecha: We agreed to leave it for now, and work on caching if performance ever become a problem.

bochecha:
@yohanboniface your comments should now be addressed in the latest push to this branch, including a redesigned API for the settings system.

ideascube/serveradmin/models.py
```diff
@@ -0,0 +1,55 @@
...
+            raise TypeError(
+                '%s is not a %s setting' % (setting, type.__name__))
+
+    @classmethod
+    def get_string(cls, namespace, key, default=None):
+        return cls._get_typed(namespace, key, str, default=default)
+
+    @classmethod
+    def set(cls, namespace, key, value, actor):
+        try:
```
yohanboniface: Why not using https://docs.djangoproject.com/en/1.9/ref/models/querysets/#update-or-create ?

ideascube/serveradmin/templates/serveradmin/name.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans "Change Server Name" %}</h2>
```
yohanboniface: Not sure about the uppercased Server and Name. I'd keep them lowercase.

ideascube/serveradmin/models.py
```diff
@@ -0,0 +1,55 @@
...
+    actor = models.ForeignKey(settings.AUTH_USER_MODEL)
+    date = models.DateTimeField(auto_now=True)
+
+    @classmethod
+    def _get_typed(cls, namespace, key, type, default=None):
+        try:
+            setting = Setting.objects.get(namespace=namespace, key=key)
+
+        except Setting.DoesNotExist:
+            if default is not None:
```
mgautierfr: None is not a valid default value ?
Maybe use a kwargs and look if default is in it ?

ideascube/serveradmin/models.py
```diff
@@ -0,0 +1,55 @@
...
+                return default
+
+            raise
+
+        else:
+            if isinstance(setting.value, type):
+                return setting.value
+
+            if default is not None:
```
mgautierfr: Do we really want to return the default value if the value in the db is not the intended type ?
For me, it sound like a error, whatever a default argument is provided or not.

yohanboniface: At the time the error will raise, there will be no developer to be able to fix the situation.
We need to be defensive as much as possible.

ideascube/serveradmin/templates/serveradmin/name.html
bochecha: No: http://www.grammar-monster.com/lessons/capital_letters_title_case.htm

I understand why you would think that though, since typographic rules for title case in French are different (and somewhat incredible) : http://typographisme.net/post/Une-question-d%C3%A9licate-%3A-les-capitales-dans-les-titres

ideascube/serveradmin/models.py
bochecha: Because I always forget this method exists, of course!

Thanks, fixed in the latest push to this branch.

ideascube/serveradmin/models.py
bochecha: https://github.com/ideascube/ideascube/pull/403#discussion_r71860671

ideascube/serveradmin/models.py
bochecha: > For me, it sound like a error, whatever a default argument is provided or not.

I actually agree with you.

I had very strong doubts when I wrote this code this morning.

But then I thought that...

> At the time the error will raise, there will be no developer to be able to fix the situation.

So the question is, what else could we do here?

mgautierfr: > At the time the error will raise, there will be no developer to be able to fix the situation.

There are a lot of situation when no developer is around to fix a bug :)

How the values will be set ? Is it the user (ansiblecube) directly setting the value in the db or we use API that will put a value of the right type in the db ?

If the user set directly in the db, then the db is like a config file and we need a defensive approach.

Else, there is no reason* to be more defensive than elsewhere. We (developer) write something in the db, and we can't read it back. This should not happen.

\* It may be one reason : evolution of the settings. But I think that it should be to the using code to be prepared to and to handle this case.

However, trying to have a defensive approach is still a valid argument,.I will not block this pr for this. (If, at least, there is a warning message)

ideascube/serveradmin/models.py
mgautierfr: Maybe use … ?
This is unhabitual but there is less chance that … is a valid default value than None.

(Just a idea, not sure it is a good one :) )

ideascube/serveradmin/models.py
yohanboniface: >  It may be one reason : evolution of the settings.

That is :)
We could also catch the error every where we use call for in-db setting, but for what benefit? Raising a warning? What the user will do with the warning?
And still, if in some place we do want to catch the error externally, we just do not provide a default.

bochecha: > We could also catch the error every where we use call for in-db setting, but for what benefit? 

This, to me, is the real argument.

There is nothing we could do with that error that would be useful to the user.

At least returning the default value gives them a degraded experience (default value instead of what they expected), which is infinitely better than no experience at all, being blocked on an error message they can't, and won't understand.

mgautierfr: Ok with that. (we need somehow a way to stop discuss during hours of somethings "not so important" or "I would agree with that")

The warning is not necessarily for the user, but we need a way to know that we used the default value cause the value in the db is wrong. A message in the log is ok.

ideascube/models.py
```diff
@@ -279,3 +285,71 @@ class SortedTaggableManager(_TaggableManager):
...
+                '%s is not a %s setting' % (setting, type.__name__))
+
+    @classmethod
+    def get_string(cls, namespace, key, default=...):
+        return cls._get_typed(namespace, key, str, default=default)
+
+    @classmethod
+    def set(cls, namespace, key, value, actor):
+        setting, _ = Setting.objects.update_or_create(
+            namespace=namespace, key=key, actor=actor,
```
yohanboniface: I think `actor` should be in defaults, no?

bochecha: You're obviously right, fixed.

PR #407, opened 2016-07-22
Use the document's kind instead of 'Document' for the Download buton

### LINKED ISSUE #399: Button label
In French (I guess this is the same in English), when viewing a video, the label of the button is "Download the document". It should be "Download the video".

mgautierfr:
Fix #399.


bochecha:
s/buton/button/ for the commit message

Otherwise I haven't tested it, but it looks obviously correct.

bochecha:
And now I tested it, and it works as advertised.

mgautierfr:
Commit message fixed.

PR #409, opened 2016-07-23
db_router should be inside ideascube package

yohanboniface:


bochecha:
Doh! :+1:

yohanboniface:
@mgautierfr the main info is this one:

    db_router.py → ideascube/db_router.py

Then I took the opportunity to move the setting into base.py just because settings should only be used as a loader that deals with local configuration. Everything else is better is base.py which is more readable.

mgautierfr:
>@mgautierfr the main info is this one:

>> db_router.py → ideascube/db_router.py

I would say it doesn't explain why, but I took a look on the code base, and there is almost no module at root dir, so yes move it.

It seems to me that if a configuration is a django configuration made by ideascube and not an ideascube configuration itself, it should not be in base.py (as everything in base.py can be overwritten by local config)
This seems to be the case for DATABASE_ROUTERS (and btw, also for INSTALLED_APPS, MIDDLEWARE_CLASSES, TEMPLATES, ... It a more global point that just this pr).

DATABASE_ROUTERS may be changed by a admin if he want to deactivate multi-db, but :
- I'm not sure it is supported
- It should be named MULTIDB_SUPPORT, SEPARATE_TRANSIENT or something like that.

bochecha:
> I would say it doesn't explain why, but I took a look on the code base, and there is almost no module at root dir, so yes move it.

This is about Python packaging.

If you put a module outside of the python package, then distutils/setuptools will ignore it.

It won't find its way into the tarballs, it also won't get installed.

We didn't have the problem because we all ran the code in development mode (`pip install -e`).

I agree that this should have been explained in the commit message, and I regret not having asked for it when I merged this. :disappointed:

Issue #408, opened 2016-07-23
Build failling with new Pillow


yohanboniface:
https://paste.gnome.org/pglmszxiy

Workaround: downgrade it
Solution: find a way not to use wheels for compiling, should be passing `--no-binary` to pip.

bochecha:
The proper solution will be to pass `--no-binary` to pip, as you said.

This requires dh-virtualenv >= 0.8, which introduced the `--extra-pip-args` option, allowing us to do that.

Unfortunately, Jessie only has dh-virtualenv 0.7.

Fortunately, Jessie backports contain dh-virtualenv 0.10.

I'm [trying to figure out a way to have our build system depend on packages from the backports archive](http://unix.stackexchange.com/questions/298001/building-with-sbuild-and-backports), without success so far.

PR #410, opened 2016-07-23
Downgrade Pillow (cf #408)

yohanboniface:


bochecha:
This sucks, but yeah, we have to for now. :+1:

PR #411, opened 2016-07-24
Fix some rtl alignments

yohanboniface:
Before the commits:
selection_011
selection_009

After the commit:
selection_010
selection_008

During the commit (waiting for the house keys under 53° Celsius (yeah I mean Celsius) ;)
img_0885 resized

### DIFF
--- a/ideascube/static/ideascube/main.css
+++ b/ideascube/static/ideascube/main.css
@@ -331,6 +331,10 @@ input[type="password"].showable-password {
     margin-left: -25px;
     cursor: pointer;
 }
+html[dir='rtl'] .show-password {
+    margin-left: auto;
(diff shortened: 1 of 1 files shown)


ideascube/static/ideascube/main.css
```diff
@@ -1,4 +1,4 @@
-/* ************************************************* */
+611/* ************************************************* */
```
bochecha: 611?

yohanboniface: Huhu, seems the brightness is not enough for those lights :p
Good catch :)

bochecha:
CSS change looks correct. Not tested, but I trust you on those screenshots.

PR #413, opened 2016-07-26
Bardarash config

yohanboniface:
Changes needed for Bardarash deployment.


ideascube/migrations/0006_auto_20160726_1342.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('ideascube', '0005_setting'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='country_of_origin_occupation',
+            field=models.CharField(blank=True, max_length=100, verbose_name='Occupation before reaching the camp'),
```
bochecha: Commit message is specific to your current deployment, but this applies to all deployments.

Is that going to cause problems for others?

ideascube/models.py
```diff
@@ -239,7 +239,8 @@ def index_strings(self):
         _('Current occupation'), choices=OCCUPATION_CHOICES, blank=True,
         max_length=32)
     country_of_origin_occupation = models.CharField(
-        _('Occupation in the country of origin'),
+        _('Occupation before reaching {place}').format(
+                                        place=settings.IDEASCUBE_PLACE_NAME),
```
bochecha: So the verbose name is depending on a setting, but it's set by the migration to... the value from your setting at the time you generated the migration.

Is that really how it should work?

ideascube/conf/idb_irq_bardarash.py
```diff
@@ -1,9 +1,17 @@
-# -*- coding: utf-8 -*-
-"""Bardarash in Kurdistan"""
+"""Bardarash in Kurdistan, Iraq"""
 from .azraq import *  # noqa
 from django.utils.translation import ugettext_lazy as _
 
-IDEASCUBE_NAME = u"Bardarash"
```
bochecha: Trivial nit: the commit message could have reflected this change. :stuck_out_tongue_winking_eye:

ideascube/migrations/0006_auto_20160726_1342.py
yohanboniface: It will affect every other box yes, but the wording should now be generic. So this should not be an issue.

ideascube/models.py
yohanboniface: No, good catch!

PR #424, opened 2016-07-27
Save session at every request (fix #417)

### LINKED ISSUE #417: Only disconnect staff after one hour of inactivity
Now it seems the disconnection is one hour after the login, no matter if the session was active or not.
So what happen is when creating users one after the other, the disconnection appear when hitting save, and then all the form is lost (hitting back button will end with an annoying CSRF error).

So disconnection should not be while the session is active.

yohanboniface:
So that an active session is not deleted.


bochecha:
Tried this out, it seems to work as expected.

PR #427, opened 2016-07-28
Redo language handling in forms

### LINKED ISSUE #425: Allow to create content with any language from LANG_INFO instead of AVAILABLE_LANGUAGES
Now we are only allowing to create content (blog, document, …) with a language defined in the AVAILABLE_LANGUAGES setting.
It's not scalable because we are not adding all languages to this settings.
Instead we should allow any languages from the ones in `django.conf.locale.LANG_INFO`.
Thought, we should not display all those languages in the select of the form, or if we do so we should have a way to either put some languages on the top (defined from a per box setting) of the list or easily search in the list.
In the case where we only display some languages in the select, we should not override the language when editing a document which as another language.
We should then allow languages code with more than 2 letters in the db (cf #356).

AVAILABLE_LANGUAGES should then be used only for the available translations of the ideascube front (so the dropdown select in the footer).

bochecha:
Fixes #425


bochecha:
One thing missing here, which @yohanboniface pointed out, is that we need to set `LOCAL_LANGUAGES` to correct values for the configuration files of the already deployed boxes.

bochecha:
So we've decided to go with a setting in database for this, instead of the `LOCAL_LANGUAGES` in a file.

What that means is:

* a migration to set this to the list of languages for which content has already been uploaded, and if none, then a reasonable default
* an admin page to select the local languages

yohanboniface:
Quickly backlogged the discussion on IRC, a very quick answer here as you're not online there and I'll only be back online myself tomorrow around 14:00 when I'm back to Paris:
- I think we can do without the migration, I suggest not spending time on this (mainly because as for today we update the box by hand so we can also set this setting; and migrations like cannot be really unittested, so there is a risk we'll spend time on writing it and time on debugging the edge cases when we run it for real in the boxes)
- note on the overall strategy: we do not want to suppress the per-box settings files, we want them not to be mandatory

Now two hours of sleep before taking my plane to Amman, then to Paris, yeeaaah ;)

bochecha:
> I think we can do without the migration

So how to get the old boxes with the right setting for `LOCAL_LANGUAGES`?

Honestly, keeping it in the conf file seems **very** underwhelming. It can cause confusion in the long run, when admins continue modifying the conf file to no effect because the setting is in the database.

I just can't see a good reason to keep configuration in both places.

bochecha:
>> I think we can do without the migration

> So how to get the old boxes with the right setting for `LOCAL_LANGUAGES`?

Mulling this over, `LOCAL_LANGUAGES` is something I was introducing in this branch, so it's not something that needs to be migrated over for old boxes.

Plus, we had said that the default should be a small list, like maybe Arabic, English, French and Spanish.

So I've changed my branch, now it doesn't introduce a new setting in the configuration files, it just takes the `Setting` from the database and falls back on the list above.

So ignore my previous comment.

I've added a little twist, though, in that the default will be the list of languages for which content has been uploaded, if there is some. Thinking about it some more, this might be a bit intensive to query, so we might not want to do it when displaying the form?

However, I do think this is an important use case.

Imagine a box has uploaded some content in Kurdî, before we merge this pull request. Then, we update their box with a new release including this code. Finally, they try uploading some **new** content.

If we don't look at the languages of all the already uploaded content, then Kurdî wouldn't be in the list of languages for the new content form, which seems pretty terrible.

But if they already have uploaded lots of content, then the new content form might take ages to display. And it will get worse for every content they upload.

What I'd like to do in order to fix this is one-time operation that looks at the `Setting` instance for the local languages, and if it doesn't find any then it creates one with the list of languages for all the already uploaded content.

An additional benefit is that the code which gets the list of local languages can be simplified by dropping its fallback and assuming that there is always a `Setting` instance for the local languages.

I was thinking of doing a migration for that, but due to some technicalities involving the ordering of migrations and especially migrations in the content apps, I now think that it would be a pretty terrible idea.

I think the next best thing would be to run that as a `post_migrate` signal. (much like the search app does)

What do you think?

bochecha:
> Imagine a box has uploaded some content in Kurdî, before we merge this pull request. Then, we update their box with a new release including this code. Finally, they try uploading some new content.

> If we don't look at the languages of all the already uploaded content, then Kurdî wouldn't be in the list of languages for the new content form, which seems pretty terrible.

We've decided that for now, this isn't something we'll worry about, and those deployments will manually need to change their language settings.

ideascube/conf/base.py
```diff
@@ -105,27 +105,15 @@
 
 USE_TZ = True
 
-# Languages that will be available in all forms.
+# Languages that will be available as UI translations
 AVAILABLE_LANGUAGES = (
```
yohanboniface: I think at this point we could just use `LANGUAGES`, also removing the postprocess in `settings.py`.

ideascube/conf/base.py
```diff
@@ -298,6 +286,14 @@
         'description': _('Change the server name.'),
         'fa': 'info-circle',
     },
+    {
+        'is_staff': True,
+        'category': 'manage',
+        'url': 'server:languages',
+        'title': _('Languages'),
+        'description': _('Select enabled languages.'),
```
yohanboniface: What about "Enable or disable languages"?

ideascube/models.py
```diff
@@ -311,6 +311,13 @@ def get_prep_value(self, value):
         return json.dumps(value)
 
 
+class LanguageField(models.CharField):
+    def __init__(self, *args, **kwargs):
+        kwargs['choices'] = get_all_languages()
+
```
yohanboniface: why this empty line? :)

yohanboniface:
Very nice move! I'm looking forward to see this merged! :)
One thing we'll need before merging I think is a command line to change the languages setting, so it will be possible for ansiblecube to set it during the deployment (when we have 40 koombooks to deploy in a once).

ideascube/conf/base.py
bochecha: I thought about that, yes, but thought it might be too much, maybe you had a reason to have it split between `base` and `settings`... Then I forgot to ask. :smile: 

But if there isn't such a reason, then yes, let's move it all to `base`. Will do when I get a chance.

ideascube/conf/base.py
bochecha: Sure, will do.

ideascube/models.py
bochecha: Why not?

I guess I just prefer separating my own processing from the parent « normal » stuff, feels a bit more readable...

In general, I tend to be very generous with empty lines, as I do feel that makes the code more readable. A bit like the pragraphs in this response, it's all about grouping your prose/code into semantic units. :wink:

yohanboniface: One thing I really like in python is that those choices are usually not subjective: where to put a space, a blank like, etc. and this make easy for someone to read the code from someone else, because we like at home.
So my opinion is that it's not a good practice to add "subjective" blank lines. So I'd say: we should not add blank lines if not requested by pep8.
Of course, this is all subjective, but… that what I'm trying to say :p

bochecha: I strongly disagree with it.

Compare:

```python
if foo:
    do_something()
elif bar:
    do_something_else()
else:
    do_something_different()
```

With:

```python
if foo:
    do_something()

elif bar:
    do_something_else()

else:
    do_something_different()
```

Both are correct according to pep8. Yet the former is much less readable.

Blocks are like paragraphs in prose. Clearly delineating them is important to readability, otherwise the reader needs to delineate them mentally and that requires more mental effort.

Python is all about that: readability, and reducing the required mental effort to understand a piece of code. Therefore, I contend that proper use of whitespace, even when not required, is in fact the Pythonic thing to do.

Now of course, this is an argument against « we shouldn't add blank lines if not requested by pep8 » and is entirely unrelated to the specific case at hand in this review, but you moved the discussion away from it. :wink: 

---

As for this specific line, I don't care one bit, as I thought was clear from my original response, as it didn't contain any strong argument in favour of it and was written in an introspective, unassertive fashion.

yohanboniface: > Yet the former is much less readable.

That is *subjective*, and it's my purpose here. ;)
pep8 is about setting up conventions so we reduce the subjective aspects of code formatting, and thus the energy needed to read someone else code.
And this is why I think we should restrict as much as possible personal formatting.
So I think we should follow a mantra like: "if a space or a blank line is not required by pep8 we should not add it".
That said, I'm not arguing about this specific line, but more about our general habits as a team.

bochecha: >> Yet the former is much less readable.
>
> That is subjective, and it's my purpose here. ;)

I disagree. Elegance is subjective, readability isn't. But even if it were, does it actually matter?

You're not going to convince me that compact, blank-lines-free code is more readable, and maybe I won't convince you of the opposite. Does it actually matter?

We are individuals, we have our own subjectivity, and yes, sometimes that will show in the code. But let's not make the opposite mistake of pushing towards uniformity.

To quote pep8 itself:

> A Foolish Consistency is the Hobgoblin of Little Minds

---

> So I think we should follow a mantra

I think mantras are generally a bad idea, as they inevitably push us towards inflexibility and following them as if they were hard rules. Instead we should just think for ourselves and write clear code, often on a case by case basis. Sometimes it will mean adding white lines where they aren't dictated by pep8, and sometimes it won't matter at all.

I started this thread not caring, but through what I feel is a completely useless squabble over trivialities, you're honestly pushing me towards actually caring and wanting to stick to it.

Some things do matter, but this isn't one of them.

Feel free to remove that line if it bothers you that much. I feel like I've already wasted way too much time over this, so this will be my last reply in this thread.

ideascube/conf/base.py
bochecha: Looking at it, moving `LANGUAGES` to `ideascube.conf.base` would mean moving also `EXTRA_LANG_INFO` to it.

That would remove from local configs the possibility to set there `EXTRA_LANG_INFO`. Do we want this?

bochecha:
> One thing we'll need before merging I think is a command line to change the languages setting, so it will be possible for ansiblecube to set it during the deployment (when we have 40 koombooks to deploy in a once).

I'm wondering if we'd want a single configuration management command, or a specific one for each setting.

Something like `gsettings` in GNOME might be nice, indeed.

I'll try looking at that in a separate branch, as I do think it is a different (although related) issue.

ideascube/conf/base.py
bochecha: Done in the latest push to this branch.

yohanboniface:
> I'm wondering if we'd want a single configuration management command, or a specific one for each setting.

I'd vote for a generic one, we'll have more and more settings I think.

> I'll try looking at that in a separate branch, as I do think it is a different (although related) issue.

True, but imho we still need to land them in the same time, so they can manage languages from ansiblecube.

bochecha:
> True, but imho we still need to land them in the same time, so they can manage languages from ansiblecube.

Sure, that's why I started working on the command-line this morning. ;)

bochecha:
As you can see from the wall of commits included in this pull request, I've rebased this branch on top of two other branches:

* configuration-app, with its pull request #441 
* configuration-cli, which implements the command-line tool we were discussing above, and for which I'll send a pull request once the previous one is merged.

So don't merge this pull request for now, let's focus on the other 2 first.

ideascube/conf/base.py
```diff
@@ -307,7 +301,15 @@
...
         'description': _('Manage home page'),
         'fa': 'th',
-    }
+    },
+    {
+        'is_staff': True,
+        'category': 'manage',
+        'url': 'server:languages',
+        'title': _('Languages'),
+        'description': _('Enabled or disable languages.'),
```
yohanboniface: Oops, typo.

bochecha: Fixed in the latest push to this branch.

PR #428, opened 2016-07-28
Test all the config files

### LINKED ISSUE #395: Add a tests loading every configuration file
Just as a life guard, in case we have a syntax issue or any format that has changed.
Those file are not loaded by any code but the final box in production otherwise.

bochecha:
We regularly make little syntax errors in the config files, leading to lots of pain down the road.

With this test, config files will be validated at each push to the ideascube repository, which should make us happier.

And in fact, it actually found a broken config!

That config is old, and probably worked at the times of Python 2, but we moved to Python 3 since then. Good thing that box was never updated. :-/

Fixes #395


yohanboniface:
Excellent! :)

PR #429, opened 2016-07-29
Be smarter about when to reindex the database

bochecha:
See each commit for rationale.

### DIFF
--- a/ideascube/search/apps.py
+++ b/ideascube/search/apps.py
@@ -1,12 +1,12 @@
 from django.apps import AppConfig
 from django.db.models.signals import post_migrate
 
-from .utils import create_index_table
+from .utils import reindex_content
 
 
 def create_index(sender, **kwargs):
     if isinstance(sender, SearchConfig):
-        create_index_table(force=False)
+        reindex_content(force=False)
 
 
 class SearchConfig(AppConfig):
--- a/ideascube/search/migrations/0001_initial.py
+++ b/ideascube/search/migrations/0001_initial.py
@@ -1,8 +1,14 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from django.db import migrations, models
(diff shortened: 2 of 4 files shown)


ideascube/search/migrations/0002_reindex.py
```diff
@@ -1,20 +0,0 @@
-# -*- coding: utf-8 -*-
-from __future__ import unicode_literals
-
-from django.db import migrations, connections
-from ideascube.search.utils import reindex_content
-
-
-def update_reindex(*args):
-    cursor_default = connections['default'].cursor()
-    cursor_default.execute("DROP TABLE IF EXISTS idx")
```
mgautierfr: This is not done anymore.
Maybe move it in utils.create_index_table ? Even if an **create**_index_table should not make migration or delete table.

bochecha: Oh, I see what you mean. You're right, I completely missed that it was deleting the table in the **default** database.

So maybe instead of completely removing `0002_reindex`, we should transform it into `0002_remove_old_index_on_default_db`?

mgautierfr: Yes, that better than simply removing it.

(keeping the "IF EXISTS" in the request as 0001_initial do not create the db anymore)
Thinking of that, does 0001_initial have to create the index table in transient as it was creating the table in default before ? I don't think it is a problem, but I prefer ask now than fixing a bug later.

bochecha: Yes, the `0001_initial` does create the table in the transient db now.

This is all a bit of a mess, though.

Ideally we'd just want to completely remove all migrations from the search app, because the model is unmanaged, and as such we actually don't want to use migrations for it.

bochecha: Fixed in the latest push to this branch.

PR #430, opened 2016-07-29
Add a system user

bochecha:
This user won't be allowed to log in, so it won't be used by any physical person behind their screen.

However, it will be useful to change in-db settings in migrations or in a command-line too, since Setting instances require an actor.

It is a hidden user, to be used exclusively by Ideascube itself.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -23,6 +23,12 @@ def staffuser():
     return UserFactory(short_name="Hello", password='password', is_staff=True)
 
 
+@pytest.fixture()
+def systemuser():
+    # Create it the same way the migration does it
+    return UserFactory(serial='__system__', full_name='System', password='!!')
+
+
 @pytest.fixture
 def setup_dirs(monkeypatch, tmpdir, settings):
     storage_root = tmpdir.mkdir('storage')
--- a/ideascube/migrations/0009_add_a_system_user.py
+++ b/ideascube/migrations/0009_add_a_system_user.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.contrib.auth import get_user_model
+from django.db import migrations
+
+
+def add_user(*args):
+    User = get_user_model()
+    User(serial='__system__', full_name='System', password='!!').save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('ideascube', '0008_user_sdb_level'),
+        ('search', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunPython(add_user, None),
+    ]
--- a/ideascube/models.py
+++ b/ideascube/models.py
@@ -55,6 +55,19 @@ class UserManager(BaseUserManager.from_queryset(UserQuerySet)):
         user.save(using=self._db)
         return user
 
+    def get_queryset(self):
+        return super().get_queryset().exclude(serial='__system__')
+
+    def get_queryset_unfiltered(self):
+        return super().get_queryset()
+
+    def all(self, include_system_user=False):
+        if include_system_user:
+            return self.get_queryset_unfiltered().all()
+
+        else:
+            return super().all()
+
 
 class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
     """
@@ -101,6 +114,18 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
 
     def has_module_perms(self, app_label):
         return True
+
+    def validate_unique(self, exclude=None):
+        super().validate_unique(exclude=exclude)
+
+        if self.id is None:
+            # We are creating a new user.
+            # Include system user in the check for uniqueness.
+            all_users = User.objects.all(include_system_user=True)
+
+            if all_users.filter(serial=self.serial).exists():
+                raise ValidationError({
+                    'serial': _('User with this Serial already exists.')})
 
     @classmethod
     def get_data_fields(cls):
--- a/ideascube/tests/test_models.py
+++ b/ideascube/tests/test_models.py
@@ -37,6 +37,18 @@ def test_create_superuser():
     assert user.is_staff
 
 
+def test_list_users():
+    model = get_user_model()
+    model.objects.create_user('123456')
+    model.objects.create_user('__system__')
+
+    users = model.objects.all()
+    assert [u.serial for u in users] == ['123456']
+
+    users = model.objects.all(include_system_user=True)
+    assert [u.serial for u in users] == ['__system__', '123456']
+
+
 def test_client_login(client, user):
     assert client.login(serial=user.serial, password='password')
 
--- a/ideascube/tests/test_views.py
+++ b/ideascube/tests/test_views.py
@@ -150,6 +150,11 @@ def test_user_list_page_should_be_accessible_to_staff(staffapp, user):
     response.mustcontain(str(user))
 
 
+def test_system_user_does_not_appear_in_list(staffapp, systemuser):
+    response = staffapp.get(reverse('user_list'), status=200)
+    response.mustcontain(no=str(systemuser))
+
+
 def test_user_list_page_is_paginated(staffapp, monkeypatch):
     monkeypatch.setattr(UserList, 'paginate_by', 2)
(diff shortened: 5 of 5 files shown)


ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from ideascube.models import User
+
+
+def add_user(*args):
+    User(serial='__system__', password='!!').save()
```
yohanboniface: Maybe add a full_name (like "System") so when we'll display the user in the front it will be more user friendly?

bochecha: Except it's not shown in the front. :smile:

yohanboniface: I know by it may be some day (for example when we'll display who edited the settings), and it will be too bad to need a migration just for that at this point.

bochecha: > when we'll display who edited the settings

That's a very good point, I'm convinced, I'll add a `full_name` when I get the chance.

bochecha: Done in the latest push to this branch.

ideascube/models.py
```diff
@@ -55,6 +55,12 @@ def create_superuser(self, serial, password, **extra):
         user.save(using=self._db)
         return user
 
+    def list_users(self):
+        users = self.get_queryset()
+        users = users.exclude(serial='__system__')
```
yohanboniface: `return users.exclude…` ?

yohanboniface:
I wonder if we should not also prevent the system user to be edited from the UI and even deleted.

bochecha:
> I wonder if we should not also prevent the system user to be edited from the UI and even deleted.

Did I not do that? I thought I had!

Hmm, right, I can edit it by going to http://localhost:8000/en/user/1/ directly.

Not the easiest thing to find, but security through obscurity is definitely not the best thing.

I'll fix it.

yohanboniface:
> Not the easiest thing to find

With the current core, I think one can find the user while searching.

bochecha:
> With the current core, I think one can find the user while searching.

Indeed, that too.

bochecha:
I've reworked the way I hide the system user, and it should now be blocked from viewing, editing, deleting or changing its password.

Should be good to review again.

ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from ideascube.models import User
```
yohanboniface: I wonder if we should be importing user from `get_user_model` instead here.

ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def add_user(*args):
+    User(serial='__system__', full_name='System', password='!!').save()
```
yohanboniface: Do we really need a password? And if yes to we need it in clear?

ideascube/models.py
```diff
@@ -55,6 +55,19 @@ def create_superuser(self, serial, password, **extra):
         user.save(using=self._db)
         return user
 
+    def get_queryset(self):
+        return super().get_queryset().exclude(serial='__system__')
+
+    def get_queryset_unfiltered(self):
+        return super().get_queryset()
+
+    def all(self, really=False):
```
yohanboniface: Not a big fan of the kwarg pattern that changes a default Django method.
What about a separate method (`unfiltered`?) instead?

ideascube/models.py
```diff
@@ -102,6 +115,17 @@ def has_perm(self, perm, obj=None):
     def has_module_perms(self, app_label):
         return True
 
+    def validate_unique(self, exclude=None):
+        super().validate_unique(exclude=exclude)
+
+        if self.id is None:
+            # We are creating a new user
+            same = User.objects.all(really=True).filter(serial=self.serial)
```
yohanboniface: I'd have used the `.exists()` method: https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.exists

ideascube/migrations/0009_add_a_system_user.py
bochecha: The fact that it is stored like this means that no hashed password will ever match this `!!` value.

ideascube/migrations/0009_add_a_system_user.py
bochecha: You are right, fixed.

ideascube/migrations/0009_add_a_system_user.py
bochecha: Incidentally, this is exactly how this works on Linux. :)

```
# grep mathieu /etc/shadow
mathieu:the_hash_of_my_password:16988:0:99999:7:::
# useradd tmptest
# grep tmptest /etc/shadow
tmptest:!!:17032:0:99999:7:::
```

That doesn't mean it is a good idea, but I figured I would at least mention where I took the idea from.

ideascube/models.py
bochecha: Isn't that exactly what kwargs are for, optionally influencing the behaviour of a method?

I don't mind adding a second method, bu it just feels like a waste of a perfectly good and pythonic coding pattern.

ideascube/models.py
bochecha: Doh!

Fixed in the latest push to this branch.

ideascube/migrations/0009_add_a_system_user.py
yohanboniface: > The fact that it is stored like this means that no hashed password will ever match this !! value

Yeah, makes sense, thought about it after closing my computer :p

ideascube/models.py
yohanboniface: OK, deal. :)

yohanboniface: Second though: what about renaming the kwarg to `include_system_user`?

conftest.py
```diff
@@ -23,6 +23,12 @@ def staffuser():
     return UserFactory(short_name="Hello", password='password', is_staff=True)
 
 
+@pytest.fixture()
+def systemuser():
+    # Create it the same way the migration does it
+    return UserFactory(serial='__system__', full_name='System', password='!!')
```
yohanboniface: humm, at the end, why not making `add_user` (from the migration) an util we call from the tests too?

ideascube/models.py
bochecha: I had used `really` as in « yes I **really** want all of them, even the ones we normally would filter ».

Thinking about it some more now, we're probably never going to hide other users.

In addition, if we went the way you suggest and decided to hide more users in the future, we'd have to change that kwarg in favour of something expressing there are other users hidden. Doing that would be trivial, as a quick `grep include_system_user` would return all the occurences we'd have to change.

However, a `grep really` is sure to return lots and lots of results irrelevant in this context.

So yes, I agree with your suggestion, will change this ASAP.

conftest.py
bochecha: Where would that util live?

Would it be a generic `add_user` function that we'd have to use everywhere? (and if so, why not utils for every other model?) Or just an `add_system_user` util?

I'm not entirely convinced this is needed, since it's only in two places that we ever need to create a system user: in a migration and in a test fixture.

Sure, arguably it would ensure that we create them the same way.

However I would say that is wrong too, for a simple reason: if we ever need to change that system user in the database, modifying the `add_system_user` util would **not** be enough, we would need a migration that **modifies** the user.

In fact, in the future if we want to modify that system user, we would:

* create a migration that **modifies** the user in the DB
* modify the code **creating** the user in the tests accordingly
* **not** modify the original migration which **created** the user in the DB

In effect, the two code parts (migration and fixture) are divergent, and I'm not sure it makes sense to try and make them use the same code.

ideascube/models.py
bochecha: Fixed in the latest push to this branch.

PR #434, opened 2016-08-11
Idb grc malakasa

letompouce:
Dirty conf file needs more love!

### DIFF
--- a/ideascube/conf/idb_grc_malakasa.py
+++ b/ideascube/conf/idb_grc_malakasa.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""Ideasbox Malakasa, Grece"""
+from .idb import *  # noqa
+from django.utils.translation import ugettext_lazy as _
+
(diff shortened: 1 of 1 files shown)


ideascube/conf/idb_grc_malakasa.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
```
yohanboniface: (I know, I know, it's everywhere else but…) we can get read of this now that we are in python 3 :)

ideascube/conf/idb_grc_malakasa.py
```diff
@@ -0,0 +1,93 @@
...
+"""Ideasbox Malakasa, Grece"""
+from .idb import *  # noqa
+from django.utils.translation import ugettext_lazy as _
+
+#IDEASCUBE_NAME = u"Malakasa"
```
yohanboniface: You can remove comment, imho.

PR #435, opened 2016-08-11
Fix book's import from isbn.

### LINKED ISSUE #433: ISBN import broken
erreurimportisbn

Seems to have been forgotten when switching to python 3 :/

mgautierfr:
Fix #433.

- Strip the isbn to be sure there is no space.
- Decode the request return as json parse str, not bytes
- If we don't want to fail in any way, do not return None.


yohanboniface:
> If we don't want to fail in any way, do not return None.

:D

Issue #375, opened 2016-07-07
Display cards for installed packages


yohanboniface:
Challenge: define how to group cards (for example to group all wikipedia packages (by language) in the wikipedia card).

mgautierfr:
In the catalog specification we've made a kiwix hackathon (http://www.kiwix.org/wiki/Catalog), we've had a 'content_group' attribute for this case.

yohanboniface:
Very quick and rough follow up after backlogging on IRC:
- I suggest in a first time we do not add a card model
- we want a dedicated Setting that will only store card ids (and certainly a boolean status)
- every content package (even the media center ones) when installed will register itself as card (so appending its id to the dedicated setting)
- a staff user can then decide to show/remove a card (by changing its status in the setting)
- we don't have submenus anymore in the cards
- I suggest in a first time we do not are about cards order: they are in the order they have been installed; we'll see in a later step to let the user reorder them
- card contents are all in HTML files, and if no dedicated HTML exist for the given id, we use the package metadata with the generic HTML template, more or less like today
- we only are talking about package content related cards: admin and system cards will still be only in settings files for now

Moving to the airport now! :)

mgautierfr:
This is not so simple :)

There is several things needed to correctly render a cards:
- The name of the package itself (wikipedia_en, wikipedia_fr, wikibook_en, france.maps)
- The name of the card (ie, the name of the html template to use) (wikipedia, wikibook, maps, ...)
- Data that need to be used by the template (lang, url (that may be constructed at rendering), name)

At first look, we can say that we can infer the two latters from the first one and that is.
However, let's see how the card rendering is made :
Cards in the config are "complex" object (dict), that may contains an 'id' (defaulting to 'custom' if not present). This id is used as the card template name to use.
What you propose is that we store in the settings another dict object that behave differently (there are the id of the package to use (not the card template name), and a boolean to display it or not).

But, as we want to keep old admin and system cards, we have to handle the two different dict nicely, at choice :
- In the index.html template. If the dict as a package_id, ask the catalog for the package metadata, look in that metadata which template to use, render this template. Else check for 'id' and render this template, 'else' ... This seem to me a bit to complex for a html template. And error prone (what if a admin template has a 'package_id' metadata ?)
- Do not change the index.html and reconstruct a correct card dict from the 'card_package_settings' dict in the view. Thus, we get a string from the settings db, json parse it to get a complex object ... and recreate another complex objects cause the ones we've got are not what we need.
[...]

PR #436, opened 2016-08-12
Package card auto

mgautierfr:
This branch follow the second proposition of #375

There are a lot of dependencies on the catalog module (In the index, mediacenter view, search, ..)
Not sure it's a problem right now, but we should find a way to decouple module, and try to cache useful information (instead of creating a catalog instance each time)

For your proposition, ybon, to have the settings also pointing to admin and system cards, I think the simpler is to add a id to the cards in the config, add the ids of the cards we want to display and let the function build_package_card_info look in the config to rebuild the cards info.


bochecha:
System user was merged in #430, please rebase.

conftest.py
```diff
@@ -22,6 +22,10 @@ def user():
 def staffuser():
     return UserFactory(short_name="Hello", password='password', is_staff=True)
 
+@pytest.fixture()
+def systemuser():
+    return UserFactory(serial='__system__', password='!!')
```
bochecha: This fixture was added in the "system user" pull request. You can just remove it from yours when you rebase.

ideascube/models.py
```diff
@@ -109,6 +109,10 @@ def has_module_perms(self, app_label):
         return True
 
     @classmethod
+    def get_system_user(cls):
+        return cls.objects.get(serial='__system__')
```
bochecha: This won't work, because the default queryset (see #430) completely hides the system user.

In addition, this should probably be a manager method?

ideascube/configuration/registry.py
```diff
@@ -22,6 +22,12 @@
             'default': getattr(settings, 'IDEASCUBE_NAME', 'Ideas Cube'),
         },
     },
+    'ideascube': {
```
bochecha: All config options are ideascube configs. Maybe `home` would be a better namespace?

Also, I prefer my bikeshed painted black!

ideascube/serveradmin/catalog.py
```diff
@@ -126,6 +129,10 @@ def remove(self, package):
         self._removed.append(package)
 
     def commit(self):
+        visible_packages = get_config('ideascube', 'displayed-package-ids')
+        [visible_packages.remove(p.id) for p in self._removed if p.id in visible_packages]
```
bochecha: Wow. This is a bit convoluted. You're creating a list which you don't keep, just for the side-effect of modifying another list.

I'd prefer something like this:

```python
for pkg in self._removed:
    try:
        visible_packages.remove(pkg.id)
    except ValueError:
        # It's not in there, nothing to remove them
        pass
```

Or we could make `self._removed` (and `self._installed`) a dictionary, using the package ids as keys, and then you could do:

```python
visible_packages = [id for id in visible_packages if id not in self._removed]
```

Or even something like this, keeping `self._removed` the list it currently is, why not:

```python
removed_ids = [p.id for p in self._removed]
visible_packages = [id for id in visible_packages if id not in removed_ids]
```

Or let's be crazy !

```python
removed_ids = [p.id for p in self._removed]
visible_packages = list(set(visible_packages) - set(removed_ids))
```

Each of these requires less mind contortion than yours, because they are variation of the typical `filter` pattern, and this is exactly what you are trying to do: filtering elements out of a list.

I honestly needed to pause and re-read your line a few times, drawing the connections mentally to figure out what it was doing.

Simple things should be done in obvious ways. Filtering items out of a list is a very simple (and extremely common) thing to do.

ideascube/serveradmin/catalog.py
```diff
@@ -126,6 +129,10 @@ def remove(self, package):
...
+        visible_packages.extend(p.id for p in self._installed)
+        set_config('ideascube', 'displayed-package-ids', visible_packages, User.objects.get_system_user())
```
bochecha: The last line is too long.

Also, eventually we might want to use the actual `user` who installed the packages instead of the system user. But that will only make sense when we have a UI to install packages from ideascube (right now we can only use the CLI) so we can cross that bridge when we get there. Something to keep in mind, though.

bochecha: So with this commit, the base `Handler` essentially becomes coupled with the Ideascube DB.

I'm not sure I'm comfortable with that.

Wouldn't it be better to move this out of the `Handler.commit` method and into the `Catalog` instead?

The `Catalog` is a higher level thing. If we decided to split all this code out of Ideascube then the base `Handler` certainly wouldn't be able to access the Ideascube DB any more.

But moving this code to the `Catalog`, then Ideascube could extend the `Catalog` class to make it do those DB operations, and the base `Catalog` wouldn't have to know about the DB.

Does that make sense?

bochecha:
> - The card info are build from the package info from the catalog.
> - No more several subnav per card (lang, maps)
> - The card get from the config is STAFF_HOME_CARDS. No more HOME_CARDS.
>   As the card templates have changed, the values in the config will not
>   be up to date, and the templates will not apply. (This should be fixed)

I'm not a fan of bullet points for commit messages, but whatever, that's not important.

A few things about this commit message, though:

> The card info are build

"built"

> No more several subnav

That doesn't make sense. Do you mean that there is no more subnav? Or that we don't need to have several subnav any more (i.e just one) ? Also, what is a "subnav" ? :smiley: 
 
> The card get from the config is STAFF_HOME_CARDS.

What does this mean?

> (This should be fixed)

Then why isn't it fixed?

conftest.py
```diff
@@ -122,3 +123,36 @@ def pytest_configure(config):
     #
     # Somehow the default database is always in memory, though.
     settings.DATABASES['transient']['TEST_NAME'] = ':memory:'
+
+
+@pytest.yield_fixture()
+def installed_packages_catalog():
+    with mock.patch('ideascube.serveradmin.catalog.Catalog', spec=True) as MockClass:
```
bochecha: You can use the `mocker` fixture, which avoids the top-level import.

No, it does not change anything that actually matters, just mentioning it. :smile:

conftest.py
```diff
@@ -122,3 +123,36 @@ def pytest_configure(config):
...
+        pseudo_package1 = mock.NonCallableMock(spec_set='id name description template_id language staff_only'.split())
```
bochecha: I didn't know about `spec_set`, that's really cool!

(Line is too long though, if you had used a list instead of `split`ting a string you could have more easily wrapped it.)

conftest.py
```diff
@@ -122,3 +123,36 @@ def pytest_configure(config):
...
+                                                    'A package without template'
+                                       )
+
+        instance = MockClass.return_value
+
+        instance.list_installed.side_effect = lambda ids_list: [
+            p for p in (pseudo_package1, pseudo_package2, pseudo_package3)
+            if ( (p.id in ids_list) or ('*' in ids_list) )]
+
+        yield MockClass
```
bochecha: Any reason you used a `yield_fixture` if you're not doing anything after `yield`ing?

bochecha: Love the way you're mocking the Catalog here, I've had difficulties doing similar things in the past because the `mock` documentation lacks clear examples, and this is a great one!

ideascube/views.py
```diff
@@ -25,23 +25,49 @@
 from ideascube.library.models import Book
 from ideascube.mediacenter.models import Document
 
+# Do not import directly Catalog as we want to mock it for test.
+from ideascube.serveradmin import catalog as catalog_mod
```
bochecha: You're using `catalog_mod` here to avoid clashing between the `catalog` module and your `catalog` variable.

Why not directly import `from ideascube.serveradmin.catalog import Catalog` since that's the only thing you ever seem to use from the module?

ideascube/views.py
```diff
@@ -25,23 +25,49 @@
...
+
 from .forms import UserForm, CreateStaffForm
 from .mixins import CSVExportMixin
 
 user_model = get_user_model()
 
+def build_package_card_info():
+    package_card_info = []
+    catalog = catalog_mod.Catalog()
+    packages_to_display = catalog.list_installed(get_config('ideascube', 'displayed-package-ids'))
```
bochecha: Things like these make me think that the `Catalog` API is not that bad after all. :smile: 

Line is too long, though. :wink:

ideascube/views.py
```diff
@@ -25,23 +25,49 @@
...
+
+    for package in packages_to_display:
+        template_id = getattr(package, 'template_id', None)
+        if template_id is None:
+            # We probably have a catalog not up to date.
+            # We don't know how to display the card, so ignore it.
+            continue
```
bochecha: Log an error maybe?

ideascube/mediacenter/views.py
```diff
@@ -16,6 +16,9 @@
 from ideascube.decorators import staff_member_required
 from ideascube.mixins import FilterableViewMixin, OrderableViewMixin
 
+# Do not import directly Catalog as we want to mock it for test.
```
bochecha: You could still mock it even if you imported it directly, though. Why would it be a problem?

ideascube/mediacenter/views.py
```diff
@@ -47,9 +50,17 @@ def _set_available_kinds(self, context):
...
+
+        context['available_package_ids'] = {
+            package.id:package.name for package in catalog.list_installed(available_package_ids)}
+
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         self._set_available_kinds(context)
+        self._set_available_package_ids(context)
         self._set_available_langs(context)
         self._set_available_tags(context)
```
bochecha: Not your fault, but this pattern it pretty ugly. Passing a mutable to a function that modifies it is fine in C where it's often the only sane way to do something, but it's really not pythonic.

Plus, modern Django (since 1.5) passes the `view` instance to the template, so the above code is the perfect example of when it would be better to simply use:

```python
    @property
    def available_kinds(self):
        ...

    @property
    def available_package_ids(self):
        ...

    ...
```

And then in the template just use `{{ view.available_kinds }}`, etc...

Again, that code predates your change (and probably also predates django 1.5 :smile: ) so I'm not sure you should move to the new style in the middle of code doing the old style. I'm just mentioning it because few people seem to know about this, and to try and sway us gently towards the new style in the future.

ideascube/serveradmin/catalog.py
mgautierfr: This is not all the same (in spirit)

My version (and your "to long" version) loop over self._removed.
Your other versions loop over visible_packages.

But at the end, all versions finish by looping over both self._removed and visible_packages, so its more a conceptual difference that a true one. (But it can be in another context)

The [some(foo) for foo in foos if test(foo)] seems to become the standard way of writing (composition of) map/filter.

Even Guido suggests to remove map/filter/reduce from python3 (http://www.artima.com/weblogs/viewpost.jsp?thread=98196)
From now on, I tend towards this syntax all the time.
It's now a habit I have to just "drop" the "[]" in my mind in this context and read the line as a powerful map/filter syntax.

I would prefer have a standard way of looping over a iterator while dropping the element to write something like `loopover(visible_packages.remove(p.id) for p in self._removed if p.id in visible_packages)` and avoid create a list. But loopover is missing in python.

I think that `visible_packages = [p for p in visible_packages if p.id not in self._removed]` would be ok for every one (?)

ideascube/serveradmin/catalog.py
mgautierfr: Yes we could move the dependency to the DB in the Catalog.

However, your second proposition to let Ideascube extend the Catalog to make the DB operation seems overengineering right now.
It may be a good idea when we move the catalog out of ideascube but now, it will end with a base class never directly used and extended by only one child class.

ideascube/conf/base.py
```diff
@@ -298,6 +298,14 @@
         'description': _('Change the server name.'),
         'fa': 'info-circle',
     },
+    {
+        'is_staff': True,
+        'category': 'manage',
+        'url': 'server:cards',
+        'title': _('Cards'),
+        'description': _('Manage displayed cards'),
```
bochecha: I wonder if « cards » isn't internal jargon. Is this something users generally understand in the field?

To me at least, the only place I've ever heard of "cards" are in the source code and the settings files, so if I wasn't already familiar with these I don't think I would understand.

Maybe at least change it to « Manage cards displayed on the home page », which could be clearer?

Or maybe we should just talk about "apps", as that's what users tend to be more familiar with... But then my view might be biased by the fact my sample for "users" only includes people in developed countries with access to an app ecosystem on their computer/smartphone, and as such our target users might not know of "apps" any better than they do of "cards", so that's a moot point.

ideascube/conf/base.py
```diff
@@ -298,6 +298,14 @@
...
+        'fa': 'eye',
```
bochecha: This seems like the wrong metaphor to use here.

I'd use `th` or `th-large` instead.

ideascube/serveradmin/templates/serveradmin/index.html
```diff
@@ -22,6 +22,7 @@
             <li>{% fa 'life-ring' 'fa-fw' %} <a href="{% url 'server:backup' %}">{% trans "Manage backups" %}</a></li>
             <li>{% fa 'plug' 'fa-fw' %} <a href="{% url 'server:battery' %}">{% trans "Monitor battery" %}</a></li>
             <li>{% fa 'wifi' 'fa-fw' %} <a href="{% url 'server:wifi' %}">{% trans "Manage Wi-Fi" %}</a></li>
+            <li>{% fa 'eye' 'fa-fw' %} <a href="{% url 'server:cards' %}">{% trans "Manage visible cards" %}</a></li>
```
bochecha: Same comment about the `eye`.

bochecha: You use "visible" here, "displayed" on the card itself. Can we use the same in both places?

conftest.py
mgautierfr: This is not exactly the same :)
The mocker.path if a wrapper around mock.path (or unittest.path, it depends of the python version. It unittest.path in our case)

One important change is that it start/stop the mock automatically.
But if we use the `with` syntax, the mock is then start/stop a second time.
I don't know if it works (and I don't want to know)

Another change : mocker do not have all the attribute of mock. (ie NonCallableMock)

conftest.py
mgautierfr: I discover it for this patch :)

Without it, `hasattr(pseudo_package3, 'template_id') == True)` :/

ideascube/views.py
mgautierfr: As said in the comment, it is to be able to mock the Catalog class.

If we `from ideascube.serveradmin.catalog import Catalog`, the Catalog class is directly referenced. So a mock.patch of `ideascube.serveradmin.catalog.Catalog` will have no effect.

ideascube/views.py
mgautierfr: Log, yes.

ideascube/serveradmin/tests/test_views.py
```diff
@@ -642,3 +642,29 @@ def test_staff_forgets_non_existing_connection(mocker, staffapp):
...
+    set_config('ideascube', 'displayed-package-ids', ['test.package1'], systemuser)
+    res = staffapp.get(reverse("server:cards"), status=200)
+    assert 'Test package1' in res.unicode_body
+    assert 'Test package2' in res.unicode_body
+    assert 'Test package3' not in res.unicode_body
+
+    form = res.forms['cards']
+    inputs = form.fields['visible']
+    for _input in inputs:
+        should_be_checked = (_input._value == 'test.package1')
```
bochecha: `_value` ?

Isn't it a bit risky to access private API?

ideascube/serveradmin/tests/test_views.py
```diff
@@ -642,3 +642,29 @@ def test_staff_forgets_non_existing_connection(mocker, staffapp):
...
+        assert (should_be_checked == _input.checked)
+
+    # inverse the checked
+    for input_ in form.fields['visible']:
```
bochecha: `_input` before, `input_` now... Why not just `input` in both cases? :smiley:

ideascube/mediacenter/views.py
mgautierfr: Yes, but I would have to mock `ideascube.mediacenter.views.Catalog`. (And everywhere else the Catalog class is imported).

Tests should no be dependent from where the class is imported. They should mock the class only where it is defined (ideascube.serveradmin.catalog)

ideascube/mediacenter/views.py
mgautierfr: I didn't know. That's nice.

But the available_* are compute from the context. Is it possible to get the context from the view (self in the property) ?

ideascube/configuration/registry.py
mgautierfr: Yes, `home` is better.

ideascube/conf/base.py
mgautierfr: Maybe see what @yohanboniface or @barbayellow think about this.
They have discussed with users :)

ideascube/conf/base.py
mgautierfr: Didn't know th. Indeed, this is better.

ideascube/serveradmin/templates/serveradmin/index.html
mgautierfr: Agree

bochecha:
> fix 1aa320

Except there is no commit 1aa320 in this pull request, or in master.

Given the code in this commit, I assume it is commit be0523aaa4c0fd7342d39fe36525311d676221cf you're fixing, but you rebased at some point and its hash changed.

In any case, please merge it with the commit it is fixing.

ideascube/conf/base.py
yohanboniface: "manage home page" ?

ideascube/serveradmin/tests/test_views.py
mgautierfr: Well, _input.value return None, if the _input is not checked. :(

ideascube/serveradmin/tests/test_views.py
mgautierfr: Oo, I missed this.
_input everywhere is better.
input is a builtin. I prefer to avoid name clash.

ideascube/conf/base.py
yohanboniface: oops, sorry confused two conversation it seems :p

mgautierfr:
>>    No more several subnav

> That doesn't make sense. Do you mean that there is no more subnav? Or that we don't need to have several subnav any more (i.e just one) ? Also, what is a "subnav" ? :smiley:

No more *several* subnav.
Some card templates have sub links. For example, the wikipedia card has a link per languages.
(And the links have the class subnav in the html/css).
With this commit, all cards now have only one (subnav) link. 

>>    The card get from the config is STAFF_HOME_CARDS.

> What does this mean?

Before this commit, the list of card to display was HOME_CARDS (which was most of the time STAFF_HOME_CARDS + custom cards depending on setting).

Now, we still get cards from the settings\* and we get it from STAFF_HOME_CARDS.

\* This commit message were older than when we decided to call setting from configuration files 'setting' and settings from database 'configuration'.

>>    (This should be fixed)

> Then why isn't it fixed?

Don't know exactly, but I would say cause this has to be fixed in setting files. (I know, it not a reason, but that's a explanation :) )

>>    fix 1aa320

> Except there is no commit 1aa320 in this pull request, or in master.

This should not have been pushed. This is internal to my process to amend commits in my history before pushing a branch.

ideascube/serveradmin/catalog.py
bochecha: > The [some(foo) for foo in foos if test(foo)] seems to become the standard way of writing (composition of) map/filter.

That's not becoming the standard, that's exactly how you compose a map and filter. Or rather, it is usually `bars = [some(foo) for foo in foos if test(foo)]` (or reassigning to `foos` if the initial value is not needed any more).

Except that's not what you did.

What you did is `[foos.remove(foo) for foo in bars if foo in foos]`. There are two key differences:

1. See that `bars` in the middle?
2. During the iteration you're modifying the condition being tested

You're iterating over `bars`, applying a function to each item. So you're pretty much doing a `map`. However, the function you apply has a side-effect that modifies another list which turns out to be your `filter`.

And you're doing that in a list-comprehension, not storing the return value, which makes it look at a first glance like it does nothing. (and is honestly just weird, this is the first time I ever see a list comprehension not being assigned to a variable)

> Even Guido suggests to remove map/filter/reduce from python3 (http://www.artima.com/weblogs/viewpost.jsp?thread=98196)

Cool, thanks for the link, I didn't know about his opinion on this. However you'll notice that I didn't suggest using any of these functions, so I don't see how that's relevant. :wink: 

> I think that visible_packages = [p for p in visible_packages if p.id not in self._removed] would be ok for every one (?)

Nope, it would not work, because `visible_packages` contains package ids, not package objects.

ideascube/search/templates/search/filter_by_package_id.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n ideascube_tags %}
+{% if available_kinds %}
+<div class="card tinted">
+    <h4><span class="theme discover">{% trans "browse" %}</span> {% trans 'by package' %}</h4>
+    {% for id, label in available_package_ids.items %}
```
yohanboniface: I think we are fine for this time, but we'll quickly face the issue of having too much package installed.
Also, I'm not sure the concept of "package" is really end user ready. I'll think about a suggestion.

ideascube/serveradmin/catalog.py
bochecha: > However, your second proposition to let Ideascube extend the Catalog to make the DB operation seems overengineering right now.

I agree, and that's why I didn't say we should do it right now.

ideascube/search/utils.py
```diff
@@ -12,7 +12,7 @@ def create_index_table(force=True):
         cursor_transient.execute("DROP TABLE IF EXISTS idx")
         cursor_transient.execute("CREATE VIRTUAL TABLE idx using "
                                  "FTS4(id, model, model_id, public, text, "
-                                 "lang, kind, tags)")
+                                 "lang, kind, tags, package_id)")
```
yohanboniface: That's a bit sad: I'd love to keep this table out of business specifics.
Can't we sort of manage this with a tag? Not saying it's a better idea, just starting the brainstorming :)

conftest.py
bochecha: > One important change is that it start/stop the mock automatically.

Right, of course, it does that for the lifecycle of the test function.

See, I'm glad I made this seemingly useless comment, because it allwed you to remind me of something I had forgotten. :laughing:

ideascube/views.py
bochecha: Yes, you'd mock `ideascube.views.Catalog` instead, as we normally do.

But now I understand: because you do the mock in a fixture and not in the actual test, you can't know what you would need to mock, because that depends on the test using that fixture.

Hmmm....

Maybe that comment could be improved then, to make it completely obvious and avoidus breaking it in a future « refactoring » ?

Something explicitly saying that we need to be able to mock ideascube.serveradmin.Catalog.Catalog globally for all tests at once?

ideascube/serveradmin/tests/test_views.py
yohanboniface: I sort of remember PEP8 suggesting to append the underscore instead.

ideascube/mediacenter/views.py
bochecha: > Tests should no be dependent from where the class is imported. They should mock the class only where it is defined (ideascube.serveradmin.catalog)

Nope: https://docs.python.org/3/library/unittest.mock.html#where-to-patch

> The basic principle is that you patch where an object is looked up, which is not necessarily the same place as where it is defined.

It is perfectly acceptable for a test of `B` to mock things only in `B` and not touch at `A` where the mocked thing is defined.

And what you did is fine too.

When I started these threads I had just missed a simple fact: you're mocking in a fixture, not in the individual tests, so you have to do it this way.

ideascube/mediacenter/views.py
bochecha: Sure it is, call `self.get_context_data()`.

Store that in a cached property to avoid calling it multiple times.

But it probably would be more interesting to look back at that `_search_for_attr_from_context` method.

Anyway, like I said this isn't something for this PR, I merely wanted to mention it. (and I'm glad I did, since you apparently didn't know about it :smiley: )

ideascube/serveradmin/tests/test_views.py
bochecha: Ah right, I remember we discussed this yesterday.

Why don't you give an `id` to each input? That would allow you to `id`entify them uniquely.

I mean, it's the same in the view, you don't get in `POST` a list of all the checkbox with a `checked` attribute set to `True` or `False`, you only get the list of values of the `checked` checkboxes.

ideascube/serveradmin/tests/test_views.py
bochecha: > input is a builtin. I prefer to avoid name clash.

True, although it is such a rarely used one in the context of a webapp that I don't think it matters one bit in this case.

Leading underscores by convention indicate private attributes, which doesn't make any sense in the local scope, so that seems a bit weird to me. And it might trip over programmers new to python but coming from other languages where `self.attribute` is directly accessible as `attribute`. (which is something I personally detest)

Anyway, it doesn't matter much.

bochecha:
>> No more several subnav.
> With this commit, all cards now have only one (subnav) link. 

Then the way to express this in English is to say « Cards do not have several subnavs any more ».

« No more » indicates there aren't any at all any more.

Still, please refrain from using abbreviations in commit messages. We don't need to count the characters in commit messages, and even Twitter now allows messages of arbitrary length.

I still have no idea what a "subnav" is in this context. :)

>>> The card get from the config is STAFF_HOME_CARDS.
>> What does this mean?
> Before this commit, the list of card to display was HOME_CARDS (which was most of the time STAFF_HOME_CARDS + custom cards depending on setting).
>
> Now, we still get cards from the settings* and we get it from STAFF_HOME_CARDS.

Then why didn't you write this text in the commit message?

I'm not trying to be picky here, I **really** didn't understand what the commit message meant. Now I do. :smile:

ideascube/search/templates/search/filter_by_package_id.html
bochecha: > Also, I'm not sure the concept of "package" is really end user ready.

I'm convinced it isn't, and never will be.

ideascube/serveradmin/catalog.py
mgautierfr: > Nope, it would not work, because visible_packages contains package ids, not package objects.
That explaining why I didn't right it at first time :p

What about this ? :

```python
removed_ids = {p.id for p in self._removed}
visible_packages = [id for id in visible_packages if id not in removed_ids]
```

ideascube/serveradmin/tests/test_views.py
mgautierfr: > Why don't you give an id to each input? That would allow you to identify them uniquely.

Cause this id would be used only for testing. And I would like not modify the generated html just to allow a test to be clever written.

> I mean, it's the same in the view, you don't get in POST a list of all the checkbox with a checked attribute set to True or False, you only get the list of values of the checked checkboxes.

I'm not testing the POST request. I'm testing the generated html.
And all the logic is base on the value attribute. If I start testing id attribute, I'm not testing the right thing.

ideascube/serveradmin/tests/test_views.py
mgautierfr: `input_` then ?

ideascube/views.py
mgautierfr: Yes, I will rewrite this comment.

mgautierfr:
> I'm not trying to be picky here, I really didn't understand what the commit message meant. Now I do. :smile:

Ok, ok, I will rewrite this commit message :smile:

ideascube/serveradmin/tests/test_views.py
bochecha: >> Why don't you give an id to each input? That would allow you to identify them uniquely.
> Cause this id would be used only for testing.

Well, no, it could be used also in the CSS, the Javascript, ... That's really very standard HTML.

> And all the logic is base on the value attribute.

And maybe that's the problem? `value` doesn't have to be unique in an HTML page, it doesn't even have to be unique in a form. Using `value` to identify (see the first 2 letters in that word?) an object seems wrong to me.

It would be like trying to identify an element of a dictionary by its value, rather than by its key.

I really don't see why you wouldn't be testing the right thing here.

You could set `id="pkg_{{ pkg.id }}" value="{{ pkg.id }}"` in your HTML, then in the tests use `id` to identify which ones should be checked, and even verify that `id == 'pkg_%s' % value` if you want.

There's really no reason you can't bind the two like this.

ideascube/views.py
```diff
@@ -25,23 +25,49 @@
...
+    return package_card_info
 
 def index(request):
     if not user_model.objects.filter(is_staff=True).exists():
         return HttpResponseRedirect(reverse_lazy('welcome_staff'))
     content = Content.objects.published().order_by('published_at').first()
     random_book = Book.objects.available().order_by('?').first()
     random_doc = Document.objects.order_by('?').first()
+
+    cards = settings.STAFF_HOME_CARDS + build_package_card_info()
```
yohanboniface: Aren't we loosing the libary, blog, media center (généric) cards?

yohanboniface:
I don't get how you do the mapping package <=> template (i.e. to use map.html for world.map or kiwix.html for wikipedia.en) ?

ideascube/serveradmin/tests/test_views.py
mgautierfr: > Well, no, it could be used also in the CSS, the Javascript, ... That's really very standard HTML.

In our case, it will be use only by tests.

> It would be like trying to identify an element of a dictionary by its value, rather than by its key.

I do not try to identify a field from its value. I loop over all the input field named 'visible' and a do different test depending of their 'value'

> I really don't see why you wouldn't be testing the right thing here.

HTML form use the `value` attribute. This is how the POST list is constructed.
`id` is not used at all by django.
If I test `id` and not `value`, I don't test an attribute used and I test an attribute not used.

mgautierfr:
> I don't get how you do the mapping package <=> template (i.e. to use map.html for world.map or kiwix.html for wikipedia.en) ?

We have to add a template_id in the package's metadata in the catalog.

ideascube/serveradmin/tests/test_views.py
bochecha: > If I test id and not value, I don't test an attribute used and I test an attribute not used.

Wait, where did I say you shouldn't test for the value?

Maybe that wasn't clear, so let me expand in excruciating details. :smile: 

Your test code makes several assumptions: that there will be several checkboxes (you're not verifying this assumption), that one of them will have its value set to `'test.package1'` (you're not verifying that assumption either). In addition you, the programmer, make one more assumption: that only one of these checkboxes will have its value set to `'test.package1`'. (you're not verifying it either)

And then your test verifies that the checkbox with that specific value is checked, and that all others aren't.

You decided to make all those assumptions when writing the code. What I'm arguing for is that you could very well decide to make one more (as long as you verify it) : that for each checkbox the id is set accordingly to the value. (for example `id = 'pkg_%s' % value`)

Making this assumption doesn't change at all what your test is verifying (I'm not suggesting you stop checking for the value), and it allows you to stop using a private API from your dependency, so that seems like a net win to me.

ideascube/conf/base.py
yohanboniface: I suggest "Manage home page".

conftest.py
```diff
@@ -122,3 +123,36 @@ def pytest_configure(config):
     #
     # Somehow the default database is always in memory, though.
     settings.DATABASES['transient']['TEST_NAME'] = ':memory:'
+
+
+@pytest.yield_fixture()
+def installed_packages_catalog():
```
yohanboniface: Not a big fan of such a big mocking feature: the more we mock, the higher the risks of having falsy tests.
Also I generally tend to use mock for I/O (that we cannot reproduce during tests), but

ideascube/search/utils.py
yohanboniface: So: being able to combine other filters with the package_id is very nice. But, adding more column to the Search model does not seem future proof to me.
Also, I think that that a given media or a media is part of a package is only internal business, and I feel like user will get what a package is.
So my suggestion is to use tags for that: when creating a media from a package, we also add its *package name* as tag. And then when linking from a package card in the home page we just use this tag, which will appear as a tag on the filters and so on.
Of course, we may face the case where two packages have the same name (while not the same id), or that someone add a package name as tag to a media not related to this package. I think that's acceptable, and *may* also be considered as a feature :)

ideascube/serveradmin/catalog.py
```diff
@@ -18,10 +18,13 @@
 import yaml
 import mimetypes
 
+
```
yohanboniface: I see a line here ;)

ideascube/serveradmin/catalog.py
yohanboniface: > Wouldn't it be better to move this out of the Handler.commit method and into the Catalog instead?

Agreed

ideascube/serveradmin/templates/serveradmin/cards.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans "Manage displayed cards" %}</h2>
```
yohanboniface: I'd use "Manage home page" too here, then.

yohanboniface:
> We have to add a template_id in the package's metadata in the catalog.

As discussed in IRC, I think the package metadata from a remote catalog should not contain any ideascube specific, given that not only ideascube my consume it.
The overall direction we want to take is: moving as much data to the package metadata (when they are generic, like the icon url), so all package can be displayed with the same generic template, but we'll not be there that soon. Plus we may still want to support custom templates for some specific cases.
So, my suggestion for this iteration is to make the Package *model* being responsible of defining the package `template_id`. For example, the ZippedZim one would always return `kiwix`, while StaticSite would do a sort of regex to define if it's a map or w2eu, etc.
I think the Handler (or maybe the Package itself) should be responsible to define the package URL.
As we only are dealing with a dict at cards building time, I think the url, template_id and such should be inserted into this dict by the Package and Handler.
That's transitional, but I prefer to have ideascube code to deal with that instead of deporting our design troubles to the remote catalogs. ;)
And it's easier to refactor when we don't have to synchronize with third party.
I may have missed some implementation details here, but I think that's the best approach so far to move forward without going a too long path or compromising too much imho.

ideascube/search/utils.py
mgautierfr: I disagree here.
Maybe package_id is a wrong name here, I concede. But it have to be seen as "from where this content comes" and it seems legit to have this information in this table.
Of course, we could put it in tags, but :
- Why not put everything in tag then (lang, kind, text, public) ?
- Tag is somehow free form information. There is no mandatory information nor mandatory format attached to them. If a tag become mandatory... were sure it will be there and we can move it out of tags and give him a real place.
- It can mix a lot with all other tags (not only package name). We cannot differentiate a tag from the other, so we cannot know if a tag is a package name or not.

Package_id (or other name) is a mandatory field. We (will) always have this information, and we need this information to filter media (this is use for by the link from the card).
Of course, we can **also** put the package's name in the tags to allow free search. But this is another feature.

conftest.py
mgautierfr: > Also I generally tend to use mock for I/O (that we cannot reproduce during tests), but avoid them for internal code.

It is not unittest then. If I want to test a function depending of a module, I have to mock the module, not the dependencies of the module.

However, I agree that it mocks a lot of thing, especially creating packages that we may not want. I will try to rewrite it to have a better API.

bochecha:
First commit message:

> The used setting is 'ideascube.displayed-package-ids'

No it's not. :stuck_out_tongue_winking_eye:

ideascube/serveradmin/catalog.py
```diff
@@ -604,9 +606,17 @@ def list_nothandled(self, ids):
 
         return sorted(pkgs, key=attrgetter('id'))
 
+    def update_displayed_packages_on_home(self, *, to_remove_ids=[], to_add_ids=[]):
```
bochecha: Warning! Don't use `[]` as a default value for a keyword argument!

bochecha: Should the function be private?

ideascube/serveradmin/catalog.py
```diff
@@ -701,13 +716,16 @@ def upgrade_packages(self, ids):
             except Exception as e:
                 printerr("Failed installing {0.id}\n".format(upkg))
                 printerr(e)
+                removed_and_not_upgraded_ids.append(upkg.id)
```
bochecha: This seems wrong when thinking of the UX: after an update, it is possible that the user won't see their cards on the home page. :confused: 

I don't think an update should change the setting.

Sure, it is possible that the update somehow fails to install the new version. But then the admin will fix the problem so that the new version ends up on the machine, and the setting should still have it.

conftest.py
```diff
@@ -127,3 +131,43 @@ def pytest_configure(config):
...
+    def list_installed(self, ids_list):
+        if '*' in ids_list:
+            return self.packages
+
+        return [p for p in self.packages if p.id in ids_list]
+
+    def __enter__(self):
+        if self.mocked is not None:
+            raise RunTimeError(
+                "CatalogMocker is not a reentrant context manager")
```
bochecha: Are we, high-level Python developers, all familiar with the concept of reentrancy? Or should that message just state explicitly that you can't use the same `CatalogMocker` instance as context managers without the first one having exited first?

In general, I tend to prefer messages to explain how to fix the issue when it is a programmer error, rather than just stating the issue.

conftest.py
```diff
@@ -127,3 +131,43 @@ def pytest_configure(config):
...
+        return self.mocked
+
+    def __exit__(self, type, value, traceback):
+        self.patcher.__exit__(type, value, traceback)
+        self.mocked = None
+
+
+@pytest.fixture()
+def catalog():
+    return CatalogMocker()
```
bochecha: Shouldn't this fixture require the `db` and `systemuser` ones?

ideascube/serveradmin/catalog.py
```diff
@@ -281,6 +281,14 @@ def assert_is_zipfile(self, path):
             os.unlink(path)
             raise InvalidFile('{} is not a zip file'.format(path))
 
+    # Ideally (in next futur), template_id will solelly be defined from the
```
bochecha: "future", "solely"

ideascube/serveradmin/catalog.py
```diff
@@ -324,6 +332,13 @@ def remove(self, install_dir):
         for path in glob(os.path.join(datadir, '*', zimname)):
             rm(path)
 
+    @property
+    def template_id(self):
+         template_id = self.id.rsplit('.', 1)[0]
```
bochecha: Why not `super().template_id` ?

ideascube/serveradmin/catalog.py
```diff
@@ -346,6 +361,15 @@ class StaticSite(SimpleZipPackage):
...
     handler = Nginx
 
+    @property
+    def template_id(self):
+         if self.id == 'w2eu':
+             return self.id
+         template_id, extension = self.id.rsplit('.', 1)
+         if extension == 'map':
+             template_id = 'maps'
+         return template_id
```
bochecha: So many special cases. :disappointed: 

Isn't it easier to rename the card template file names?

ideascube/templates/ideascube/includes/cards/deus-ex-silicium.html
```diff
@@ -1,3 +1,2 @@
 {% load i18n %}
-{% trans "learn" as theme %}
-{% include 'ideascube/includes/cards/_kiwix.html' with card=card class="deus-ex-silicium" theme_class="learn" name="Deus Ex Silicium" description="Vidéos dédiées à la science des semi-conducteurs" theme=theme %}
+{% include 'ideascube/includes/cards/_kiwix.html' with card=card class="deus-ex-silicium" theme="learn" %}
```
bochecha: It would probably have been better to split this `theme_class` vs `theme` change in its own commit, since it's not related to the package+card management change.

ideascube/tests/test_views.py
```diff
@@ -47,6 +49,10 @@ def test_home_page_should_redirect_to_welcome_staff_if_no_staff(app, user):
     assert response.forms['model_form']
 
 
+def test_home_page_should_not_redirect_if_staff_exists(app, staffuser):
+    app.get(reverse('index'), status=200)
```
bochecha: Where is this coming from?

ideascube/tests/test_views.py
```diff
@@ -511,3 +517,55 @@ def test_by_tag_page_is_paginated(app, monkeypatch):
     assert not response.pyquery.find('.next')
     assert response.pyquery.find('.previous')
     response = app.get(url + '&page=3', status=404)
+
+
+def test_build_package_card_info_must_no_fail_for_no_package(app, systemuser):
```
bochecha: `..._must_not_fail_...`

ideascube/tests/test_views.py
```diff
@@ -511,3 +517,55 @@ def test_by_tag_page_is_paginated(app, monkeypatch):
...
+                           name='Test package2',
+                           description='Test package2 description',
+                           template_id='template_name',
+                           language='fr',
+                           staff_only=False)
+    set_config('home-page', 'displayed-package-ids', ['test.package1.fr'], systemuser)
+
+    # The main information here is that the card id is the package id minus
+    # the last .fr
+    with catalog:
```
bochecha: Hmm, so you use the catalog fixture as a context manager only when you want to list the installed packages, but not when you create new ones?

Or rather, you use it as a context manager only when you want it to mock the real underlying catalog API, but not when you call a function of the mocker itself.

Seems error prone, but maybe it's just me?

ideascube/tests/test_views.py
```diff
@@ -511,3 +517,55 @@ def test_by_tag_page_is_paginated(app, monkeypatch):
...
+
+
+def test_build_package_card_info_must_no_fail_for_no_package(app, systemuser):
+    from ideascube.configuration import set_config
+    from ideascube.views import build_package_card_info
+    set_config('home-page', 'displayed-package-ids', ['test.package1'], systemuser)
+    assert build_package_card_info() == []
+
+
+def test_build_package_card_info(app, systemuser, catalog):
```
bochecha: This test and the previous one don't seem like view tests.

Or maybe it is the `build_package_card_info` function which isn't specific to the views and should thus be placed elsewhere?

Or maybe I'm just talking nonsense and you should ignore me, not sure here. :stuck_out_tongue:

bochecha:
Third commit message:

> Some package instal media

I think you mean "install".

> The source attribute is only use

"used".

ideascube/mediacenter/tests/test_views.py
```diff
@@ -10,7 +10,32 @@
 from .factories import DocumentFactory
 from ideascube.search.models import Search
 
-pytestmark = pytest.mark.django_db
+# Almost every call to the mediacenter view need a catalog.
```
bochecha: "almost"

Just worried it might have an impact on the execution time of the tests... but then we're doing much worse with the various storage directories so maybe I shouldn't worry about this too much.

ideascube/mediacenter/tests/test_views.py
```diff
@@ -10,7 +10,32 @@
...
+# Provide a default mocked one for all tests.
+@pytest.yield_fixture()
+def default_catalog(catalog):
+    catalog.create_package(id = 'test.package1',
```
bochecha: space around the `=` operator

conftest.py
mgautierfr: In general, I google the error message.
But it's not opposite to a more explicit message. I will add it.

conftest.py
mgautierfr: It just mock the catalog (and only the list_installed method).
It's not creating any real package or change the db so no need for other fixture.

ideascube/mediacenter/tests/test_views.py
mgautierfr: The "almost" is not true anymore now we do not display package_id filter in the views.
So, indeed, no need to mock the catalog all the time.

I will change this.

ideascube/serveradmin/catalog.py
mgautierfr: No reason, I will change this.

ideascube/serveradmin/catalog.py
mgautierfr: We could simplify a bit yes (by rename package to w2eu.<lang>)
But for maps, we use a single template for several packages and it seems wrong to me to rename all map packages to maps.<extension>

We agreed that template_id should be determined from the package kind. We are not doing it right now cause we still have few things to define (icons, ...).
Maybe we could delay the special cases fix to latter when we will fix the icon/theme/.. stuffs.

ideascube/serveradmin/catalog.py
mgautierfr: Only if the arguments are changed in the function. Else it is useless.

And I thing that the `to_remove_ids = to_remove_ids or []` pattern is not so easy to read so I prefer to not use it if not necessary.

mgautierfr: yes.

ideascube/templates/ideascube/includes/cards/deus-ex-silicium.html
mgautierfr: Yes, but I will not do it :)
Changes about the theme_class and the package+card management update the same lines in a lot of files and I will not go through all those files and rewrite a patch only for a better commit splitting. (Plus the _kiwix.html will change again in a next future to just be kiwix.html)

ideascube/tests/test_views.py
mgautierfr: I was surprised that my tests on the home page return a 302.
Then I found that if there is no staffuser the home page redirect and that is tested.
But no test that everything work correctly if there is a staffuser.
So, I add it, but yes, I could create another pr for this.

ideascube/tests/test_views.py
mgautierfr: yes

ideascube/tests/test_views.py
mgautierfr: `build_package_card_info` is a internal function only used by the views.

I don't know where to put the function elsewhere to views.py and so, I test it in test_views.py.
If you have an idea, I take it :smile:

ideascube/tests/test_views.py
mgautierfr: The `create_package` is to configure the catalog mocker. It's not important if you configure the mock inside the context or not.

You just have to add the package to the mock before your code **use** the (mocked) catalog.
And the context manager is just here to mock the real underlying catalog API.

ideascube/serveradmin/catalog.py
mgautierfr: Hum, yes, and it simplify the code :)

ideascube/mediacenter/views.py
```diff
@@ -50,6 +59,19 @@ def get_context_data(self, **kwargs):
         self._set_available_kinds(context)
         self._set_available_langs(context)
         self._set_available_tags(context)
+        # We assume that if there is a source, it is a package_id.
```
bochecha: Currently that's the only thing we use the `source` for, but is that going to be a safe assumption going forward?

Are we already planning to use the `source` for something else?

ideascube/mediacenter/views.py
```diff
@@ -50,6 +59,19 @@ def get_context_data(self, **kwargs):
...
+        package_id = context.get('source')
+        if package_id:
+            catalog = catalog_mod.Catalog()
+            try:
+                package = catalog.list_installed([package_id])[0]
+            except IndexError:
+                # We canno't find the package.
```
bochecha: cannot

ideascube/mediacenter/views.py
```diff
@@ -50,6 +59,19 @@ def get_context_data(self, **kwargs):
         self._set_available_kinds(context)
         self._set_available_langs(context)
         self._set_available_tags(context)
+        # We assume that if there is a source, it is a package_id.
+        package_id = context.get('source')
+        if package_id:
+            catalog = catalog_mod.Catalog()
+            try:
+                package = catalog.list_installed([package_id])[0]
```
bochecha: Should we have a `catalog.get_package(package_id)` method instead?

ideascube/mixins.py
```diff
@@ -72,10 +74,10 @@ def _set_available_tags(self, context):
 
     def get_context_data(self, **kwargs):
         context = super(FilterableViewMixin, self).get_context_data(**kwargs)
-        for key in ('q', 'kind', 'lang'):
+        for key in ('q', 'kind', 'lang', 'source'):
             context[key] = self.request.GET.get(key)
         context['tags'] = self.request.GET.getlist('tags')
-        current_filters = [(k, context[k]) for k in ('kind', 'lang')
+        current_filters = [(k, context[k]) for k in ('kind', 'lang', 'source')
```
bochecha: Can't we do that at the same time we set `context[key]` just above?

ideascube/serveradmin/templates/serveradmin/home_page.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'serveradmin/index.html' %}
+{% load i18n %}
+
+{% block twothird %}
+<h2>{% trans "Manage displayed cards" %}</h2>
```
bochecha: Didn't we say « manage home page » ?

ideascube/serveradmin/templates/serveradmin/home_page.html
```diff
@@ -0,0 +1,20 @@
...
+<form name="cards" method="post" enctype="multipart/form-data" id="cards">
```
bochecha: Any reason you're explictly specifying the `enctype` attribute?

ideascube/serveradmin/tests/test_views.py
```diff
@@ -642,3 +642,46 @@ def test_staff_forgets_non_existing_connection(mocker, staffapp):
...
+    with catalog:
+        res = staffapp.get(reverse("server:home_page"), status=200)
+        assert 'Test package1' in res.unicode_body
+        assert 'Test package2' in res.unicode_body
+        assert 'Test package3' in res.unicode_body
+
+        form = res.forms['cards']
+        inputs = form.fields['displayed']
+        for input_ in inputs:
+            should_be_checked = (input_._value == 'test.package1')
```
bochecha: I'm going to insist here, can you add a comment explaining why you're accessing a private attribute, linking to the bug report? https://github.com/Pylons/webtest/issues/165

ideascube/serveradmin/tests/test_views.py
```diff
@@ -642,3 +642,46 @@ def test_staff_forgets_non_existing_connection(mocker, staffapp):
...
+
+    with catalog:
+        res = staffapp.get(reverse("server:home_page"), status=200)
+        assert 'Test package1' in res.unicode_body
+        assert 'Test package2' in res.unicode_body
+        assert 'Test package3' in res.unicode_body
+
+        form = res.forms['cards']
+        inputs = form.fields['displayed']
+        for input_ in inputs:
```
bochecha: What if `inputs` is empty? Can you ensure there are exactly 3 inputs?

ideascube/serveradmin/tests/test_views.py
bochecha: What if no input has its value set to `test.package1`? Can you ensure there is exactly 1 which does?

ideascube/serveradmin/tests/test_views.py
```diff
@@ -642,3 +642,46 @@ def test_staff_forgets_non_existing_connection(mocker, staffapp):
...
+            should_be_checked = (input_._value == 'test.package1')
+            assert (should_be_checked == input_.checked)
```
bochecha: I find the `for` loop a poor choice here.

You know you have exactly 3 inputs (which is very few) and you know exactly in which order they are.

Why not simply the following?

```python
assert inputs[0]._value == 'test.package1'
assert inputs[0].checked

assert inputs[1]._value == 'test.package2'
assert not inputs[1].checked

... same for the third one
```

The above has the advantage of verifying the same things you are, plus the things that I mentioned you aren't.

Doing these with the for loop would be more convoluted.

ideascube/serveradmin/catalog.py
bochecha: > Only if the arguments are changed in the function. Else it is useless.

It's also a big trap for the next person coming by.

Suppose we evolve this function and start modifying the argument? Then we'll need to remember to also change the default value... except we will forget, and the reviewer won't see it because the diff context won't show that line.

Using `[]` as a default kwargs value is just too risky.

As for `to_remove_ids = to_remove_ids or []` I don't like it any more than you, which is why I pretty much never use it. :smiley: 

I much prefer the explicit versions:

```
if to_remove_ids is None:
    to_remove_ids = []

to_remove_ids = to_remove_ids if to_remove_ids is not None else []
```

conftest.py
bochecha: True.

ideascube/serveradmin/catalog.py
bochecha: Maybe, yeah.

ideascube/templates/ideascube/includes/cards/deus-ex-silicium.html
bochecha: > I will not go through all those files and rewrite a patch only for a better commit splitting.

That's a shame, the story we tell with our RCS history is extremely important. :disappointed:

ideascube/tests/test_views.py
bochecha: Not necessarily another PR, but at least another commit.

ideascube/templates/ideascube/includes/cards/deus-ex-silicium.html
mgautierfr: I agree that it would have been better to have it in a separate commit, and I'm sorry you spend more time on this than necessary

As you said, the two potential commits would change the same lines. It not as just split different chunks in different commits but rewrite all lines by hand.
If we consider that these files will be simply suppressed in a next future and that you already understand these changes, it is a lot of work for a very small improvement.

ideascube/mixins.py
mgautierfr: We do not set current_filters for `q`. It would complicate the for loop with a inner if.

ideascube/serveradmin/tests/test_views.py
mgautierfr: You're right, I will rewrite it without loops.

ideascube/tests/test_views.py
mgautierfr: I will split this one.

ideascube/mediacenter/views.py
bochecha: > But, do we also need get_installed_package, get_available_package, get_upgradable_package and get_nothandled_package ?

We currently don't need all of these, only the `get_installed_package` version.

But yeah, that's just a minor cosmetic nit, if we end up finding we need it more then maybe we can add it at that point.

ideascube/mixins.py
bochecha: > It would complicate the for loop with a inner if.

But it would mean less looping, and thus less work for the poor Koombooks.

Anyway, not something for this PR, but in general I find we do a lot of unnecessary looping in many places of the code, and it might pay to try and make an effort at some point in the future on reducing that.

bochecha:
At of this writing, on master:

```
$ flake8 . | wc -l
253
```

Also as of this writing, on the package_card_auto branch:

```
$ flake8 . | wc -l
299
```

I'm certainly not saying we should go and fix all the ones in master (this has a real cost on following the VCS history), but should we try not adding new ones? Should we make this a condition for merge, like the tests passing?

ideascube/serveradmin/catalog.py
```diff
@@ -604,9 +606,19 @@ def list_nothandled(self, ids):
 
         return sorted(pkgs, key=attrgetter('id'))
 
+    def _update_displayed_packages_on_home(self, *, to_remove_ids=None, to_add_ids=None):
```
mgautierfr: To force later arguments to be keyword arguments only.

ideascube/serveradmin/catalog.py
```diff
@@ -604,9 +606,19 @@ def list_nothandled(self, ids):
...
+        displayed_packages = get_config('home-page', 'displayed-package-ids')
+        if to_remove_ids:
+            displayed_packages = [id for id in displayed_packages if id in to_remove_ids]
```
mgautierfr: Whaouh !!
Definitely. And it's missing a test also !

conftest.py
```diff
@@ -1,11 +1,15 @@
...
 import os
+import mock
 
 from django.core.urlresolvers import reverse
 
 from ideascube.tests.factories import UserFactory
 from ideascube.search.utils import create_index_table
+from ideascube.serveradmin.catalog import Package
+
+from contextlib import contextmanager
```
mgautierfr: yes, relic of the previous version of this patch.

ideascube/mediacenter/tests/test_views.py
```diff
@@ -10,7 +10,8 @@
 from .factories import DocumentFactory
 from ideascube.search.models import Search
 
-pytestmark = pytest.mark.django_db
+
+pytestmark = [pytest.mark.django_db]
```
mgautierfr: Relic of the previous version of the patch (where the pytestmark was a list of several marks)

I'll remove this chunk.

ideascube/mixins.py
mgautierfr: Agree with your general argument to try to remove unnecessary loops.

In this particular case, we are looping over 3 items in a list comprehension without any complex call (sql request, ...). I have no idea about how it competes with a extra `if` in a classical loop but I pretty sure the difference is negligible :)

ideascube/tests/test_views.py
mgautierfr: > Does it even have to be a context manager? Couldn't it just be a regular fixture? Couldn't you enter/exit the context manager directly in the catalog fixture?

Yes, this is a far more simple API.

ideascube/serveradmin/catalog.py
bochecha: I know what using `*` in a function signature does. My question was **why** you would use it here.

Only local stuff should use this method, so when calling it we'd have the source to see what the parameters should be. (and it is private, so it's not API, so it's fine to have to read the code to understand it IMHO)

Plus, I know it's possible to call a function specifying the kwargs as positional arguments, but that's just terribly bad taste and should always be avoided. I know I wouldn't accept merging something like this in a review:

```
def foo(bar=None):
    ...

foo('value')
```

(aside: I've always thought that Python itself should enforce passing kwargs as kwargs, that the snippet I wrote above should raise an error)

So all in all, I don't really see the point of the `*`, and I think it has too much potential for confusion. It's just too easy to mistake it for a `_` and think the function takes a positional argument that it doesn't use. It is also so rarely used, and so hard to search for (try searching for "python *" on a search engine :stuck_out_tongue_winking_eye:) that it certainly would cause trouble to a new developer.

Basically what I'm saying is that I dislike this `*` construction and I think it shouldn't be used because the benefits are negligible compared to the benefits, as long as we are careful to always pass kwargs as kwargs and never as positional args.

But that's just my opinion, which is why I asked you **why** you used it, because it is very possible that I'm wrong here and I was interested in your opinion, in case you had arguments that could make me change my mind. :smiley:

mgautierfr: # General answers 

> Only local stuff should use this method, so when calling it we'd have the source to see what the parameters should be. (and it is private, so it's not API, so it's fine to have to read the code to understand it IMHO)

I disagree here. A function is made to encapsulate a implementation. If you have to read the source code to know how to use a function you failed somewhere. Whatever the function is public or private.
I don't say that you don't need to read the code to understand the function in this case (there is no docstring, ...) but I disagree with the argument "it's private, we don't care of the interface".

In `def foo(bar=None): ...` bar is not a keyword argument, it's a argument with a default value.

> It's just too easy to mistake it for a _ and think the function takes a positional argument that it doesn't use.

Then python will complain if you pass a positional argument.

>  It is also so rarely used, and so hard to search for (try searching for "python *" on a search engine :stuck_out_tongue_winking_eye:) that it certainly would cause trouble to a new developer.

Well, if nobody use it cause nobody use it, nobody will use it :wink: 

> Basically what I'm saying is that I dislike this * construction and I think it shouldn't be used because the benefits are negligible compared to the benefits, as long as we are careful to always pass kwargs as kwargs and never as positional args.

Using `*` is the way to define a argument as a kwarg and not as a positional arg.
In python2 there is no way to define a kwarg but use a `**kwargs` and look into the dict (and a a docstring to tell which kwarg you can pass)
The default value is not related to the type of the argument.

I prefer to force the use of kwargs when the arguments are optional and the user will use a combination of arguments depending of its use.
There is no arguments "stronger" than an other. If you use positional argument, you imply an order, you can omit the second while passing the first but not the contrary.

# Our case

`to_remove_ids` is not "superior" to `to_add_ids`. Both are at the same level and one can be use without the other (not with the default value for the other, **without the other**)

With `def _update_displayed_packages_on_home(self, to_remove_ids=None, to_add_ids=None)` you can write :
```python
self._update_displayed_packages_on_home(a_list)
self._update_displayed_packages_on_home(a_list, another_list)
self._update_displayed_packages_on_home(a_list, to_add_ids=another_list)
self._update_displayed_packages_on_home(to_remove_ids=a_list)
self._update_displayed_packages_on_home(to_add_ids=a_list)
self._update_displayed_packages_on_home(to_remove_ids=a_list, to_add_ids=another_list)
```

With `def _update_displayed_packages_on_home(self, *, to_remove_ids=None, to_add_ids=None)` you can write :
```python
self._update_displayed_packages_on_home(to_remove_ids=a_list)
self._update_displayed_packages_on_home(to_add_ids=a_list)
self._update_displayed_packages_on_home(to_remove_ids=a_list, to_add_ids=another_list)
```

I prefer the second version cause, as you mention, it enforces passing kwargs as kwargs (cause you specify that `to_remove_ids` and `to_add_ids` are kwargs)

If you don't what a user to use `foo('value')` cause you think he should use a kwarg, explicit it !
As someone said, "Explicit is better than implicit"  :stuck_out_tongue:

bochecha: > In def foo(bar=None): ... bar is not a keyword argument, it's a argument with a default value.

The difference is academic. In practice, there just isn't one.

> With def _update_displayed_packages_on_home(self, to_remove_ids=None, to_add_ids=None) you can write :
> 
```
self._update_displayed_packages_on_home(a_list)
self._update_displayed_packages_on_home(a_list, another_list)
```

And what I'm saying is that this is a terrible idea and I wouldn't accept code like that during a review.

Just because it is possible to do something doesn't mean it's a good idea.

> If you don't what a user to use foo('value') cause you think he should use a kwarg, explicit it !

And I've always considered that defining the signature of a function with kwargs means you should call it as kwargs. The order might change, new ones might be added, calling it as kwargs is the only safe and unconfusing way to call it.

The fact that Python accepts calling with positional arguments a function defined with kwargs is a deisgn mistake in the language IMHO. The behaviour with the `*` is what should be the default.

If you strongly feel we should use `*` to make it explicit, then I suggest we start using `*` everywhere and outright ban the alternative. Not to be done in this PR, but it's a discussion we should have for the future. :wink:

ideascube/serveradmin/catalog.py
```diff
@@ -604,9 +606,20 @@ def list_nothandled(self, ids):
 
         return sorted(pkgs, key=attrgetter('id'))
 
+    @staticmethod
+    def _update_displayed_packages_on_home(*, to_remove_ids=None, to_add_ids=None):
+        displayed_packages = get_config('home-page', 'displayed-package-ids')
+        if to_remove_ids:
+            displayed_packages = [id for id in displayed_packages if id not in to_remove_ids]
+        if to_add_ids:
+            displayed_packages.extend(id for id in to_add_ids if id not in displayed_packages)
```
bochecha: Wondering if this isn't a job for `set`s?

```python
        displayed_packages = set(get_config('home-page', 'displayed-package-ids')
        displayed_packages -= to_remove_ids
        displayed_packages |= to_add_ids
```

Hmm, maybe not.

bochecha:
Overall the first commit is OK.

I just don't like that `*`, for reasons I already explained.

I also dislike the use of `staticmethod` for the same reasons: it doesn't bring anything valuable, but is so rarely use that it forces you to pause and causes a context switch when reading code. I prefer trying to keep the code as obvious as possible and avoiding to cause pause in the reader.

But we're clearly not going to convince each other so whatever, I'm not going to insist on this any more.

Now to go and review the other commits...

ideascube/mediacenter/tests/test_views.py
```diff
@@ -223,30 +223,53 @@ def test_tags_link_should_update_querystring(app):
...
+    # We use the available_packages in the view to get the display name
+    # of package in the remove filter link.
+    # But to have some available_packages, we need content that match the search.
+    DocumentFactory(lang='en', title='foo', tags=['tag1', 'tag2'], kind='video', package_id='test.package1')
+
+    # Also add the test.package1 in the catalog.
+    catalog.add_mocked_package(Package('test.package1', {
+        'name':'Test package1',
+        'description':'Test package1 description',
+        'template_id':'template_name',
```
bochecha: Same here, why the `template_id`?

ideascube/mediacenter/tests/test_views.py
```diff
@@ -223,30 +223,53 @@ def test_tags_link_should_update_querystring(app):
...
     resp = app.get("{}{}".format(reverse('mediacenter:index'),
                                  links[0].attrib['href']),
                    status=200)
-    assert {'q': ['foo'], 'tags': ['tag1', 'tag2'], 'kind': ['video']} == \
-        resp.request.GET.dict_of_lists()
+    assert resp.request.GET.dict_of_lists() == {
+        'q': ['foo'],
+        'tags': ['tag1', 'tag2'],
+        'kind': ['video'],
+        'source': ['test.package1']}
```
bochecha: Isn't that just checking that Django did its job right? Shouldn't we just trust them to populate `request.GET` correctly?

ideascube/mediacenter/tests/test_views.py
```diff
@@ -264,16 +292,38 @@ def test_remove_filter_should_be_present(app):
...
                    status=200)
-    assert {'lang': ['en'], 'q': ['foo'], 'tags': ['tag1', 'tag2']} == \
-        resp.request.GET.dict_of_lists()
+    assert resp.request.GET.dict_of_lists() == {
+        'lang': ['en'],
+        'q': ['foo'],
+        'tags': ['tag1', 'tag2'],
+        'source': ['test.package1']}
+
+
```
bochecha: One line too many.

ideascube/search/templates/search/filter_cloud.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n ideascube_tags %}
 {% spaceless %}
-{% if q or kind or lang or tags %}
+{% if q or kind or lang or tags or source %}
```
bochecha: Don't you wish you could use `any()` in Django templates? :-°

ideascube/serveradmin/catalog.py
mgautierfr: Yes, but I want to preserve the order.
Not that it is really important cause the user cannot change this order for now.
But I would be surprising if cards' order changes when a package in installed or removed.

bochecha: > Yes, but I want to preserve the order.

Excellent point.

ideascube/serveradmin/catalog.py
```diff
@@ -344,10 +367,23 @@ class StaticSite(SimpleZipPackage):
...
     handler = Nginx
 
+    # [FIXME] This propertie looks like hacks.
+    # We should rewrite few templates to have just one able to handle all
+    # static sites.
+    @property
+    def template_id(self):
+         if self.id == 'w2eu':
+             return self.id
+         base_name, extension = self.id.rsplit('.', 1)
```
yohanboniface: We may have language specific maps soon (like `jordan.map.ar`, `jordan.map.en`…), even if this is a "temporary hack", I'd use something like `if ".map" in base_name` to handle this.

ideascube/serveradmin/views.py
```diff
@@ -188,3 +197,18 @@ def wifi_history(request):
...
         request, 'serveradmin/wifi_history.html',
         {'wifi_list': wifi_list.values()})
+
+
+@staff_member_required
+def home_page(request):
+    if request.POST:
+        displayed_packages = request.POST.getlist("displayed")
+        set_config('home-page', 'displayed-package-ids', displayed_packages, request.user)
+
```
yohanboniface: Let's also add a message like "Home packages updated successfully" or something?

ideascube/serveradmin/catalog.py
```diff
@@ -322,6 +325,26 @@ def remove(self, install_dir):
...
 
+    # [FIXME] Thoses two properties look like hacks.
+    # We may want to find a way to find those information in the package or
+    # catalog metadata.
+    # For now, use special cases.
+    @property
+    def theme(self):
+        base_name, _ = self.id.rsplit('.', 1)
+        if base_name in ("wikipedia", "wikivoyage", "vikidia"):
+            return "discover"
```
yohanboniface: We need to make the strings translatable here: calling gettext or gettext_lazy to make the string discoverable by gettext.

ideascube/conf/idb_grc_eleonas.py
```diff
@@ -26,25 +26,6 @@
         'id': 'mediacenter',
     },
     {
-        'id': 'wikipedia',
```
mgautierfr: @fheslouin said me on IRC that those contents not installed by package use the ".old" template.
But yes, it's better to synchronized with ansiblecube team before the merge.

ideascube/serveradmin/catalog.py
mgautierfr: The theme is used twice in the kiwix template :
- As the css class to use (so must not be translated)
- As a text. And it is translated in the template.

So no need for me to translate the theme here.

ideascube/serveradmin/catalog.py
mgautierfr: Agree for the `".map" in base_name`

ideascube/serveradmin/views.py
mgautierfr: Agree

ideascube/mediacenter/tests/test_views.py
bochecha: Then please remove it? :smiley:

ideascube/serveradmin/views.py
bochecha: Please don't use the word « packages » in the UI.

ideascube/serveradmin/views.py
```diff
@@ -188,3 +197,20 @@ def wifi_history(request):
...
         {'wifi_list': wifi_list.values()})
+
+
+@staff_member_required
+def home_page(request):
+    if request.POST:
+        displayed_packages = request.POST.getlist("displayed")
+        set_config('home-page', 'displayed-package-ids',
+                   displayed_packages, request.user)
+        messages.success(request, _('Displayed home cards updated successfully'))
```
bochecha: I'd just say « Home cards updated successfully ».

The ones on the home page are the ones that get displayed, whereas the list in the admin page represents all the available cards.

ideascube/serveradmin/catalog.py
yohanboniface: > As a text. And it is translated in the template.

The issue is that if you don't call `gettext('xxx')` somewhere, with `xxx` hardcode, not a variable I mean, then `xxx` will never be discovered by gettext, then never added in .po files.

ideascube/serveradmin/views.py
yohanboniface: I think `package` is fine in the admin actually.

PR #438, opened 2016-08-17
Configuration draft for a demo server hosted at BSF.

letompouce:


letompouce:
Merged the wrong way, but still, merged.

PR #439, opened 2016-08-17
Config file for KoomBook Tanzanie

fheslouin:
This new configuration file add 4 new cards:
- biologie-tout-compris
- bouquineux
- deus-ex-silicium
- e-penser

We might have an issue with the TED card as there might be to many zim file.
Can I call two time the same card, ex. "ted", or should I leave it like this with the 10 zim file ?


ideascube/templates/ideascube/includes/cards/bouquineux.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n %}
+<div class="card tinted bouquineux">                                                                                                                                                          
+  <h4 class="flow"><span class="theme learn">{% trans 'info' %}</span><span><a href="http://kiwix.koombook.lan/bouquineux_07_2013/">{% trans "Bouquineux" %}</a></span></h4>               
+  <p>{% trans "Bibliothèque de livres numériques" %}</p>
```
yohanboniface: We should either have the string in English or not translate it. Same for other strings around.

ideascube/conf/kb_tza_ambassadeDeFrance.py
```diff
@@ -0,0 +1,91 @@
...
+        'id': 'universcience',
+        'languages': ['fr']
+    },
+    {
+        'id': 'biologie-tout-compris',
+        'languages': ['fr']
+    },
+    {
+        'id': 'bouquineux',
+        'languages': ['fr']
```
yohanboniface: Given it has a custom template, no need for the languages key.

fheslouin:
Can we merge this one and release a package ?

PR #444, opened 2016-08-19
Custom domain for the demoserver.

letompouce:


ideascube/conf/idb_bsf_demoserver.py
```diff
@@ -2,6 +2,9 @@
 """Configuration for a demo server hosted here, at BSF Headquarters"""
 from .idb import *  # noqa
 
+ALLOWED_HOSTS = ['.demo.ideascube.ideas-box.fr', '.ideasbox.lan.', 'localhost']
+DOMAIN = 'demo.ideascube.ideas-box.fr'
```
yohanboniface: Why not demo.ideascube.org at the end?

letompouce: At this time the existing demo was at `ideascube.ideas-box.fr` so it was consistent to turn this non-working demo to an explanation of what ideascube is, and append a (working) `demo.` sub-sub-sub-sub-domain.

It turned out that OVH sucksalot and the sub-sub-sub-domain never propagated ; then while I left for holidays, @fheslouin was pragmatic enough to move this to `demo.ideascube.org`.

PR #449, opened 2016-08-25
New configuration app

### LINKED ISSUE #441: Move in-database configuration management to its own app
This is growing too much to be left in the core `ideascube` module.

bochecha:
Our in-database settings were starting to grow a bit much for keeping them in the core.

This pull request splits them out into their own app.

In addition, the API is overhauled, to make the whole thing more robust and simple to use:

* we can now list all known configuration options
* we now know the default value of a configuration, which could allow resetting it
* we now enforce the type of a configuration, both when storing and retrieving it

I'm renaming it as well, because Django has its concept of "settings" which are the ones based on python files, and I was starting to get confused between the two.

Finally, this also ports the currently used settings (well, one) to the new API, both in terms of code and actually stored values.

Fixes #441


ideascube/configuration/errors.py
```diff
@@ -0,0 +1,21 @@
+class InvalidConfigurationValue(Exception):
```
yohanboniface: nit: I usually use`exceptions.py` as module name instead, thougts?

bochecha: > thougts?

Absolutely none.

ideascube/configuration/models.py
```diff
@@ -0,0 +1,76 @@
...
+            raise NoSuchConfigurationKey(namespace, key)
+
+    @classmethod
+    def get(cls, namespace, key):
+        key_registry = cls._get_key_registry(namespace, key)
+        expected_type = key_registry['python_type']
+
+        config = cls.objects.get(namespace=namespace, key=key)
+
+        if not isinstance(config.value, expected_type):
```
yohanboniface: Should we raise? Or instead log something and return the default?
I think we should be defensive as much as possible: better a setting not taken into account than a broken server.

ideascube/configuration/registry.py
```diff
@@ -0,0 +1,20 @@
...
+REGISTRY = {
+    # 'namespace': {
+    #     'key': {
+    #         'python_type': str,
+    #         'default': 'Default value',
+    #     },
+    # },
+    'server': {
+        'site-name': {
+            'python_type': str,
```
yohanboniface: nit: `type` may be simpler and clear enough?

ideascube/library/views.py
```diff
@@ -134,7 +134,8 @@ class SpecimenCreate(CreateView):
     def get_initial(self):
         book = get_object_or_404(Book, pk=self.kwargs['book_pk'])
 
-        return {'book': book, 'location': get_server_name()}
+        return {
+            'book': book, 'location': Configuration('server', 'site-name')}
```
yohanboniface: nit: a shorted name will prevent us to deal with the 79 chars lines ;)
Cfg?

ideascube/configuration/registry.py
bochecha: Next pull request adds a « pretty type » which I have called `type`.

I wanted to be explicit that this one is the actual Python type.

Plus it's an implementation detail, not something anybody would ever type outside of the `Configuration.get` and `Configuration.set` methods.

ideascube/configuration/models.py
bochecha: I would agree with you on this point if it were possible to have an incorrect type stored in the DB. (i.e what we have currently in master)

But the `set` method enforces the same constraint.

The only way to have an invalid type stored into the db is if someone manually goes and modify the DB. And frankly, I don't think I want to defend against this.

With this PR we're providing a simple to use API that enforces type constraints, and with the next PR we'll provide a configuration tool that will be simple to use and enforce the same type constraints.

If an admin goes and modify the DB manually and break the server, which is harder to do than using the provided tools, then they seriously had it coming.

If someone looks at the sqlite db file like a binary blob and modifies random bits (changes 0's into 1's and the opposite) in it, then sqlite will just barf. That is because the correct thing to do is to use the sqlite API and it enforces certain constraints on how the 0's and 1's are layed out in the binary db file. Essentially, the sqlite contract is « use my API and I guarantee that you'll always get a correct DB you can interact with; if you mess with the DB file you're on your own, I don't want to have anything to do with that ».

I strongly feel that with the set of constraints we're introducing in this PR on the `Configuration` model, we should have a similar contract: « use our API and we guarantee that the types will always be correct; if you mess up with the configuration records in the DB you're on your own, we don't want to have anything to do with that ».

Now, we might need to introduce more constraints (and testing!) to be able to provide such a guarantee. But I **strongly** suggest that we try moving in the direction of such a contraints+guarantee contract, than trying to defend against things that just shouldn't happen in the first place.

ideascube/library/views.py
bochecha: I dislike abbreviations for API names.

Having to introduce a newline is the perfectly normal thing to do in this specific case IMHO.

And in general, I'd rather have a discussion about raising the limit to (for example) 120 before I'd consider writing poor API.

yohanboniface: `Cfg` is a very common way of writting `configuration`, just like we write `pkg` for example. But ok, that was just a suggestion.

ideascube/configuration/models.py
yohanboniface: > But I strongly suggest that we try moving in the direction of such a contraints+guarantee contract, than trying to defend against things that just shouldn't happen in the first place.

I think we can do both :)
In general we should avoid raising exception, if we are not catching them ourself.

ideascube/library/views.py
bochecha: And the class in the catalog API is called `Package`, not `Pkg`. ;)

I saif I dislike abbreviations for **API names**, I didn't say anything about abbreviations for variables, for example.

That means I would have absolutely no problem writing code like the following:

```python
config = Configuration.get('server', 'site-name')
cfg = Configuration.get('server', 'site-name')
c = Configuration.get('server', 'site-name')
```

However, I think abbreviations make for very poor API names.

--- 

So, in this specific instance, if the line break bothers you then I would fix it with an intermediary variable:

```python
        site_name = Configuration('server', 'site-name')
        return {'book': book, 'location': site_name}
```

Because I consider it a lesser evil than a poor API.

yohanboniface: No no, the line break is OK, I was trying to prevent having line breaks everywhere we need to use the configuration in the future, as one line break is ok, but if have too much we lose a bit of readability. But OK, let's proceed :)

ideascube/configuration/models.py
bochecha: Except (pun intended :stuck_out_tongue:) what I'm saying the exception just won't ever happen if we stick to using the API properly.

If the exception ever happens, then it is a very strong sign that somebody did something terribly wrong.

To reiterate, if someone modified the DB manually instead of using the API, then I seriously don't care, unless you can find me a valid use-case for why that would ever happen.

If however the code enforcing the constraints is broken, then... let's review it properly and write tons of tests to ensure that's not the case?

yohanboniface: If it's not supposed to happen, why checking and raising, then? :p

ideascube/configuration/errors.py
mgautierfr: errors.py is a functional name.
exceptions.py is more a implementation detail name.
But both are valid, I don't especially care.

However, I prefer to name exception themselves with "Error" in the name.
Python itself do the same, all exception for error contain Error if the name.
Other exceptions are not Error (StopIteration, KeyboardInterupt, GeneratorExit, ...)

bochecha: > Python itself do the same, all exception for error contain Error if the name.
> Other exceptions are not Error (StopIteration, KeyboardInterupt, GeneratorExit, ...)

So Python does it except when it doesn't? I don't understand.

ideascube/configuration/models.py
mgautierfr: Several things here:

- Indeed, if it is not suppose to happen, no need to check, but
- Early error detection is better that latter one
- I still disagree to return the default value if a value is present but with a wrong type. (Especially if the set should prevent a wrong type to be inserted in the base)

On top of that, I generally prefer to use a cast/check method instead of using isinstance.

    try:
        return expected_type(config.value)
    except:
        # Something get wrong during the cast
        raise InvalidStoredConfiguration(...)

For basic type, it is almost the same, it even allow to handle wrong type value stored (int(5) == int('5')). If what we care is the API, this is still valid (you will get a object of the type set in the REGISTERY)
Moreover, it allows the check of more complex type (List of string, python object build from dict).

ideascube/configuration/models.py
```diff
@@ -0,0 +1,76 @@
...
+
+        return config
+
+    @classmethod
+    def get_value(cls, namespace, key):
+        try:
+            return cls.get(namespace, key).value
+
+        except cls.DoesNotExist:
+            return REGISTRY[namespace][key]['default']
```
mgautierfr: The difference between get and get_value is not obvious from the name only.
One will need to read the code to get the behavior. 

Python dict.get return the default value if the key is not present. Here this is get_value who return the default. Maybe we may change the API somehow (the name, make one function private) ?
At least a docstring.

ideascube/library/views.py
mgautierfr: Yes, I disagree to change the name of a API to fulfill caller coding style.

User can still use : 
`from ideascube.configuration.models import Configuration as Cfg`

ideascube/configuration/models.py
bochecha: The thinking is that `Configuration.get` being a class method it returns a `Configuration` instance, whereas `get_value` is a useful shortcut to get only the value (which is an attribute of the instance) of that configuration.

The names seem expressive to me, but since I wrote the code I might not be the right judge.

However, both are useful in different contexts, so neither should be private.

I agree this is lacking documentation, but I've been kind of cooled down on writing docs since that time @yohanboniface asked me to remove some doc I had written, so now I wait for someone to ask for doc before writing it. :stuck_out_tongue_winking_eye: I'll add some here.

ideascube/library/views.py
mgautierfr: Configuration**.get**('server', 'site-name') ?

ideascube/configuration/models.py
bochecha: > I still disagree to return the default value if a value is present but with a wrong type.

Me too, but I gave up on that and the next push will do as @yohanboniface asked.

> it even allow to handle wrong type value stored (int(5) == int('5')).

And this would be a completely invalid value.

ideascube/library/views.py
```diff
@@ -13,10 +13,10 @@
 from django.views.generic import (CreateView, DeleteView, DetailView, FormView,
                                   ListView, UpdateView, View)
 
+from ideascube.configuration.models import Configuration
```
mgautierfr: Importing from configuration.**models** hints the user that Configuration is a model and will be use as it (query, create instance, ...). However, we do not use Configuration this way (even if in the implementation, it is a model)

Maybe use a wrapper or simply make the Configuration class directly importable from ideascube.configuration ?

bochecha: Hmmm...

Abstracting the underlying ORM/model API is a conscious design choice, because I want to deemphasize the usual query API in favour of the `get`/`set` custom API which handles the verifications.

Your suggestion goes even further that way. I like it.

ideascube/configuration/errors.py
mgautierfr: Python use exception to control a normal execution.
For example, the StopIteration to stop a for loop.

If the Exception is a real error, the exception is named with Error (IOError, ValueError, ...)

ideascube/configuration/models.py
mgautierfr: If get return a Configuration instance, how it could return the default value ? It is a plain python type, not a Configuration instance.

ideascube/configuration/errors.py
bochecha: In the last push to this branch I renamed the file `exceptions.py` and added `Error` to the end of the class names.

ideascube/configuration/models.py
bochecha: Last push doesn't raise any more here, it logs and returns the default value. :disappointed:

ideascube/configuration/registry.py
bochecha: I renamed it `type` in the latest push to this PR.

I'll call the future one `pretty_type` instead.

ideascube/library/views.py
bochecha: The API was overhauled as we discussed, and as part of that I fixed this issue in the latest push to this branch.

Note that `Configuration.get` was just as wrong, it should have been `Configuration.get_value`, which clearly shows that the API wasn't clear at all. :stuck_out_tongue:

ideascube/configuration/models.py
bochecha: The API was completely overhauled in the latest push to this branch, there is no more `get` method returning an instance, so the confusion should be gone.

ideascube/library/views.py
bochecha: The API was completely overhauled in the latest push to this branch, in big part following what you suggested here.

Thanks!

ideascube/configuration/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+
+def get_config(namespace, key):
+    """Get a configuration option
+
+    Args:
+        namespace (str): The configuration namespace.
+        key (str): The configuration key inside the namespace.
+
+    Returns:
+        The current value (str) for this option.
```
mgautierfr: The type depends on what is in the REGISTRY. It may not be a str.

bochecha: Good catch!

Fixed.

ideascube/configuration/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+
+    expected_type = get_expected_type(namespace, key)
+
+    if not isinstance(value, expected_type):
+        raise InvalidConfigurationValueError(
+            namespace, key, expected_type, type(value))
+
+    Configuration.objects.update_or_create(
+        namespace=namespace, key=key,
+        defaults={'value': value, 'actor': actor})
```
mgautierfr: Do we update the existing object ?
(Then we need to update the date)

Or we create a new one with a different date ?
(And get the more recent one)

ideascube/configuration/migrations/0002_settings_to_configuration.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+from ideascube.models import Setting
```
yohanboniface: I think we'll remove this model at some point (now?), so this import will break.
Not sure exactly what the proper way to deal with that situation in Django, but I guess we should load the model from the registry, more or less like in https://docs.djangoproject.com/en/1.10/ref/migration-operations/#runpython ?
Also, there are very few servers with the Setting model being used, and it's only for the server name, so it may be just a good idea to just delete the model and put back the server name by hand when we update.

bochecha: > I think we'll remove this model at some point (now?), so this import will break.

I don't think we should remove it now, it's what enables the migration.

We can remove it at some point in the future, sure, but... is there any rush?

The fact there are very few doesn't mean we shouldn't try and do the right thing. It's done, it works, it didn't take much time developing, and it would take more time removing it.

When we do remove the model in the future, we can just guard the `import` in a `try`/`except ImportError` block, so that the migration doesn't do anything when the model doesn't exist any more.

yohanboniface: > I don't think we should remove it now, it's what enables the migration.

By "now", I meant "after the data is migrated" ;)

bochecha: > By "now", I meant "after the data is migrated" ;)

But we don't know when it will be migrated on all servers, since we don't really have an upgrade policy (yet).

What I'm saying is that any talk of removing that model and what will happen when it is removed is premature at this point. I'm honestly seeing us keep it for the next few years, and I don't think it would be any problem at all.

yohanboniface: If we want to do the data migration, I think the right thing to do is import the model from the versioned registry and then add migration to delete it.

yohanboniface: Crossed answers: I think adding the new confing system AND having a data migration AND having a migration to delete the Setting model is the right thing to do.

bochecha: How do you delete a model in a migration? A migration that modifies the code doesn't seem like something I'd be comfortable with. :confused:

yohanboniface: https://docs.djangoproject.com/en/1.10/ref/migration-operations/#deletemodel :)
The only thing is indeed to never import models directly in migrations: not only because they could have been deleted, but also because they may evolve and then do not fit anymore in the migration being run.
I think it's better to always import models from the registry, which knows where in the history we are when applying this migration.

bochecha: Ok, so we just weren't talking about the same thing:

* I meant model as in the code that maps a db table into a Python object
* you meant the db table

Indeed, we certainly can remove the db table, and we can do it right away, in the same operation that migrates the existing `Setting`s into `Configuration`s. I'll add that.

> The only thing is indeed to never import models directly in migrations: not only because they could have been deleted

This is what confuses me.

If we remove the model from the code, then whether you import it from the registry or directly, the import will still fail, won't it?

yohanboniface: No, I meant both the table AND the code :)

> If we remove the model from the code, then whether you import it from the registry or directly, the import will still fail, won't it?

I don't think so.
Try removing the code, then run a makemigrations, and you should have a DeleteModel migration that removes it from the registry (and delete the table, and I think remove the ContentType). But this model still exists in the registry for the steps *before* this given migration AFAIK.
It should be just straightforward.

yohanboniface: According to [this thread](http://stackoverflow.com/a/29046154) the ContentType is not removed by the migration actually. We should make sure we run migrate with `--no-input` btw.

bochecha: We do run with `--noinput` in package updates.

I've removed the model (with proper migration) in the latest push to this branch.

PR #451, opened 2016-08-26
Fix few catalog errors

mgautierfr:


ideascube/serveradmin/tests/test_catalog_command.py
```diff
@@ -93,11 +93,16 @@ def test_cannot_add_duplicate_remote(tmpdir, settings, monkeypatch):
 
     old_mtime = remotes_dir.join('foo.yml').mtime()
 
+    _, _ = capsys.readouterr()
```
bochecha: maybe just drop the `_, _` ?

ideascube/serveradmin/tests/test_catalog_command.py
```diff
@@ -93,11 +93,16 @@ def test_cannot_add_duplicate_remote(tmpdir, settings, monkeypatch):
...
+
     # Adding the same remote with the same url should not fail.
     call_command(
         'catalog', 'remotes', 'add', remote['id'], remote['name'],
         remote['url'])
 
+    out, err = capsys.readouterr()
```
bochecha: Either test `err`, or name it `_`?

ideascube/serveradmin/tests/test_catalog_command.py
mgautierfr: Yes, fix in last push --force.

ideascube/serveradmin/tests/test_catalog_command.py
mgautierfr: Remamed to _ in last push --force

ideascube/serveradmin/management/commands/catalog.py
```diff
@@ -224,7 +224,7 @@ def add_remote(self, options):
                 raise CommandError(
                     ('There already is a "{0.id}" remote and urls differ '
                      '({0.url} and {1})').format(e.remote, options['url']))
-            print('Not adding already existing remote: "{}"'.format(id))
+            print('Not adding already existing remote: "{}"'.format(options['id']))
```
bochecha: Wow, great catch!

Did we never try running this? o_O

ideascube/serveradmin/catalog.py
```diff
@@ -762,21 +763,26 @@ def update_cache(self):
...
 
-            # TODO: Verify the download with sha256sum? Crypto signature?
-            urlretrieve(remote.url, tmppath, reporthook=_progress)
-
-            catalog = load_from_file(tmppath)
-
-            # TODO: Handle content which was removed from the remote source
-            self._available.update(catalog['all'])
-
-            os.unlink(tmppath)
```
bochecha: Doesn't that cause the temporary files to accumulate on the filesystem?

ideascube/serveradmin/catalog.py
```diff
@@ -762,21 +763,26 @@ def update_cache(self):
...
             # TODO: Get resumable.urlretrieve to accept a file-like object?
-            fd, tmppath = tempfile.mkstemp()
-            os.close(fd)
+            with tempfile.NamedTemporaryFile() as fd:
+                tmppath = fd.name
 
-            def _progress(*args):
-                self._progress(' {}'.format(remote.name), *args)
+                def _progress(*args):
+                    self._progress(' {}'.format(remote.name), *args)
```
bochecha: Maybe keep the definition outside of the `with` block?

Less lines changed, less stuff to review. :stuck_out_tongue:

ideascube/serveradmin/catalog.py
```diff
@@ -762,21 +763,26 @@ def update_cache(self):
...
+                    print("Warning: Impossible to connect to the remote"
+                          " {remote.name}({remote.url}).\n"
+                          "Continue anyway without this remote."
+                          .format(remote=remote))
+                    continue
+
+                catalog = load_from_file(tmppath)
+
+                # TODO: Handle content which was removed from the remote source
+                self._available.update(catalog['all'])
```
bochecha: This whole block is harder to review than it should be, because you didn't just catch the `ConnectionError` as the commit message says, you also indented everything by using the `NamedTemporaryFile` as a context manager.

The fact you did both at once is what confused me into thinking the temporary files might never get removed, I think.

Splitting it into two commits would have been nice. :wink:

ideascube/serveradmin/catalog.py
mgautierfr: NamedTemporaryFile use as a contextManager automatically unlink the tempfile.

bochecha: Hmm, seems not since the `NamedTemporaryFile` gets destroyed as soon as it's closed.

bochecha:
Looks good to me.

Again, I would have preferred you split the two unrelated changes (using `NamedTemporaryFile` as a context manager and not failing when a remote can't be contacted) into 2 commits, but I won't block the review on this.

ideascube/serveradmin/catalog.py
mgautierfr: From doc:
> If delete is true (the default), the file is deleted as soon as it is closed.

=> The file is unlink when the fd is closed

With the `with ... as fd:`, we close the fd when we exit the context

So, we unlink the file when we exit the context.

bochecha: Yup, I know, that's what I said in my previous comment.

bochecha:
Why did I not merge this yet? o_O

PR #452, opened 2016-08-26
Correctly catch output from logging to stderr.

mgautierfr:
The default Handler from logging store "in cache" the sys.stderr when it's
create.
When capsys from py.test try to mock sys.stderr, it is to late.

But logging has a Handler who always print to sys.stderr and do not cache
it.

We configure logging to use it, then, mocking from py.test works correctly.


bochecha:
Confirming this fixes the issue I had in the configuration-app branch.

Only thing I'd like is a link to the pull request you've opened to fix this directly in pytest. :stuck_out_tongue_winking_eye:

PR #454, opened 2016-08-30
Remove language skills from default USER_FORM_FIELDS

yohanboniface:
IDB are overwritting this config anyway, and we don't want them
for Koombooks.


bochecha:
Fixed the tests for you. :stuck_out_tongue:

bochecha:
I'm going to trust you on the fact that we do want this, as I have absolutely no idea.

However, this seems like the obviously right way to do it. :+1:

PR #455, opened 2016-08-30
Move back to the latest upstream release of resumable-urlretrieve

bochecha:
We don't need to point ot our fork, now.

The reason we needed it went away as upstream merged out pull request:

berdario/resumable-urlretrieve#1


mgautierfr:
Nothing to say.

PR #460, opened 2016-09-02
A CLI to manage the in-database configurations

### LINKED ISSUE #440: A command-line tool to manage in-database settings
That would be nice, to manage the boxes properly.

It would have several benefits:

* manage the settings from ansible
* introduce experimental settings/behaviours that we wouldn't expose in the UI at first, giving us some time to test them in the field, eventually exposing them in the admin UI later on

I'm thinking of something like the `gsettings` tool in GNOME:

```
$ ideascube settings get content.local-languages
['ar', 'en', 'es', 'fr']
$ ideascube settings set content.local-languages ['en', 'zh-hant']
```

I would think the desirable features would be:

* `settings get <namespace>.<key>` to get the value of a setting
* `settings set <namespace>.<key> <value>` to set the value of a setting
* `settings list` to get the list of setting namespaces
* `settings list <namespace>` to get the list of setting keys in a namespace
* `settings reset <namespace>.<key>` to reset a setting to its default value
* `settings help <namespace>.<key>` to obtain help on a setting (description, possible values, default value)

The first two wouldn't be too hard to implement right now.

However, the other ones are a bit more tricky, as we currently don't have any of the information they would need anywhere.

So we'd need some kind of a registry for the settings...

It's starting to look like [we should move the settings management to its own app](https://github.com/ideascube/ideascube/issues/441).

bochecha:
Fixes #440

### DIFF
--- a/ideascube/configuration/__init__.py
+++ b/ideascube/configuration/__init__.py
@@ -3,11 +3,15 @@ import logging
 
 __all__ = [
     'get_config',
+    'reset_config',
     'set_config',
 ]
 
 
 logger = logging.getLogger(__name__)
+IMPOSSIBLE_MESSAGE = (
+    'The stored value for %s is of type %s instead of %s. This should never '
+    'have happened.')
 
 
 def get_config(namespace, key):
@@ -40,13 +44,43 @@ def get_config(namespace, key):
         return default_value
 
     if not isinstance(config.value, expected_type):
-        logger.error(
-            'The stored value for %s is of type %s instead of %s. This should'
-            ' never have happened.'
-            % (config, type(config.value), expected_type))
+        logger.error(IMPOSSIBLE_MESSAGE % (
+            config, type(config.value), expected_type))
         return default_value
 
     return config.value
+
+
+def reset_config(namespace, key):
+    """Reset a configuration option to its default value
+
+    Args:
+        namespace (str): The configuration namespace.
+        key (str): The configuration key inside the namespace.
+
+    Raises:
+        ideascube.configuration.exceptions.NoSuchConfigurationNamespaceError:
+            The requested namespace does not exist.
+        ideascube.configuration.exceptions.NoSuchConfigurationKeyError:
+            The requested key does not exist in the requested namespace.
+    """
+    from .models import Configuration
+    from .registry import get_expected_type
+
+    expected_type = get_expected_type(namespace, key)
+
+    try:
+        config = Configuration.objects.get(namespace=namespace, key=key)
+
+    except Configuration.DoesNotExist:
+        # The configuration had never been modified, nothing to reset
+        return
+
+    if not isinstance(config.value, expected_type):
+        logger.error(
+            IMPOSSIBLE_MESSAGE % (config, type(config.value), expected_type))
+
+    config.delete()
 
 
 def set_config(namespace, key, value, actor):
--- a/ideascube/configuration/management/commands/config.py
+++ b/ideascube/configuration/management/commands/config.py
@@ -1,9 +1,12 @@
 import argparse
 
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
 
+from ideascube.configuration.exceptions import (
+    NoSuchConfigurationNamespaceError,
+    )
 from ideascube.configuration.registry import (
-    get_all_namespaces,
+    get_all_namespaces, get_namespaced_configs,
     )
 
 
@@ -17,6 +20,9 @@ class Command(BaseCommand):
 
         list = subs.add_parser(
             'list', help='List configuration namespaces and keys')
+        list.add_argument(
+            'namespace', nargs='?',
+            help='Only list configuration keys for this namespace')
         list.set_defaults(func=self.list_configs)
 
         self.parser = parser
@@ -29,5 +35,17 @@ class Command(BaseCommand):
         options['func'](options)
 
     def list_configs(self, options):
-        for namespace in get_all_namespaces():
-            print(namespace)
+        namespace = options['namespace']
+
+        if namespace is None:
+            for namespace in get_all_namespaces():
+                print(namespace)
+
+            return
+
+        try:
+            for key in get_namespaced_configs(namespace):
+                print("%s %s" % (namespace, key))
+
+        except NoSuchConfigurationNamespaceError as e:
+            raise CommandError(e)
--- a/ideascube/configuration/management/commands/config.py
+++ b/ideascube/configuration/management/commands/config.py
@@ -11,7 +11,8 @@ from ideascube.configuration.exceptions import (
     NoSuchConfigurationNamespaceError,
     )
 from ideascube.configuration.registry import (
-    get_all_namespaces, get_config_data, get_namespaced_configs,
+    get_all_namespaces, get_config_data, get_default_value,
+    get_namespaced_configs,
     )
 
 
@@ -41,6 +42,10 @@ class Command(BaseCommand):
             'namespace', nargs='?',
             help='Only list configuration keys for this namespace')
         list.set_defaults(func=self.list_configs)
+
+        report = subs.add_parser(
+            'report', help='Print a full report of the current configuration.')
+        report.set_defaults(func=self.report_config)
 
         reset = subs.add_parser(
             'reset', help='Reset a configuration option to its default value')
@@ -126,6 +131,15 @@ class Command(BaseCommand):
 
         except NoSuchConfigurationNamespaceError as e:
             raise CommandError(e)
+
+    def report_config(self, _):
+        for namespace in get_all_namespaces():
+            for key in get_namespaced_configs(namespace):
+                value = get_config(namespace, key)
+                default = get_default_value(namespace, key)
+
+                print('%s %s: %r (default: %r)' % (
+                    namespace, key, value, default))
 
     def reset_config(self, options):
         namespace = options['namespace']
--- a/ideascube/configuration/management/commands/config.py
+++ b/ideascube/configuration/management/commands/config.py
@@ -4,7 +4,7 @@ import ast
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
 
-from ideascube.configuration import get_config, set_config
+from ideascube.configuration import get_config, reset_config, set_config
 from ideascube.configuration.exceptions import (
     InvalidConfigurationValueError,
     NoSuchConfigurationKeyError,
@@ -35,6 +35,12 @@ class Command(BaseCommand):
             'namespace', nargs='?',
             help='Only list configuration keys for this namespace')
         list.set_defaults(func=self.list_configs)
+
+        reset = subs.add_parser(
+            'reset', help='Reset a configuration option to its default value')
+        reset.add_argument('namespace', help='The configuration namespace')
+        reset.add_argument('key', help='The configuration key')
+        reset.set_defaults(func=self.reset_config)
 
         set = subs.add_parser(
             'set', help='Set the value of a configuration option')
@@ -94,6 +100,17 @@ class Command(BaseCommand):
                 print("%s %s" % (namespace, key))
 
         except NoSuchConfigurationNamespaceError as e:
+            raise CommandError(e)
+
+    def reset_config(self, options):
+        namespace = options['namespace']
+        key = options['key']
+
+        try:
+            reset_config(namespace, key)
+
+        except (NoSuchConfigurationNamespaceError,
+                NoSuchConfigurationKeyError) as e:
             raise CommandError(e)
 
     def set_config(self, options):
--- a/ideascube/configuration/management/commands/config.py
+++ b/ideascube/configuration/management/commands/config.py
@@ -1,0 +1,21 @@
+import argparse
+
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = 'Manage server configuration'
+
+    def add_arguments(self, parser):
+        subs = parser.add_subparsers(
+            title='Commands', dest='cmd', metavar='',
+            parser_class=argparse.ArgumentParser)
+
+        self.parser = parser
+
+    def handle(self, *_, **options):
+        if 'func' not in options:
+            self.parser.print_help()
+            self.parser.exit(1)
+
+        options['func'](options)
--- a/ideascube/configuration/management/commands/config.py
+++ b/ideascube/configuration/management/commands/config.py
@@ -11,7 +11,7 @@ from ideascube.configuration.exceptions import (
     NoSuchConfigurationNamespaceError,
     )
 from ideascube.configuration.registry import (
-    get_all_namespaces, get_namespaced_configs,
+    get_all_namespaces, get_config_data, get_namespaced_configs,
     )
 
 
@@ -28,6 +28,12 @@ class Command(BaseCommand):
         get.add_argument('namespace', help='The configuration namespace')
         get.add_argument('key', help='The configuration key')
         get.set_defaults(func=self.get_config)
+
+        describe = subs.add_parser(
+            'describe', help='Describe a configuration option')
+        describe.add_argument('namespace', help='The configuration namespace')
+        describe.add_argument('key', help='The configuration key')
+        describe.set_defaults(func=self.describe_config)
 
         list = subs.add_parser(
             'list', help='List configuration namespaces and keys')
@@ -85,6 +91,25 @@ class Command(BaseCommand):
             raise CommandError(e)
 
         print('%r' % value)
+
+    def describe_config(self, options):
+        data = {
+            'namespace': options['namespace'],
+            'key': options['key'],
+        }
+
+        try:
+            data.update(get_config_data(data['namespace'], data['key']))
+
+        except (NoSuchConfigurationNamespaceError,
+                NoSuchConfigurationKeyError) as e:
+            raise CommandError(e)
+
+        print(
+            '# {namespace} {key}\n\n'
+            '{summary}\n\n'
+            '* type: {pretty_type}\n'
+            '* default value: {default!r}'.format(**data))
 
     def list_configs(self, options):
         namespace = options['namespace']
--- a/ideascube/configuration/management/commands/config.py
+++ b/ideascube/configuration/management/commands/config.py
@@ -1,9 +1,12 @@
 import argparse
+import ast
 
+from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand, CommandError
 
-from ideascube.configuration import get_config
+from ideascube.configuration import get_config, set_config
 from ideascube.configuration.exceptions import (
+    InvalidConfigurationValueError,
     NoSuchConfigurationKeyError,
     NoSuchConfigurationNamespaceError,
     )
@@ -33,6 +36,13 @@ class Command(BaseCommand):
             help='Only list configuration keys for this namespace')
         list.set_defaults(func=self.list_configs)
 
+        set = subs.add_parser(
+            'set', help='Set the value of a configuration option')
+        set.add_argument('namespace', help='The configuration namespace')
+        set.add_argument('key', help='The configuration key')
+        set.add_argument('value', help='The new value')
+        set.set_defaults(func=self.set_config)
+
         self.parser = parser
 
     def handle(self, *_, **options):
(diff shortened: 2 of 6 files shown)


bochecha:
@fheslouin and @letompouce I'd really like you to give me your thoughts on the UX for this command, since you'll be the ones using it mostly.

Here are a couple of usage sessions:

* https://gist.github.com/bochecha/594bf0454473d20a6540aa2a8413b3ce
* https://gist.github.com/bochecha/29a13a7ab5db1ec5fc1e41caa6887c04

But don't hesitate to checkout this branch and try it out. :smiley: 

Also, I'm interested in how you'll use it in Ansible. For example, I didn't make it do anything interactively, because you won't be able to do that in ansiblecube. But I'm sure there are lots of other things I didn't think about that we should try and get right so that this CLI is as useful to you as it can.

ideascube/configuration/__init__.py
```diff
@@ -40,15 +44,46 @@ def get_config(namespace, key):
...
+        # The configuration had never been modified, nothing to reset
+        pass
+
+    else:
+        if not isinstance(config.value, expected_type):
+            logger.error(IMPOSSIBLE_MESSAGE % (
+                config, type(config.value), expected_type))
+
+        config.delete()
+
```
mgautierfr: I have been never fan of this syntax (try: ... except: ... else: ...)
The `else` always means 'in the other case'. So it means (to me) 'not the normal code path'
With the `try`, `else` means if no exception (so 'the normal code path')
I try to avoid if possible.

In this case for example, the `try` block is here to test if config has been set. At first sight, I tend to think that the else block is done if the config is not set. This is totally opposed of what we do as it is actually the normal behavior.

If we want to do early exit, I prefer this syntax :
```python
try:
    config = Configuration.objects.get(namespace=namespace, key=key)
except Configuration.DoesNotExist:
    # The configuration had never been modified, nothing to reset
    return

if not isinstance(config.value, expected_type):
    logger.error(IMPOSSIBLE_MESSAGE % (
        config, type(config.value), expected_type))

config.delete()
```

It has avantage to make the exit early pattern more visible and do not hide the normal code path behind a `else`
It also remove a level of indentation *

(*) authority argument : This is why Linus Torvald want 8 space indentation in the kernel : Avoid too many indentation.
https://www.kernel.org/doc/Documentation/CodingStyle

> Now, some people will claim that having 8-character indentations makes
the code move too far to the right, and makes it hard to read on a
80-character terminal screen.  The answer to that is that if you need
more than 3 levels of indentation, you're screwed anyway, and should fix
your program.

ideascube/configuration/management/commands/config.py
```diff
@@ -0,0 +1,153 @@
...
+    def handle(self, *_, **options):
+        if 'func' not in options:
+            self.parser.print_help()
+            self.parser.exit(1)
+
+        options['func'](options)
+
+    def _evaluate_value(self, value):
+        if value in ('true', 'false'):
+            # Special case booleans for a better UX
```
mgautierfr: Do we also handle 0/1, on/off, yes/no ?

ideascube/configuration/management/commands/config.py
```diff
@@ -0,0 +1,153 @@
...
+                NoSuchConfigurationKeyError) as e:
+            raise CommandError(e)
+
+        print(
+            '# %s %s\n\n'
+            '%s\n\n'
+            '* type: %s\n'
+            '* default value: %r' % (
+                namespace, key, config_data['summary'],
+                config_data['pretty_type'], config_data['default']))
```
mgautierfr: Maybe use the format method with named fields ?

ideascube/configuration/management/commands/config.py
```diff
@@ -0,0 +1,153 @@
...
+        key = options['key']
+
+        try:
+            value = get_config(namespace, key)
+
+        except (NoSuchConfigurationNamespaceError,
+                NoSuchConfigurationKeyError) as e:
+            raise CommandError(e)
+
+        print('%r' % value)
```
mgautierfr: Maybe find a way to tell the user if the config is not set and we are returning the default value ?

I not sure of how to do it.
A simple print could break (or complicate) using script.
Just print en stderr ?
Add another command or option to get_config ?

ideascube/configuration/registry.py
```diff
@@ -12,19 +12,29 @@
...
             'default': getattr(settings, 'IDEASCUBE_NAME', 'Ideas Cube'),
         },
     },
 }
 
 
+def get_all_namespaces():
+    for namespace in sorted(REGISTRY):
+        yield namespace
+
```
mgautierfr: yield from ?

ideascube/configuration/tests/test_command_config.py
```diff
@@ -0,0 +1,340 @@
...
+@pytest.mark.usefixtures('db')
+@pytest.mark.parametrize(
+    'value, expected, default',
+    [
+        (True, 'True', False),
+        (42, '42', 0),
+        ('A string', "'A string'", 'default'),
+        (['A', 'list'], "['A', 'list']", []),
+    ],
+    ids=[
```
mgautierfr: I didn't know the ids keyword argument. Thanks :)

ideascube/configuration/tests/test_command_config.py
```diff
@@ -0,0 +1,340 @@
...
+    monkeypatch.setattr('ideascube.configuration.registry.REGISTRY', {})
+
+    with pytest.raises(CommandError) as exc_info:
+        call_command('config', 'get', 'namespace1', 'key1')
+
+    assert (
+        'Unknown configuration namespace: "namespace1"') in str(exc_info.value)
+
+
+def test_get_no_config(monkeypatch):
```
mgautierfr: rename it to `test_get_no_key` ?

ideascube/configuration/tests/test_command_config.py
```diff
@@ -0,0 +1,340 @@
...
+    monkeypatch.setattr('ideascube.configuration.registry.REGISTRY', {})
+
+    with pytest.raises(CommandError) as exc_info:
+        call_command('config', 'help', 'namespace1', 'key1')
+
+    assert (
+        'Unknown configuration namespace: "namespace1"') in str(exc_info.value)
+
+
+def test_help_for_no_config(monkeypatch):
```
mgautierfr: same here `test_help_for_no_key` ?

ideascube/configuration/tests/test_command_config.py
```diff
@@ -0,0 +1,340 @@
...
+
+def test_list_namespaces(capsys, monkeypatch):
+    monkeypatch.setattr(
+        'ideascube.configuration.registry.REGISTRY',
+        {'namespace2': {}, 'namespace1': {}})
+
+    call_command('config', 'list')
+
+    out, err = capsys.readouterr()
+    assert out.strip().split('\n') == ['namespace1', 'namespace2']
```
mgautierfr: What happens if namespaces contain keys ?
Does the list also print the keys for all namespaces ?
Maybe add a key in one of the namespaces to explicit that keys are not listed ?

ideascube/configuration/tests/test_main_api.py
```diff
@@ -77,6 +77,59 @@ def test_get_default_configuration(monkeypatch):
...
+                'setting2': {'type': int},
+            },
+            'tests2': {
+                'setting3': {'type': list},
+            },
+        })
+
+    set_config('tests', 'setting1', 'A string', user)
+    set_config('tests', 'setting2', 42, user)
+    set_config('tests2', 'setting3', ['A', 'list'], user)
```
mgautierfr: Add a `assert Configuration.objects.count() == 3` before reset_config ?

ideascube/configuration/management/commands/config.py
bochecha: One thing to keep in mind is that this is a CLI, and as such all it can receive as input are strings.

We want to transform (`eval`uate) those strings into Python objects.

So if the admin uses the command `config set foo bar 1` we can't really know whether they meant the integer `1`, the string `'1'`, or the boolean `True`.

The way to deal with this ambiguity IMHO is to avoid it at all cost. That means assuming that `1` will always be an integer, period.

That's why to set a config to the string `'1'` you'd need to use the command as follows:

```
$ python manage.py config set server site-name "'1'"
```

The first line of quotes get completely swallowed by the shell (at least that's the case with bash) so if you used `'1'` or `"1"` the `config.py` code only gets `'1'` (i.e a one-character long string containing only the character `1`) which gets evaluated as the integer `1`. Passing the parameter as above means that `config.py` gets `"'1'"` (i.e a thee-character long string containing a single quote, the character `1` then another single quote), which gets evaluated as a string properly.

If we wanted to let the admin type `python manage.py config set server site-name 1` and magically evaluate that `1` as a string, we'd need to peek into the configuration registry to find the expected type, and then try to evaluate it into something of that type.

Sure, that's easy for the parameter `1`: we can evaluate it as a string simply with `str(value)`. Now if the expected type is a list and `config.py` receives the string `foo`, should we pass it through `list()` and thus store `['f', 'o', 'o']`? Of course we shouldn't, that would be a case of an admin trying to store an invalid value, and we should error out.

Similarly, if the expected type is a string and the admin called passed the value as `"['en', 'fr', 'zh-hant']"` (the syntax for passing a list), we also shouldn't pass it to `str()` and store that awful string in the database.

So we'd have special cases only for things that evaluate as integer, if the expected type is a string.

Or, as you suggested, if the passed value evaluates as an integer and the expected type is a boolean.

And then we quickly get into special-corner-cases-territory, and soon we'll start wondering where did we go wrong.

On top of that, I really don't think the CLI layer should go and peek inside the registry on its own, except for things like `list` and `help` which are purely about presenting information to the user (i.e read-only).

So no, I don't think we should handle 0/1 as booleans. (I know, I took a slight detour :laughing:  )

As for on/off and yes/no, I don't think they are needed, but I'll accept I might be wrong. That's also why I'd really like to have people who don't read the code to try it out, because it is something we would figure out instantly if they tried setting a boolean value with something else than true/false. I'm not going to mention them in this conversation because reading the thought process I detailed above (and in fact your very question) would influence how they try using it, and I'm really interested in their feedback as first-time users. :smiley: 

This begs the question as to why I added this special case, then. Well, without it the CLI would only accept the strings `True` and `False` as valid for booleans, and I felt that refusing "true" because it doesn't have the proper case would be incredibly sadistic.

ideascube/configuration/management/commands/config.py
bochecha: > Maybe find a way to tell the user if the config is not set and we are returning the default value ?

I don't think it matters one bit for the user.

You want to know what languages are currently allowed for content uploads on the box. Whether it's the default value or not is irrelevant, all you want to know is the actual value.

Also, I totally expect the `get` command to be used very rarely (whether interactively or in scripts). The workflows I expect are:

* ansiblecube will use `set`
* an admin will sometimes want to `reset` if a value is causing problems on a box
* an admin will want to `get` the current value, and if it's not what they expected then `set` it to something else (or just `set` directly without even looking at the current value, because it often doesn't matter what the current value is when you're about to change it anyway)
* to do the above, an admin will need to `list` because they keep forgetting how that key is called, and then `help` to figure out what to put after `set`

Those are the primary workflows I designed the CLI for. Knowing whether the current value is the default isn't one of them (and it's possible anyway with `get` followed by `help`) because I don't expect it to ever be important.

Of course, actual use will test those assumptions, and sometimes prove them wrong. Again, this is why I'm interested in uninformed feedback from the main stakeholders here. (and I still won't mention them here to avoid influencing their testing)

We can always get back on this after such testing/feedback happened.

As for your actual suggestions, printing on stdout would break scripts as you said, so would printing on stderr (plus it would be bad taste). The only workable solution IMHO would be a `--verbose` or `--details` option. (and if we went there, I would think we could merge the functionality of `help` into that option of `get`)

ideascube/configuration/registry.py
bochecha: I don't understand your question.

ideascube/configuration/tests/test_main_api.py
bochecha: That's already verified in the tests for `set_config`, so it's not necessary IMHO.

In any case, it doesn't hurt, so I'll add it.

ideascube/configuration/registry.py
mgautierfr: It was a suggestion to use `yield from sorted(REGISTRY)` instead of the `for` loop.

But it's just a suggestion. Few people use it.
The only "massive" use of `yield from` I know about was with asyncio before `await` came into python.

bochecha: > It was a suggestion to use yield from sorted(REGISTRY) instead of the for loop.

Wait, that works? :open_mouth: 

[goes to try it out...]

Wow! :open_mouth: 

> The only "massive" use of yield from I know about was with asyncio before await came into python.

Yeah, coroutine stuff (not necessarily async) is the only place I've ever seen `yield from` being used.

I'm going to say no here, due to how rarely this seems to be. Familiarity isn't a trump-all card, but in this case, I'm going to stand by it.

Thanks for teacking me something, though!

ideascube/configuration/__init__.py
bochecha: First of all, I dislike the argument of authority. It does a disservice to the rest of your argument.

As for the actual issue, I'm used to the `try/except/else` clause and quite like it. There are even cases where I would argue it is better and more understandable than the alternative.

But this is not such a case. This is a case where both possibilities are just as understandable as the other, the only difference being that your suggestion removes some indentation.

I do dislike unnecessary indentation, so I've changed the code to the pattern you suggested in the latest push to this branch.

ideascube/configuration/management/commands/config.py
bochecha: You're right, that's nicer. Somehow I'm so used to `%s`-like string interpolation that I always forget to use `format`.

Done in the latest push to this branch.

ideascube/configuration/tests/test_command_config.py
bochecha: I renamed it to `test_get_config_no_key` to be more consistent with the rest of the tests in the latest push to this branch.

ideascube/configuration/tests/test_command_config.py
bochecha: I renamed it to `test_help_for_config_no_key to be more consistent with the rest of the tests in the latest push to this branch.

ideascube/configuration/tests/test_command_config.py
bochecha: Good point. Done in the latest push to this branch.

ideascube/configuration/tests/test_main_api.py
bochecha: Done in the latest push to this branch.

bochecha:
I think I addressed all the comments so far. Please send more. :smile:

ideascube/configuration/management/commands/config.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if namespace is None:
+            for namespace in get_all_namespaces():
+                print(namespace)
+
+            return
+
+        try:
+            for key in get_namespaced_configs(namespace):
+                print("%s %s" % (namespace, key))
```
yohanboniface: Tempted to also display the current value and the default value. 
Something like:

    namespace | key         | value  | default
    server    | server-name | blabla | Ideas Cube

Thoughts?

ideascube/configuration/management/commands/config.py
```diff
@@ -0,0 +1,153 @@
...
+        print(
+            '# {namespace} {key}\n\n'
+            '{summary}\n\n'
+            '* type: {pretty_type}\n'
+            '* default value: {default!r}'.format(**data))
+
+    def list_configs(self, options):
+        namespace = options['namespace']
+
+        if namespace is None:
```
yohanboniface: I'd keep the `namespace` option, but when not given I'd return the whole list of all keys from all namespaces instead.

ideascube/configuration/registry.py
yohanboniface: I'm a user of `yielf from` too, so I'd have arbitrated in favor of it :)

ideascube/configuration/management/commands/config.py
bochecha: Same as before, that means you lose a way to just list keys for ansible/scripts.

Also, you lose a nice property of the current command:

```
$ config list server
server site-name
$ config get server site-name
'Foo Bar'
```

I tried to make it so that you could just copy-paste the return of most commands to use them directly in other commands. That's nice for admins, it's also great for scripts.

As for the previous comment, I'd like to not change the UX until we start using this thing, at which point we can challenge the design assumptions I made and revisit the UX.

yohanboniface: > As for the previous comment, I'd like to not change the UX until we start using this thing, at which point we can challenge the design assumptions I made and revisit the UX.

I was actually using the command, and I may be one of the actual main users of that script ;)

ideascube/configuration/registry.py
bochecha: You know what? This is the kind of syntactic sugar I normally enjoy, and the only thing I had against is lack of familiarity.

But I actually want to get familiar with this one. It's elegant, quick, and not entirely obscure.

So I changed it in the latest push to this branch.

ideascube/configuration/management/commands/config.py
bochecha: But you do see how your suggestion would make scripting super hard, because suddenly the script would have to scrape the output and parse it?

It also entirely replaces the `get` command, which suddenly becomes completely useless.

Also, if we just added the summary and type to your suggestion, then it would completely replace the `help` command (which I'm starting to think I should have called `describe`).

Basically, your suggestion implies completely rethinking all the design assumptions I made, so if you want to make changes, then instead of a local suggestion I'd like you to think about the big picture and actually design the whole thing.

I really think that before we start adding some bells and whistles to the command output, we need:

* someone to try and write the scripts that will use this command, as that requires a certain kind of simple output (no scraping/parsing, directly the things needed for the script to continue)
* usage in the field, to see the kind of output that would make more sense to a human and make their job easier

As far as I know, nobody started on the former, so we have no idea what works and what doesn't with the current scripting UX. (and I'd contend that the current input/output would actually work really well for scripts)

You've started on the latter, which is great, but I'd argue that this is too few to really start getting a good feeling of the pain points with the current UX. Can we use it in a few deployments first?

One thing that I think would make a much better UX is tab-completion:

```
$ ideascube config get <tab><tab>
content home server
$ ideascube config get s<tab>
$ ideascube config get server <tab><tab>
site-name
$ ideascube config get server s<tab>
$ ideascube config get server site-name
'Chez Oim'
```

Implementing the above would require a way to list namespaces (and only get namespaces), a way to get keys in a namespace (and only get that).

This is all things I've carefully thought about when designing the UX for this CLI, and so far the local suggestions you and @mgautierfr made would break one thing or another.

Now I'm certainly not saying the UX I designed is perfect, far from it. But it is a global, consistent design based on certain assumptions, and if we're going to change it, I'd like us to instead challenge these very assumptions with sufficient feedback from real-world usage, and then evolve a new global, consistent design.

yohanboniface: > Now I'm certainly not saying the UX I designed is perfect, far from it. But it is a global, consistent design based on certain assumptions, and if we're going to change it, I'd like us to instead challenge these very assumptions with sufficient feedback from real-world usage, and then evolve a new global, consistent design.

I understand that, and also that my suggestion is not fully aligned with the current design. I actually think the design is good.
But what I mean is that I've been trying to use the command, and the first thing I've wanted to see is "how is the config here overall" (discover the keys, see the values, etc.), and I felt not convenient to need to go step by step to have all the information.
So we may just add a `overview` (or `all` or…) command or an option to list or whatever. But I'm sure I'm needing this now, and I will in the future, and not only me. Because I really think from now on when connecting to a server for debugging that will be one of the first thing we'll do to have the big picture. And it's also something we may want to add on logs at some points.
I'm just giving a user feedback, and making a suggestion, but that suggestion can be discussed to end on something better at the end to have this "config overview".

bochecha: > the first thing I've wanted to see is "how is the config here overall"

Now this is a very interesting one, and I hadn't thought about it. Some kind of full report of the whole configuration of the box.

This is why I've been pushing back by the way: to get an actual use-case rather than merely a UI suggestion. :wink: 

I'm totally going to implement this today, because it's such a useful feature and I'm kind of sad I didn't think about it. Plus, it could probably be added to the CLI without interfering with any of the rest.

Thanks! :smiley:

bochecha:
I've just rebased the branch, and changed two things:

* I added a `config report` command, which prints a list of all the config options on the box, their current and default values;
* I renamed the `help` command into `describe`, which better... describes (haha) what the command actually does;

PR #462, opened 2016-09-03
Display pdf/audio preview if any in document page

yohanboniface:
Before
screenshot from 2016-09-03 21-00-01

After
screenshot from 2016-09-03 20-57-18


bochecha:
Looks like the obvious thing to do. :+1:

PR #464, opened 2016-09-04
Tag list

yohanboniface:
I'm often using this from the command line to help me clean the tags (like merging learn and learning).


ideascube/management/commands/tags.py
```diff
@@ -111,3 +115,10 @@ def replace(self, options):
             content.tags.add(new)
         old.delete()
         notice('Deleted "{}"'.format(old))
+
+    def list(self, options):
+        row = '{:<40}{:<40}{}'
+        print(row.format('name', 'slug', 'slug'))
```
mgautierfr: Twice slug ?

Maybe `print(row.format('name', 'slug', 'count'))` ?

yohanboniface: Erf, good catch :p

ideascube/management/commands/tags.py
```diff
@@ -50,6 +50,10 @@ def add_arguments(self, parser):
         replace.add_argument('new', help='New tag name.')
         replace.set_defaults(func=self.replace)
 
+        list_ = subs.add_parser('list',
+                                  help='List tags')
```
bochecha: pep8 would be sad.

yohanboniface: erf, I may had falafels instead of eyes when I coded this :p

bochecha:
I only have one thing slightly bothering me with this pull request: you move it from the mediacenter app to the core ideascube app.

Sure, it's not specific to the mediacenter, so it has no reason to be there.

At the same time, it's also not specific to ideascube.

What I'm saying is that maybe we should try and send that upstream to Taggit?

Other than that, it looks good now.

PR #465, opened 2016-09-05
Media queries reboot

### LINKED ISSUE #447: Cards : Text overlap the image
On some card, if the text is too long, it is overlapping the image

ideascube

yohanboniface:
This does three things:

- completely reboot media queries: they were done the wrong way (not using cascading, so rules were declared for every media query), some rules were obsolete, and some rules were just too specific; I'm not saying this is fixing every issue we can have with responsiveness, but it's much better on the tablets, and the code is much much simpler (200 lines vs 35), so easier to maintain and fix in the future
- add mhalf and thalf break point triggers (m as mobile, t as tablet, as for other rules created before): allow to better divide the screen on small sizes
- fix #447 by adding a generic padding on home cards (rtl support too)

### DIFF
--- a/ideascube/static/ideascube/main.css
+++ b/ideascube/static/ideascube/main.css
@@ -1107,10 +1107,16 @@ table.stock .actions {
     .twide {
         width: 100%;
     }
+    .thalf {
+        width: 50%;
+    }
 }
 @media screen and (max-width: 767px) {
     .mwide {
         width: 100%;
+    }
+    .mhalf {
+        width: 50%;
     }
     header {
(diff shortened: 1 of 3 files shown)


yohanboniface:
### Before on the media center:

screen shot 2016-09-05 at 10 30 44
screen shot 2016-09-05 at 10 30 33
----
### After on the media center:
screen shot 2016-09-05 at 10 29 22
screen shot 2016-09-05 at 10 30 56

bochecha:
Looks much better. Not just the site but also the CSS code. :+1:

PR #478, opened 2016-09-08
Dynamic js caching

### LINKED ISSUE #469: Version /jsi18n/ and /js/
Those two URLs are requested on every loaded page, but their content only change when we release a new version of ideascube, so the client may cache it.

I suggest to add the ideascube package version in the urls:

    /jsi18n/0.11.5/
    /en/js/0.11.5/

(This has to be done in `base.html`, for example using the `version` context processor.)

And adding very aggressive cache-control headers to those views.

bochecha:
Well, not just caching, see each commit for details.

Fixes #469

### DIFF
--- a/ideascube/views.py
+++ b/ideascube/views.py
@@ -263,10 +263,3 @@ class AjaxProxy(View):
             return HttpResponse(content, status=status_code,
                                 content_type=mimetype)
 ajax_proxy = AjaxProxy.as_view()
-
-
-def javascript(request):
-    response = render(request, 'ideascube/includes/js.html',
-                      {'settings': settings})
-    response['Content-Type'] = 'text/javascript'
-    return response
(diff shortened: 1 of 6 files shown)


ideascube/views.py
```diff
@@ -265,8 +267,6 @@ def get(self, *args, **kwargs):
 ajax_proxy = AjaxProxy.as_view()
 
 
-def javascript(request):
-    response = render(request, 'ideascube/includes/js.html',
-                      {'settings': settings})
-    response['Content-Type'] = 'text/javascript'
-    return response
+@cache_control(max_age=31536000)
```
yohanboniface: You should be able to use the decorator in the urls.py directly, something like:

    url(r'^jsi18n/$', cache_control(max_age=31536000)(javascript_catalog)),

bochecha: Huh, why did I not think of that?

Not only is it obvious, but there is another URL using a decorator from the same django module in exactly the same way a few lines above.

Thanks, fixed.

PR #479, opened 2016-09-09
Fix deprecation warnings

### LINKED ISSUE #401: Fix deprecation warnings
https://github.com/kelvinwong-ca/django-select-multiple-field/issues/12

Note that this is going to prevent us from moving to Django 1.10.

bochecha:
Fixes #401

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 batinfo==0.3
 Django==1.9.8
 django-countries==3.4.1
-django-select-multiple-field==0.4.2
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,7 +1,6 @@
 batinfo==0.3
 Django==1.9.8
```
yohanboniface: Shouldn't we also upgrade to Django 1.10 then?

bochecha: That's in #481 :wink:

PR #486, opened 2016-09-15
Medias improvements

### LINKED ISSUE #468: Client cache for medias

bochecha:
See each commit for the details.

Fixes #468

### DIFF
--- a/ideascube/mediacenter/forms.py
+++ b/ideascube/mediacenter/forms.py
@@ -4,7 +4,7 @@ from django.conf import settings
 from ideascube.widgets import LangSelect
 
 from .models import Document
-from .utils import guess_kind_from_content_type
+from .utils import guess_kind_from_content_type, guess_kind_from_filename
 
 import os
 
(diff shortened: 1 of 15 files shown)


ideascube/mediacenter/forms.py
```diff
@@ -19,21 +19,29 @@ class Meta:
...
+
         if commit:
-            if not document.id:
-                # m2m need the document to have an id.
-                document.save()
-            self.save_m2m()
-            document.save()  # Search index needs m2m to be saved.
+            # Need to save a second time to index the m2m
+            document.save()
+
```
mgautierfr: It seems we've lost the `self.save_m2m()` call

ideascube/templatetags/ideascube_tags.py
```diff
@@ -184,3 +184,11 @@ def remove_qs(context, **kwargs):
     params.update(MultiValueDict(existing))
     params.pop('page', None)  # Changing search context, reset page.
     return '?%s' % params.urlencode()
+
+
+@register.simple_tag
+def media(instance, attribute):
```
mgautierfr: The `media` name seems wrong here.
`media` is for what use the tag, not what the tag is doing.
Without the commit comment or the media source code, it is difficult to know what this tag is doing.

We could rename it something like `gen_timed_url` or `unique_url` (I'm open to better idea)

bochecha: The goal was to mirror the `static` tag. Eventually I would expect (and in fact hope) we don't know or care about what the tag does, we just get into the habit of always using for all media files, just like we use `static` for all static files.

This might not be a good idea of course, but this was my reasoning, and I still think it makes (some) sense. :smile:

ideascube/mediacenter/forms.py
bochecha: Nope, it's done by the `super` call, as it should always have been.

bochecha: In fact, we shouldn't even be subclassing this method at all, the parent should do the right thing... Except somehow we need to save the `document` **twice** (the form's `save` method already saves the `document) for the indexing to work properly.

We really should fix this so that saving the document once is enough for the indexing to work, and then we could completely drop this method.

Issue #456, opened 2016-08-31
Migrations broken on servers having run previous migrations after #429


yohanboniface:
http://paste.ubuntu.com/23116110/

The workaround is to run a reindex (that will crash but create the idx table anyway), than a migrate again.

We need a way to fix this for installed servers.

One option may be to have a pre-migrate that create the idx table.

yohanboniface:
To reproduce:
- remove any db in storage/
- git checkout 0.5.0
- python manage.py migrate
- python manage.py dummydata
- git checkout master
- python manage.py migrate

This should fail with the current HEAD, and we should find a way to fix this scenario.

bochecha:
I can't reproduce, because the `dummydata` step fails here... because the code at the time wasn't Python3 compatible :D

PR #489, opened 2016-09-16
Force the re-creation of the index table when we make migration.

mgautierfr:
The structure of the index table may change and the migrate command
should handle those changes.

As the index table is not managed, we cannot detect those changes and
we must recreate the table all the time to be sure we've got the last
structure version.


bochecha:
Please open your pull requests based on the latest master. :wink:

bochecha:
Just to be clear, this does not improve anything with regards to #456.

bochecha:
I think #490 is a better fix for your issue, as it also fixes #456.

PR #490, opened 2016-09-16
search: Recreate the index table before migrating

### LINKED ISSUE #456: Migrations broken on servers having run previous migrations after #429

bochecha:
We have two problems with the search app currently.

The first is that if an instance of ideascube gets updated from a very old version, the migrations fail. (#456)

The second is that because the Search model is not managed by Django, if we add new fields to it then no migration gets created, which breaks the application as well as the post-migrate reindex. (#489)

The solution to both is actually quite simple: simply always drop and recreate the index table before running the migrations.

Fixes #456
Fixes #489

### DIFF
--- a/ideascube/search/apps.py
+++ b/ideascube/search/apps.py
@@ -1,7 +1,12 @@
 from django.apps import AppConfig
-from django.db.models.signals import post_migrate
+from django.db.models.signals import pre_migrate, post_migrate
 
-from .utils import reindex_content
+from .utils import create_index_table, reindex_content
+
+
+def create_index(sender, **kwargs):
+    if isinstance(sender, SearchConfig):
+        create_index_table(force=True)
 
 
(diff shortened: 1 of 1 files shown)


yohanboniface:
Merging, but I've a concern: migrate is run at every ideascube upgrade on the servers.
reindex is taking some time on the koombooks.
I wonder if running a a reindex each time we upgrade is the best path to go.
As far as I remember, the initial plan was to add a migration with a reindex step only when needed.
Let's keep an eye on this in the future: if we have any complaint about install being longer and longer; let's see how often we would actually need a full reindex, etc.

PR #498, opened 2016-09-23
Don't rely on an unguaranteed order

bochecha:
Django offers no guarantee of the order in which the fields of a form
are cleaned.

As a result, we can't have a clean_foo method relying on the fact that
another field bar has already been cleaned.

For this kind of cases, we must use the clean method, which is what
wenow do with this commit.

### DIFF
--- a/ideascube/mediacenter/forms.py
+++ b/ideascube/mediacenter/forms.py
@@ -19,12 +19,18 @@ class DocumentForm(forms.ModelForm):
         }
         exclude = ['package_id']
 
-    def clean_kind(self):
-        kind = self.cleaned_data['kind']
-        original = self.cleaned_data['original']
+    def clean(self):
+        cleaned_data = super().clean()
+
+        kind = cleaned_data['kind']
+        original = cleaned_data.get('original')
+
+        if original is None:
+            # The 'original' field validator already found this was wrong
+            return cleaned_data
 
         if kind != Document.OTHER:
-            return kind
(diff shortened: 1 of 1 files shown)


ideascube/mediacenter/forms.py
```diff
@@ -19,12 +19,18 @@ class Meta:
...
-        kind = self.cleaned_data['kind']
-        original = self.cleaned_data['original']
+    def clean(self):
+        cleaned_data = super().clean()
+
+        kind = cleaned_data['kind']
+        original = cleaned_data.get('original')
+
+        if original is None:
+            # The 'original' field validator already found this was wrong
```
yohanboniface: nit about the comment: original may also be None just because it was not given (in case of an update). ;)
Given the following lines of code, I'm not sure this check needs a comment actually, but in case we add one I'd suggest to fix a bit the wording accordingly. :)

bochecha: `original` isn't None in case of an update.

yohanboniface: ah, right, something like an empty FileUpload, right?
In that case the comment is just fine :)

bochecha: Yes. Not modifying `original` is the case where the `guess_from_contenttype` fails (because in that case `original` does not have a `contenttype` attribute) and we fallback on `guess_from_extension`.

PR #500, opened 2016-09-28
Add koombook.lan on array ALLOWED_HOSTS

fheslouin:
When there is no IDEASCUBE_ID var defined in /etc/default/ideascube. If we access through koombook.lan access is not authorized

### DIFF
--- a/ideascube/conf/base.py
+++ b/ideascube/conf/base.py
@@ -25,7 +25,7 @@ SECRET_KEY = '16exrv_@=2(za=oj$tj+l_^v#sbt83!=t#wz$s+1udfa04#vz!'
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = bool(os.environ.get('DEBUG', True))
 
-ALLOWED_HOSTS = ['.ideasbox.lan', 'localhost']
+ALLOWED_HOSTS = ['.ideasbox.lan', '.koombook.lan', 'localhost']
 
 
 # Application definition


bochecha:
Should we consolidate `ALLOWED_HOSTS` into `base.py` then, and remove it from `kb.py` ?

https://github.com/ideascube/ideascube/blob/master/ideascube/conf/kb.py#L3

fheslouin:
Good one ! kb.py import base.py, so no change at the end I guess

fheslouin:
I would choose, move to `base.py` as when ideascube is installed for first time, there is no default config and it answer only to `http://ideasbox.lan` and not to `http://koombook.lan`

bochecha:
> I would choose, move to base.py as when ideascube is installed for first time, there is no default config and it answer only to http://ideasbox.lan and not to http://koombook.lan

What I'm saying is that maybe it also shouldn't answer on `http://ideasbox.lan` by default.

After all, Ideascube could be used in different contexts. Let's imagine some imaginary non-profit organization makes a « kiwix box » on a Raspberry Pi. They could want it to answer on `http://kiwi-pi.lan` (a terrible idea of a dessert).

In such a case, why should it also answer on ideasbox.lan and koombook.lan?

yohanboniface:
What about having `ideascube.lan` as default? And moving all others to their specific configuration (including bilbiotecamovil.lan) ?

PR #503, opened 2016-09-30
catalog: Don't persist the cache when loading it

### LINKED ISSUE #496: Cache won't clear itself when the hard drive is full
HDD is full:
```
ideascube@idb-jor-zaatari-52:~$ df -h --total 
Sys. de fichiers Taille Utilisé Dispo Uti% Monté sur
/dev/sda1          433G    411G     0 100% /
[...]
total              438G    411G  5,5G  99% -
```
Trying to clear the cache:
```
ideascube@idb-jor-zaatari-52:~$ ideascube catalog cache clear
IDEASCUBE_ID=idb_jor_zaatari
Importing settings from ideascube.conf.base

  File "/usr/share/python/ideascube/bin/django-admin", line 9, in <module>
    load_entry_point('Django==1.10.1', 'console_scripts', 'django-admin')()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/management/commands/catalog.py", line 115, in handle
    self.catalog = Catalog()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 456, in __init__
    self._load_catalog()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 750, in _load_catalog
    self._persist_catalog()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 755, in _persist_catalog
    persist_to_file(self._catalog_cache, self._available)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 50, in persist_to_file
    f.write(yaml.safe_dump(data, default_flow_style=False))
OSError: [Errno 28] No space left on device
ideascube@idb-jor-zaatari-52:~$
```

bochecha:
We don't need to do that, since we use the cache from memory afterwards.
Plus, operations which modify the cache in-memory (install, remove,
clear cache, ...) will persist the cache on disk when they are done.

This means we can do less work, and it nicely fixes #496

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -801,8 +801,6 @@ class Catalog:
             if installed is not None:
                 self._installed = installed
 
-        self._persist_catalog()
-
         self._package_caches = [self._local_package_cache]
 
     def _persist_catalog(self):


yohanboniface:
Not sure here.
Actually, I think the issue in #496 is about the packages cache, not the catalog metadata cache.
Even if the "no space left on device" has raised when saving the metadata, my understanding is that the real issue is that while installing the package download cache fill in the disk.
Not sure what would be the smartest way to address that. Maybe just a `--no-cache` option when installing, so the downloaded package is removed after each install?
Also, I think we have an issue in the workflow: BSF team should be using a persistent cache external of the server they are installing, not to dl every package every time. So if we are not here yet, it may be meaning that our design (`--package-cache`) is not good.

bochecha:
Look at the traceback in 496.

The problem is that @letompouce is trying to clear the cache (mostly packages), but before running the `Catalog.clear_cache` function, we first run the `Catalog._load_cache` one (in the constructor).

And the latter will run `Catalog._persist_cache`, which fails if the disk is full.

> Even if the "no space left on device" has raised when saving the metadata, my understanding is that the real issue is that while installing the package download cache fill in the disk.

No, @letompouce got this issue when trying to `catalog cache clear`.

> Not sure what would be the smartest way to address that. Maybe just a --no-cache option when installing, so the downloaded package is removed after each install?

That is another issue, completely orthogonal to this one.

yohanboniface:
I've read too quickly the traceback, thanks for the precisions!

PR #506, opened 2016-09-30
Card crashcourse

letompouce:


bochecha:
Does that work with the changes @mgautierfr made to the whole cards/packages thing?

Maybe as a first step, can you rebase your branch on master @letompouce?

mgautierfr:
@letompouce 
With my last change just merge yesterday, you don't need to add a custom html template.
As far as your package is a zim, it will be handle by the generic kiwix template.

You still need to declare a css class for the icon. the css class being the name of the package minus the lang suffix. (Ie, you're package must be named crashcourse.<lang>)

And of course, you don't need to add the card in the config files as the card will be automatically displayed at package installation.

PR #508, opened 2016-09-30
Conf cleanex

yohanboniface:
- sane default for idb.py
- move .old cards where needed
- add a idb_fra config file to factorize
- make zaatari import from azraq


ideascube/conf/idb_irq_bardarash.py
```diff
@@ -1,5 +1,5 @@
 """Bardarash in Kurdistan, Iraq"""
-from .azraq import *  # noqa
+from .idb_jor_azraq import *  # noqa
```
bochecha: So, hum, how come the tests never spotted this?

Can you try figuring it out, and fix the tests so that they would fail with this bad import?

mgautierfr:
Agree with @bochecha 's comment. Else, it's ok for me

ideascube/tests/test_settings.py
```diff
@@ -8,7 +8,7 @@ def test_all_settings(request):
     assert confdir.check(dir=True)
 
     for path in confdir.listdir():
-        if path.ext != 'py':
+        if path.ext != '.py':
```
mgautierfr: Hu, does it mean that no setting files were test before ?!
@bochecha did well to ask for fixing tests.

bochecha: Wait what?

I'm absolutely sure I got errors when I voluntarily introduced bad things in the config files, at the time I wrote this test. :confused: 

Could it have been a change in `py`?

bochecha: Should we ensure that at least 3 (base, idb and kb) config files were tested?

Maybe keep a counter of how many files have been tested, and assert it is bigger than 3?

Issue #472, opened 2016-09-07
Stock app : labels and language


barbayellow:
In the stock app, in the [demo ideascube](http://demo.ideascube.org/fr/monitoring/stock/item/1/edit/), the labels are in Spanish even though the selected language is different. So that means that we have weird labels "Nombre" that means name in Spanish and has a different meaning in French. It's confusing.

![colombia - mozilla firefox_001](https://cloud.githubusercontent.com/assets/4816690/18321262/82b9c7d0-752d-11e6-9096-9b3d53fd0080.png)

Cf screenshot

bochecha:
This seems to have been fixed by #517

PR #517, opened 2016-10-03
fixed translation for language switching

vinyll:
swapped to `lazyloading` for modules that load translation on module import.
getting the language from django language processor instead of directly from the settings.


yohanboniface:
Fixes #472

PR #520, opened 2016-10-04
Configuration for VDP2

letompouce:


ideascube/conf/idb_fra_vdp2.py
```diff
@@ -0,0 +1,4 @@
+"""Ideaxbox for Ville de Paris, France"""
```
yohanboniface: ideax? ;)

ideascube/conf/idb_fra_vdp2.py
```diff
@@ -0,0 +1,4 @@
...
+from .idb_fra import *  # noqa
+
+IDEASCUBE_NAME = u"Ville de Paris"
```
yohanboniface: You can remove `IDEASCUBE_NAME`, given it's a new server and you have it at hand.

PR #530, opened 2016-10-04
added gettext where missing

yohanboniface:
Also replaces "serial" by "identifier", because it was a bad naming.

### DIFF
--- a/ideascube/models.py
+++ b/ideascube/models.py
@@ -25,8 +25,8 @@ class TimeStampedModel(models.Model):
     An abstract base class model that provides self-
     updating ``created_at`` and ``modified_at`` fields.
     """
-    created_at = models.DateTimeField(auto_now_add=True)
-    modified_at = models.DateTimeField(auto_now=True)
+    created_at = models.DateTimeField(_('creation date'), auto_now_add=True)
+    modified_at = models.DateTimeField(_('modification date'), auto_now=True)
 
     class Meta:
         abstract = True
@@ -247,7 +247,7 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
     id_card_number = models.CharField(_('ID card number'), max_length=50,
                                       blank=True)
     children_under_12 = models.PositiveSmallIntegerField(
-                        ('Number of children under 12'), blank=True, null=True)
+        _('Number of children under 12'), blank=True, null=True)
     children_under_18 = models.PositiveSmallIntegerField(
         _('Number of children under 18'), blank=True, null=True)
     children_above_18 = models.PositiveSmallIntegerField(
--- a/ideascube/models.py
+++ b/ideascube/models.py
@@ -88,7 +88,7 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
 
     PRIVATE_DATA = ('short_name', 'full_name', 'email', 'phone')
 
-    serial = models.CharField(max_length=40, unique=True)
+    serial = models.CharField(_('identifier'), max_length=40, unique=True)
     short_name = models.CharField(_('usual name'), max_length=30, blank=True)
     full_name = models.CharField(_('full name'), max_length=100, blank=True)
     is_staff = models.BooleanField(_('staff status'), default=False,
@@ -129,7 +129,7 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
 
             if all_users.filter(serial=self.serial).exists():
                 raise ValidationError({
-                    'serial': _('User with this Serial already exists.')})
+                    'serial': _('User with this identifier already exists.')})
 
     @classmethod
     def get_data_fields(cls):
--- a/ideascube/monitoring/forms.py
+++ b/ideascube/monitoring/forms.py
@@ -21,7 +21,7 @@ def clean_barcode(barcode):
 class EntryForm(forms.Form):
 
     serials = forms.CharField(widget=forms.Textarea(attrs={
-            'placeholder': _('Enter user serials'),
+            'placeholder': _('Enter user identifiers'),
             'rows': 4
         }))
     module = forms.CharField(widget=forms.HiddenInput, required=False)
@@ -119,7 +119,7 @@ class LoanForm(forms.ModelForm):
             user = user_model.objects.get(serial=serial)
         except user_model.DoesNotExist:
             raise forms.ValidationError(
-                _('Serial {serial} not found'.format(serial=serial)))
+                _('Identifier {serial} not found'.format(serial=serial)))
         else:
             return user
 
--- a/ideascube/monitoring/forms.py
+++ b/ideascube/monitoring/forms.py
@@ -73,8 +73,9 @@ class SpecimenForm(forms.ModelForm):
 
 class InventorySpecimenForm(forms.ModelForm):
 
-    specimen = forms.CharField(widget=forms.TextInput(attrs={
-                               'placeholder': _('Enter a bar code')}))
+    specimen = forms.CharField(label=_('specimen'),
+                               widget=forms.TextInput(attrs={
+                                   'placeholder': _('Enter a barcode')}))
 
     def clean_specimen(self):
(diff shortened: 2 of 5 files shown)


ideascube/monitoring/forms.py
```diff
@@ -73,16 +74,17 @@ class Meta:
 
 class InventorySpecimenForm(forms.ModelForm):
 
-    specimen = forms.CharField(widget=forms.TextInput(attrs={
-                               'placeholder': _('Enter a bar code')}))
+    specimen = forms.CharField(label=_('specimen'),
```
vinyll: without `label=` for consistency with [previous usage](https://github.com/ideascube/ideascube/pull/530/files#diff-2f239666916c6ec0ead7cc06d8bee7efR91)

ideascube/monitoring/forms.py
```diff
@@ -46,18 +46,19 @@ def clean_serials(self):
         return set([s for s in serials.splitlines() if s])
 
 
-class ExportEntryForm(forms.Form):
+class SinceForm(forms.Form):
 
-    since = forms.DateField(
-        widget=forms.DateInput(format='%Y-%m-%d'),
-        required=False)
+    since = forms.DateField(label=_('since'),
```
vinyll: You don't need `label=` on _DateTimeField_ but you would on _DateField_?

yohanboniface: same same (but different)! ;)

ideascube/monitoring/models.py
```diff
@@ -86,7 +86,7 @@ class Specimen(models.Model):
     serial = models.CharField(_('Serial number'), max_length=100, blank=True,
                               null=True)
     item = models.ForeignKey(StockItem, related_name='specimens')
-    count = models.IntegerField(default=1)
+    count = models.IntegerField(_('count'), default=1)
```
yohanboniface: good point

ideascube/models.py
```diff
@@ -25,8 +25,8 @@ class TimeStampedModel(models.Model):
     An abstract base class model that provides self-
     updating ``created_at`` and ``modified_at`` fields.
     """
-    created_at = models.DateTimeField(auto_now_add=True)
-    modified_at = models.DateTimeField(auto_now=True)
+    created_at = models.DateTimeField(_('creation date'), auto_now_add=True)
+    modified_at = models.DateTimeField(_('modification date'), auto_now=True)
```
bochecha: I'd prefer you pass the verbose name using its keyword argument:

```python
    modified_at = models.DateTimeField(verbose_name=_('modification date'), auto_now=True))
```

Counting on the order in which the kwargs are defined is a terrible idea, as that order could change.

In addition, not all fields take the verbose name as the first argument (for example `ForeignKey`). Explicitly using the kwarg form is just much clearer, and avoids confusion.

ideascube/models.py
```diff
@@ -88,7 +88,7 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
 
     PRIVATE_DATA = ('short_name', 'full_name', 'email', 'phone')
 
-    serial = models.CharField(max_length=40, unique=True)
+    serial = models.CharField(_('identifier'), max_length=40, unique=True)
```
bochecha: Same here for the `verbose_name=` kwarg.

ideascube/monitoring/forms.py
bochecha: See what I meant about confusion due to not using `verbose_name=` above? You've confused @vinyll :smiley: 

(that's of course compounded by the fact that Django has two very different things called "field")

ideascube/monitoring/forms.py
```diff
@@ -46,18 +46,19 @@ def clean_serials(self):
...
+                            widget=forms.DateInput(format='%Y-%m-%d'),
+                            required=False)
```
bochecha: The way you've reindented everything after adding the `label=` argument makes this change bigger than it actually is, since we now have to review lines of code that haven't changed.

Compare with how much easier this is to review:

```diff
     since = forms.DateField(
+        label=_('since'),
         widget=forms.DateInput(format='%Y-%m-%d'),
         required=False)
```

Or even this, if you don't want to have too many lines:

```diff
     since = forms.DateField(
         widget=forms.DateInput(format='%Y-%m-%d'),
-        required=False)
+        required=False, label=_('since'))
```

And since the indentation was perfectly pep8-compliant before, I would really prefer if you had kept the change as small as possible, making it as easy to review as possible.

ideascube/monitoring/forms.py
```diff
@@ -46,18 +46,19 @@ def clean_serials(self):
...
 
 
-class ExportLoanForm(forms.Form):
+class ExportEntryForm(SinceForm):
+    ...
```
bochecha: I'm curious on your reason for using `Ellipsis` here instead of `pass`. :smiley:

ideascube/monitoring/forms.py
bochecha: I hate hate hate how Django uses the same term "field" for both models and forms. So much confusion! >_<

ideascube/monitoring/forms.py
```diff
@@ -73,16 +74,17 @@ class Meta:
 
 class InventorySpecimenForm(forms.ModelForm):
 
-    specimen = forms.CharField(widget=forms.TextInput(attrs={
-                               'placeholder': _('Enter a bar code')}))
+    specimen = forms.CharField(label=_('specimen'),
+                               widget=forms.TextInput(attrs={
+                                'placeholder': _('Enter a barcode')}))
```
bochecha: Indentation seems wrong here?

ideascube/monitoring/forms.py
```diff
@@ -93,11 +95,14 @@ class Meta:
...
 
 class LoanForm(forms.ModelForm):
-    specimen = forms.CharField(widget=forms.TextInput(attrs={
-                               'placeholder': _('Enter an item barcode')}))
-    user = forms.CharField(widget=forms.TextInput(attrs={
-                               'placeholder': _('Enter a user serial')}))
-    due_date = forms.DateField(widget=forms.DateInput(format='%Y-%m-%d'),
+    specimen = forms.CharField(label=_('specimen'),
+                               widget=forms.TextInput(attrs={
+                                'placeholder': _('Enter an item barcode')}))
```
bochecha: Wrong indentation?

ideascube/monitoring/forms.py
```diff
@@ -129,8 +134,9 @@ class Meta:
 
 
 class ReturnForm(forms.Form):
-    loan = forms.CharField(widget=forms.TextInput(attrs={
-                           'placeholder': _('Enter an item barcode')}))
+    loan = forms.CharField(label=_('specimen'),
+                           widget=forms.TextInput(attrs={
+                            'placeholder': _('Enter an item barcode')}))
```
bochecha: Indentation?

ideascube/models.py
yohanboniface: > I'd prefer you pass the verbose name using its keyword argument:

Agreed on the topic, but that's how is done for all the others fields, that would become inconsistent. I suggest to create a ticket to change this all in once in a separate move instead.

ideascube/monitoring/forms.py
yohanboniface: my linter does not complain so neither do I ;)
If you have a better suggestion, please be explicit :)

bochecha: At first I thought `'placeholder'` had one space too much as indentation, since it wasn't indented with the previous line.

Then I realized it in fact needs t be indented more than `widget=` since it is nested inside.

Indenting with one space is frankly ugly, and just too few to clearly see the needed indentation (the proof being that I missed it at first :wink: )

I'd suggest indenting with four spaces, as usual.

ideascube/models.py
bochecha: I'd be happy with that. Can you open the ticket?

yohanboniface:
> I would strongly prefer if those were split, each one in its own commit.

Done

PR #531, opened 2016-10-04
CommaSeparatedCharField is now checkboxes by default

vinyll:
Replacing multiple select combo box by a list of checkboxes

### DIFF
--- a/ideascube/fields.py
+++ b/ideascube/fields.py
@@ -1,4 +1,5 @@
 from select_multiple_field.models import SelectMultipleField
+from django.forms import CheckboxSelectMultiple
 
 
 class CommaSeparatedCharField(SelectMultipleField):
@@ -17,3 +18,11 @@ class CommaSeparatedCharField(SelectMultipleField):
             return ", ".join(str(choices.get(c, c)) for c in values if c)
 
         setattr(cls, 'get_%s_display' % self.name, _get_FIELD_display)
+
+    def formfield(self, **kwargs):
+        kwargs['widget'] = CheckboxSelectMultiple
+        return super().formfield(**kwargs)
+
+    def get_choices(self, **kwargs):
+        kwargs['include_blank'] = False
+        return super().get_choices(**kwargs)


bochecha:
Why remove the `blank=True` for all these fields?

vinyll:
@bochecha: We switched from a multiple combobox to a list of checkboxes.
we do not what something like this:

- [ ] ----------

That item appears when the form field is `required = True`, that is generated by the `blank=True`.
What we actually want is a nullable field.

vinyll:
Closed by mistake

bochecha:
> That item appears when the form field is required = True, that is generated by the blank=True.
What we actually want is a nullable field.

nullable or blankable? Django has both `null=True` and `blank=True`, so I'm not sure I follow.

vinyll:
@bochecha We only want to impact the form validation here. No change for the database so no change for `null=xxx`. Only [changing to `blank=False`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#blank)

bochecha:
> We only want to impact the form validation here.

What I'm asking since the first comment is why.

I'm asking because I genuinely don't understand why.

I know what `blank=` does, what I don't get is why it was necessary to change its value, in functional terms.

vinyll:
I guess a picture will speak better than I do ;)

With `blank=True`:
screen shot 2016-10-06 at 7 34 40 pm

After removing the blank argument:
screen shot 2016-10-06 at 7 30 18 pm

We want this result because it makes no sense to have the `[ ] ------------` is a checkbox list.

yohanboniface:
In other words: `blank=True` makes Django adding a blank choice in our choices, which we don't want.

yohanboniface:
> It seems to me we're (or Django is) confusing two unrelated things here: form validation and widget presentation.

I think that this is (sadly) the case. But we may have a better way to deal with the situation, and that was actually the first code of @vinyll that I then asked to change.

First thing: I thought `blank` argument was strictly a shortcut from the model field to control the `required` argument on the form field when automatically created by Django, but it's not, it's also used by the Model field `validate`: https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L582

Then, Django `does` add to the model field choices an empty choice when `blank=True`: https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L788 

This is only done if the `get_choices` receive `include_blank=True`, which is the case when `blank=True`: https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L853

So I'm (after years of using Django) still confused by the behaviour of `null/blank/required` :/

Anyway, I now sort of think that changing `blank` value here is not what we want to do, given it also affects the Model field behaviour. And then the initial option of @vinyll may be better: also overriding `get_choices` in our `CommaSeparatedField` to force `include_blank=False`.

bochecha:
> So I'm (after years of using Django) still confused by the behaviour of null/blank/required :/

great, you're confirming what I was feeling: that years from now I still will be confused by that. :laughing: 

> And then the initial option of @vinyll may be better: also overriding get_choices in our CommaSeparatedField to force include_blank=False.

Yes, this.

yohanboniface:
This certainly means rebasing and resetting the migration after https://github.com/ideascube/ideascube/pull/539

vinyll:
I've updated the PR considering your feedbacks and digging a little further.
It consistently simplifies the changes.

bochecha:
Oops, I approved before seeing the tests had failed.

Can you rebase on the latest master, and fix the unit tests before we merge?

vinyll:
Yup, that's in the pipe ;)

PR #540, opened 2016-10-07
package_id => package as label

yohanboniface:
Let's say it's just for the joy of adding migrations…


bochecha:
I still think showing the word « package » in the UI is a bad idea. (how do you translate that in French, in a way that makes sense to somebody who has never before heard of the concept of a "package manager"?)

However, as long as we're showing it, this is an obvious thing to do. :+1:

PR #545, opened 2016-10-08
Add button for downloading digital book specimens

yohanboniface:


ideascube/library/models.py
```diff
@@ -99,3 +101,10 @@ def __str__(self):
 
     def get_absolute_url(self):
         return reverse('library:book_detail', kwargs={'pk': self.item.pk})
+
+    @property
+    def extension(self):
+        if not self.file:
+            return ''
+        name, extension = os.path.splitext(self.file.name)
+        return extension[1:]
```
vinyll: for not so basic extension (like _.tar.gz_) it may not be reliable.

Maybe something like:
```
file_name = self.file.name
extension = file_name[file_name.find('.'):]
```

ideascube/library/tests/test_models.py
```diff
@@ -79,3 +79,18 @@ def test_unicode_returns_digital_specimen_of_book():
     book = BookFactory()
     specimen = DigitalBookSpecimenFactory(item=book)
     assert str(specimen).startswith(u'Digital specimen of')
+
+
+def test_specimen_extension():
+    specimen = DigitalBookSpecimenFactory(file__filename='book.epub')
+    assert specimen.extension == 'epub'
```
vinyll: add a test for double dotted extensions

ideascube/library/models.py
yohanboniface: Agreed that there are edge cases, but I think whatever the split method we chose we have some.
That said I think we can stay on that simple option, that should cover the expected extension ('.epub', '.pdf'…), and if someone actually upload a complex extension, that will not be such an issue.

yohanboniface: Agreed that there are edge cases, but I think whatever the split method we chose we have some.
That said I think we can stay on that simple option, that should cover the expected extension ('.epub', '.pdf'…), and if someone actually upload a complex extension, that will not be such an issue.

PR #554, opened 2016-10-13
Handle the exception when trying to list wifi connections

### LINKED ISSUE #551: Can't manage wifi if a connection is not owned by the user
When trying to manage the wifi, if a connection on the system is owned by another user, an exception is thrown.

Toniob:
if one of them is not user owned (fixes #551)

### DIFF
--- a/ideascube/serveradmin/wifi.py
+++ b/ideascube/serveradmin/wifi.py
@@ -168,7 +168,14 @@ class KnownWifiConnection(object):
         result = []
 
         for connection in NMSettings.ListConnections():
-            if '802-11-wireless' not in connection.GetSettings():
+            try:
+                conn_settings = connection.GetSettings()
+            except DBusException as e:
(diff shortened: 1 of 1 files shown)


ideascube/serveradmin/wifi.py
```diff
@@ -168,7 +168,12 @@ def all(cls):
...
 
         for connection in NMSettings.ListConnections():
-            if '802-11-wireless' not in connection.GetSettings():
+            try:
+                if '802-11-wireless' not in connection.GetSettings():
+                    continue
+            except DBusException as e:
+                # NM throws an exception if we try to use GetSettings on a
+                # connection owned by another user. We just ignore it.
                 continue
```
mgautierfr: Sorry but this is a bit difficult to read.
It could be better to first get the setting in the try/except and after test the value of the setting.

mgautierfr:
Thanks for the patch!

PR #562, opened 2016-10-14
Preset blog post author field with the current user

### LINKED ISSUE #526: Blog: preset author with current user

Toniob:
(fixes #526)


yohanboniface:
Quick unittest loading the form page and checking that the author is checked would be great :)

Toniob:
I've had the test.

Issue #470, opened 2016-09-06
Error when trying to add local catalog.yml


fheslouin:
I tried to add a local catalog.yml 

```
root@debian:/usr/local/share/kiwix# ideascube catalog remotes add "kiwix" "Temporary catalog" catalog.yml 
IDEASCUBE_ID=
Importing settings from ideascube.conf.base

Traceback (most recent call last):
  File "/usr/share/python/ideascube/bin/django-admin", line 9, in <module>
    load_entry_point('Django==1.9.8', 'console_scripts', 'django-admin')()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/management/commands/catalog.py", line 116, in handle
    options['func'](options)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/management/commands/catalog.py", line 229, in add_remote
    self.catalog.update_cache()
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/ideascube/serveradmin/catalog.py", line 772, in update_cache
    urlretrieve(remote.url, tmppath, reporthook=_progress)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/resumable/__init__.py", line 108, in urlretrieve
    headers=headers, **kwargs)) as resp:
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/requests/sessions.py", line 461, in request
    prep = self.prepare_request(req)
  File "/usr/share/python/ideascube/lib/python3.4/site-packages/requests/sessions.py", line 394, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
[...]

PR #563, opened 2016-10-14
Local url remote

blqn:
fix:#470

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -584,6 +584,7 @@ class Catalog:
             self._progress(' {}'.format(package.id), *args)
 
         filename = '{0.id}-{0.version}'.format(package)
+        urlparsed = urlparse(package.url)
 
         for cache in self._package_caches:
             path = os.path.join(cache, filename)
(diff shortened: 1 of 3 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -831,14 +832,18 @@ def _progress(*args):
...
-                    urlretrieve(remote.url, tmppath, reporthook=_progress)
-                except ConnectionError:
-                    print("Warning: Impossible to connect to the remote"
-                          " {remote.name}({remote.url}).\n"
-                          "Continue anyway without this remote."
-                          .format(remote=remote))
-                    continue
+                urlparsed = urlparse(remote.url)
+                if urlparsed.scheme in ['file', '']:
+                    shutil.copyfile(urlparsed.path, tmppath)
```
mgautierfr: copyfile may raise exceptions.
You should try/except them to handle them correctly (as we do when we use urlretrieve).

blqn:
I add the missing try catch.

ideascube/serveradmin/catalog.py
```diff
@@ -831,14 +832,25 @@ def _progress(*args):
...
                 # TODO: Verify the download with sha256sum? Crypto signature?
-                try:
-                    urlretrieve(remote.url, tmppath, reporthook=_progress)
-                except ConnectionError:
-                    print("Warning: Impossible to connect to the remote"
-                          " {remote.name}({remote.url}).\n"
-                          "Continue anyway without this remote."
-                          .format(remote=remote))
-                    continue
+                urlparsed = urlparse(remote.url)
```
bochecha: Minor nit: `urlparsed` and `urlparse` are very close to each other, and could cause confusion in the future.

But that's a very localized problem, `urlparsed` is used a grand total number of two times, so... ¯\_(ツ)_/¯

ideascube/serveradmin/catalog.py
```diff
@@ -831,14 +832,25 @@ def _progress(*args):
...
+                              .format(remote=remote, error=error))
+                        continue
+                else:
+                    try:
+                        urlretrieve(remote.url, tmppath, reporthook=_progress)
+                    except ConnectionError:
+                        print("Warning: Impossible to connect to the remote"
+                            " {remote.name}({remote.url}).\n"
+                            "Continue anyway without this remote."
+                            .format(remote=remote))
```
bochecha: You broke the proper indentation of the string here.

This is such an offense to good taste that I must immediately refuse this pull request in the most obnoxious way possible.

Just kidding, but try to be mindful about that in the future. :stuck_out_tongue_winking_eye:

ideascube/serveradmin/catalog.py
```diff
@@ -584,37 +584,55 @@ def _progress(*args):
...
                 if self._verify_sha256(path, package.sha256sum):
                     return path
+                if urlparsed.scheme in ['file', '']:
+                    try:
+                        shutil.copyfile(urlparsed.path, path)
+                    except Exception as error:
+                        print("Warning: Impossible to copy the package file"
+                              " {package.title}({package.url}).\n{error}\n"
+                              "Continue anyway without this remote."
+                              .format(package=package, error=error))
```
bochecha: You're copying a package, not a remote, so the message doesn't seem right.

Maybe something like `'Error copying the file: {}'.format(error)` to stay consistent with the download one?

bochecha:
I just merged your branch after:

* rebasing it on the latest master (`git fetch --all && git rebase origin/master`)
* removing the last few instances of `fake_urlretrieve` (I added them in another pull request, so you couldn't remove them here without rebasing first)
* fixing the indentation and some wording of an error message

Compared to what you had already done this was trivial, so I figured I'd do it myself and merge it. :smiley: 

Closing this now, thanks again for your work on this, it's something we've needed for a while and it will be very useful!

PR #581, opened 2016-10-20
catalog: Update the installed metadata

### LINKED ISSUE #513: Update installed metadatas on catalog cache update?
So, here we are in the dynamic cards worlds, and that's nice.

We can yet see some issues, though, because the metadata of the packages are not really ready for end user (titles too long, useless description…).
I think we need a way to keep those metadatas up to date in the servers when a catalog is updating them (for example we make a better description or, later, we add a new localized description or name).

I'm not sure how we can handle that properly, as there are a bunch of edge cases around. For example, we may want to update the name or description of an already installed package, but we certainly do not want to change it's size or sha. And we can even have cases were we'd prefer not to update the description (for example the description describe a new feature that is not yet in the old version of the package, currently installed).

In the same time needing to reinstall a package just because there is a typo in the name or description or because we need a new localization or the icon has changed is not a good workflow.

So let's try to have a thread on how to deal with that.

bochecha:
This fixes #513.

It also includes a few things I found while working on it.

Let me know if you want me to separate them into their own pull request.

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -779,6 +779,34 @@ class Catalog:
         persist_to_file(self._catalog_cache, self._available)
         persist_to_file(self._installed_storage, self._installed)
 
+    def _update_installed_metadata(self):
+        # These are the keys we must only ever update with an actual package
+        # upgrade, and thus not here.
+        blacklist = (
+            'sha256sum',
+            'type',
+            'version',
+        )
+
+        for pkgid in self._installed:
+            installed = self._installed[pkgid]
+            available = self._available.get(pkgid)
+
+            if available is None:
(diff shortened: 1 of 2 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -816,6 +816,34 @@ def _persist_catalog(self):
...
         persist_to_file(self._installed_storage, self._installed)
 
+    def _update_installed_metadata(self):
+        # These are the keys we must only ever update with an actual package
+        # upgrade, and thus not here.
+        blacklist = (
+            'sha256sum',
+            'type',
+            'version',
+        )
```
yohanboniface: I think we should only have a `checkum` key in the future, with a value containing the used algorithm, something like: `checksum: "sha256:xxxxx"`

PR #582, opened 2016-10-20
Revamp css to be more responsive.

mgautierfr:


ideascube/library/static/library/library.css
```diff
@@ -0,0 +1,42 @@
+/* ************************************************* */
```
yohanboniface: While a fully understand why you made this move, it may have unwanted side effects: we are trying to minimize the number of requests run in parallel, in order to let the Koombook handle more users.
Given that the static files are cached, the split may have no impact, but I think we should balance this and make sure we are not doing a mistake here. :)

yohanboniface:
Some before/after screenshots to understand what was explicitly covered by the PR would be nice :)

yohanboniface:
Also, can you detail a bit why you decided to switch to purecss? I see you readded flex support on top of it, so I'm a bit confused.
From my point of view we had a simple custom targeted grid custom that was doing the job (the responsiveness issues we had was afaik not related to the grid system but to our lack of screenwriting it in the templates), so I'm certainly missing a point.
I'm sure there are good reasons, but I need to know them :)

ideascube/library/static/library/library.css
bochecha: This is exactly what I told @mgautierfr on the parking lot of Pycon, right after the print where he started working on this.

One thing I hadn't paid attention at the time, and that he made me aware of, is that while the monolithic CSS file gets split into 6, each page will at most load 2 CSS files.

My worry about the number of requests went away when I realized that @mgautierfr wasn't adding 5 requests per page but only one.

I don't know if you had realized that or not before making this comment, but since I had fallen for it at first I figured I'd stress it, just to make sure we're all on the same page.

yohanboniface: I see three new requests for "purecss" files, plus "main.css", plus the specific css, am I missing something?

mgautierfr: (purecss is a different problem than splitting main.css. Here, I answer only about splitting main.css)

Styling the home page cards (using purecss) toke me very few time. Include purecss file, add tags in the html and done ... almost.

What took me time it to check that everything still work properly and that I've broke nothing.
As all rules were in the same css file, there were (and probably are) conflicts between them and some rules were rewriting other.
Few example:
- The  `.library` rules from monitoring where changing the font color to orange. But the rules `.grid .library` was resetting it.
- The rules for home cards background were not consistent, for blog (and two others), the rules was `.grid .blog` instead of just `.blog`. This was done to not conflict with the blog application where we use "card .blog" to display a blog article.

In conclusion, I spent a lot of time to check page by page that I break nothing. And it take a lot of time.
By splitting css in separate files included only when necessary, we avoid a lot of potential conflict and then simplify a lot our (future) work.

On serving those files:
- We replace 1 big css file by 2 smaller ones. There is indeed one connection more, but less data transferred. At worst a user will end by having all the css downloaded but it will no be at once.
- The static cache made by bochecha will greatly help here.
- In case of update of a application css, only this specific css is updated, all the others do not change and so are not downloaded again.

Associated with the assurance that if I change only the library.css file, only the library application will be impacted and that I don't have to check other application, I think it's worth it.

bochecha: I'd also add that if doing multiple requests for CSS becomes a problem, we could just pre-process the CSS to merge them into a big one.

It's often annoying, but there's a reason pretty much everybody does it: it allows you to split your work in a way that makes sense to you (the developer) while still bundling everything in a way that makes sense to users (and servers, since they have to answer less requests).

But since it's annoying to set up, we probably shouldn't deal with that until we find we absolutely have to. :laughing:

mgautierfr:
About PureCSS:

I concede that using purecss is not strictly necessary but:
- There is no notion of number of column in the grid/row/col, this is base on the min/max size of each element and the hope that it will be displayed nicely (and it's not, I will take screenshots). Even if internally purecss is base on element size (there is no choice here) the "API" is based on the number of column.
- Purecss probably handle better than use the problem of compatibility with all browser. (the `*property:` for old IE, prefix `-webkit-` or `-ms-`, ..)
- Mobile first design. Small device are more likely device where the media-screen is not supported or device with less performance. By doing mobile first design the extra rules are for other (more powerful) device, less efficient device will work better. It is probably anecdotal as we already know that our device work with our actual device but that a point.

Every thing could have been fixed in our custom system, but there is a minimalist library wildly tested that do every thing for us. I prefer to use it instead of patch our system to do the same.

I understand the argument of the duplication of css file to serve. For this will could merge all purecss files in on big one and so adding only one css file instead of three.

About the readd of flex support:
Purecss do not use flex on firefox cause old firefox do not support it (at time of the purecss 0.6.0 release). This is not needed and purecss works without it.
I force the use of flex to have all cards having the same height in the grid.
This should have been a different commit. I will split it to avoid confusion.

yohanboniface:
Make sure you tested everything in RTL ;)

mgautierfr:
## On big screen, there is now 4 columns:
### before
card_big_before
card_big_after

## On intermediate screen, cards are centered and take all the place
### before
card_768_to_978_before
### after
card_768_to_978_after

## On small screen, there is only one column
### before
card_small_before
### after
card_small_after

## Library media
### before
library_medium_before
### after
library_medium_after

## On small screen, menu is on top. Book covers are no more truncated
### before
library_small_before
### after
library_small_after

## Setting pages on small screen. 
### before
wifi_small_before
### after
wifi_small_after

bochecha:
@mgautierfr Looking good!

Looking at those before/after screenshots, did you change the font?

Also, on small screens, could the settings navigation be below the setting page? (your screenshot with the list of Wi-Fi networks). It seems to me that users would go to e.g the Wi-Fi page primarily to change the Wi-Fi settings, not to navigate to other pages.

barbayellow:
It seems to me that this PR is a big improvement for the user and the fact that we may have 1 more css to load is a minor problem (compared to what we experienced with kiwix for instance). 

@bochecha putting the navigation below the content is not a good idea in terms of UI, for most of the screen. However what you say about the wifi page is true. Adding accessibility links only for small screens ("Go to the content" and "Go to the menu") is usually a best practice in terms of usability for small screen http://ux.stackexchange.com/questions/69693/is-it-better-for-website-accessibility-to-have-the-menu-before-or-after-the-cont and that could solve the problem.

ideascube/static/ideascube/vendor/purecss/HISTORY.md
```diff
@@ -0,0 +1,439 @@
+Pure Change History
```
bochecha: I honestly wouldn't bundle the `HISTORY.md` file.

ideascube/static/ideascube/vendor/purecss/README.md
```diff
@@ -0,0 +1,177 @@
+Pure
```
bochecha: Do we really need to bundle their big `README.md` file, full of info we don't really care about?

I would prefer a simpler file we'd write, containing just a few info about what we've bundled:

* upstream URL
* version bundled
* download URL

ideascube/library/templates/library/book_card.html
```diff
@@ -1,14 +1,10 @@
...
-    <p>{{ book.authors }}</p>
-    <a href="{% url 'library:book_detail' pk=book.pk %}">
-    {% if book.cover %}
-        <img src="{% media book 'cover' %}" class="cover" />
-    {% else %}
-    	<img src="{% static "ideascube/img/default-book.png" %}" class="cover" />
-    {% endif %}
-    </a>
-</div>
+<a class="card {% if not book.cover %}default{% endif %}" href="{% url 'library:book_detail' pk=book.pk %}">
```
bochecha: Whether or not we move to purecss, a thousand times yes to this!

Having to click on the text, and not just anywhere on the card is driving me crazy.

I wish you'd do it for the home page cards as well. :smile:

ideascube/templates/ideascube/includes/cards/appinventor.html
```diff
@@ -1,5 +1,8 @@
 {% load i18n %}
-<div class="card tinted appinventor">
-        <h4><span class="theme create">create</span> <span><a href="http://appinventor.{{ domain }}">{% trans "App Inventor" %}</a></span></h4>
-    <p>{% trans "Create your own apps for Android." %}</p>
-</div>
+<a href="http://appinventor.{{ domain }}" class="card tinted appinventor">
```
bochecha: Oh wait, you are doing it for the home page as well!!!

:sunglasses:

mgautierfr:
> Adding accessibility links only for small screens ("Go to the content" and "Go to the menu") is usually a best practice in terms of usability for small screen http://ux.stackexchange.com/questions/69693/is-it-better-for-website-accessibility-to-have-the-menu-before-or-after-the-cont and that could solve the problem.

While I agree with this solution, I think it is better to postpone the creation of those accessibility links to later times. (Maybe when rethinking about the general UI, or before)

ideascube/library/static/library/library.css
mgautierfr: Pure css files are now merged into one. This now add only one new request per page instead of three (for purecss).

mgautierfr:
> Looking at those before/after screenshots, did you change the font?
It was purecss that set a specific font to be sure that letter-spacing rules using the 'em' unit work correctly on all browser. Then it reset back to the default font (sans-serif) for child elements. (https://github.com/yahoo/pure/issues/41/)
So we had a font change from our custom one (fira_sansmedium) to the default font (sans-serif).
I've add a rule to force the use of fira_sansmedium in child element.

yohanboniface:
Library cover alignment in RTL:

screenshot from 2016-10-28 04-48-40

yohanboniface:
Not a blocking one, but I feel we are missing a left/right margin in this kind of page:

screenshot from 2016-10-28 04-54-46

bochecha:
RTL seems fixed with the latest push to this branch:

screenshot from 2016-10-28 17-55-48

bochecha:
I think pretty much all pages lack left-right padding, at least on big screens.

I literally find myself turning my head to scan the grid on the home page. :)

yohanboniface:
Given we have a bit of time until next release, I suggest we fix the right-left padding before merge.

barbayellow:
Would it be feasible before the 9th of November?  There is a delivery of 10 KB on the field in Burundi. This PR would be of a great benefice for this project. Plus the RTL problem seems to appear only on big screens (ie the TV of the IDB). However, I think that if we do not correct this, we may leave it as it is, sweep this problem under the rug, and this not what we want.

yohanboniface:
> Would it be feasible before the 9th of November?

Sure. It very likely will be in next release, that can be anytime now, when we need it :)

>  Plus the RTL problem 

RTL issue is actually fixed, we only are left with missing right-left margins. That's a matter of a small fix, that's why I prefer to wait for it before merging. We'll very likely have other iterations and fixes alter, but we're good to go with that fix I think.

ideascube/library/templates/library/book_detail.html
```diff
@@ -1,5 +1,10 @@
 {% extends 'two-third-third.html' %}
-{% load i18n ideascube_tags %}
+{% load i18n ideascube_tags static %}
+
+{% block extra_head %}
+<link rel="stylesheet" type="text/css" href="{% static "library/library.css" %}">
+{% endblock %}
+
 {% block twothird %}
     <div class="row mwide twide">
```
yohanboniface: Shouldn't we remove those mwide twide now that we are relying on purecss?

yohanboniface:
About the margin: I think the margin of the header should be the same before the logo and after the "log in" link, and I think the body should be aligned on this.

yohanboniface:
Hi, me again! ;)

screenshot from 2016-11-03 17-00-39

Sounds like there is a margin-left: 10px left in the h1 that prevent a good alignment.

mgautierfr:
It depends on what are you comparing with :)

capture du 2016-11-03 17-38-38

Do we really want this aligned in *all* views as we know that we will rethink the interface later ?

yohanboniface:
> Do we really want this aligned in all views as we know that we will rethink the interface later ?

"later" can be anytime sadly, so I would say yes, let's try to do a clean job on what we are addressing (site layout). Then, if you say it's three full days to align those divs, give up and let's move forward.

PR #583, opened 2016-10-21
Packaging minutiae

bochecha:
This fixes various little issues of our Debian packaging.

It was split out from the monster packaging pull request (#341), so that we could at least fix some stuff which doesn't require coordination with ansiblecube.


bochecha:
Note that all of this had already been reviewed as part of #341 (I didn't change anything for this new pull request, I just cherry-picked in a new branch on top of the latest master).

However, a new review never hurts, so bring it on. :smile:

debian/changelog
```diff
@@ -229,7 +229,7 @@ ideascube (0.9.3) unstable; urgency=low
 
   * Missing import in TV5 Monde conf
 
- -- Yohan Boniface <hi@yohanboniface.me>  Tue, 30 Mar 2016 18:10:00 +0100
+ -- Yohan Boniface <hi@yohanboniface.me>  Wed, 30 Mar 2016 18:10:00 +0100
```
yohanboniface: :p

debian/ideascube.install
```diff
@@ -0,0 +1,10 @@
+extras/nginx/ideascube etc/nginx/sites-available
```
yohanboniface: Curious why we don't use the initial slash?

bochecha: Because it's not needed.

Advice I received from a Debian Developer friend was that what mattered the most was to be consistent: either never use them or always.

On top of that his general feeling was that since they aren't needed, Debian Developers tend to prefer not putting them in even though there might not be a strong consensus on this.

So I decided to trust him and remove them everywhere.

PR #585, opened 2016-10-21
Simplify the installation of static site packages

### LINKED ISSUE #431: Do not create a Nginx VHost for static-site packages
When installing a static-site package, such as `world.map`, ideascube tries to install a vhost in nginx, such as `world.map.ideasbox.lan`.

Chicken and egg issue: as root, the content might not be accessible by the ideascube user ; as ideascube, no write access shall be granted to `/etc/nginx/`.

Let's kill the egg and the chicken altogether: don't create a vhost for every package, but rather make the static-site contents accessible from a generic vhost. The `server_name` would look like `static-sites.ideasbox.lan/world.map/`, and its `root` would be `/var/ideascube/nginx/lebanon.map`.

This also would be consistent with the URI pattern we use for Kiwix contents, such as `kiwix.ideasbox.lan/wikipedia.en/`.

Biggest issue now is to find a good name for this FQDN :D

bochecha:
This moves us to having a single NGinx configuration file to handle all the static sites we install as packages.

I'm bundling with it a few minor changes I stumbled upon while working on this.

Fixes #431

### DIFF
--- a/extras/nginx/ideascube
+++ b/extras/nginx/ideascube
@@ -35,3 +35,16 @@ server {
     }
 }
 
+# serve the static sites
+server {
+    listen 80;
+    listen [::]:80;
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name sites.ideasbox.lan sites.ideasbox.local sites.koombook.lan sites.koombook.local;
+
+    charset utf-8;
+
+    root /var/ideascube/nginx/;
+}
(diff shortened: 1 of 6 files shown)


bochecha:
Note that I haven't done anything on the ansiblecube side.

As far as I can see, all that would be needed is to remove the old individual NGinx configuration files for the static site packages.

However, keeping them there shouldn't cause any problem, and it's a bit hard to identify which files to remove right now in ansiblecube.

PR #587, opened 2016-10-21
Properly get the cover when importing from ISBN

### LINKED ISSUE #565: Import with ISBN should import book cover
In the library app, one can import books from openlibrary.org using ISBN. It works when the book is available in openlibrary, however, the cover is not imported. 
If you want to try, here is an example of a book with an ISBN and a cover : https://openlibrary.org/works/OL9916058W/Exercices_d'oral_en_contexte_niveau_d%C3%A9butant

bochecha:
Fixes #565


ideascube/library/utils.py
```diff
@@ -67,10 +67,12 @@ def fetch_from_openlibrary(isbn):
 def read_url(url):
     try:
         response = urlopen(url)
-        return response.read().decode()
+        return response.read()
     except:
         # Catch all, we don't want to fail in any way.
-        return ''
+        import traceback
+        print(traceback.format_exc())
```
yohanboniface: Good point.
But maybe just print the error?

bochecha: So in the case of the bug fixed by the second commit, printing the error would print this:

```
'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

Printing the traceback as I did gives me this:

```
  File "/home/mathieu/Workspace/bsf/ideascube/ideascube/library/utils.py", line 70, in read_url
    return response.read().decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

The former would have been just as useless to fix this error as the silence.

We're printing to the logs here, this is not a user-facing string, so let's give ourselves as much help as possible to fix bugs.

PR #589, opened 2016-10-21
Fix typo

### LINKED ISSUE #575: Library import label "Mocam"
In the import screen, in the dropdown menu, one of the option is _Csv de "Mocam-en-ligne"_. It should be  is _Csv de "MoCCam-en-ligne"_. http://www.moccam-en-ligne.fr/
I changed it in Transifex, but I don't think I can change the English original version in Transifex.

bochecha:
MoCCam has two "c", just like Mathieu has only one "t".

Fixes #575


yohanboniface:
> Mathieu has only one "t".

@mgautierfr can you please review that specific bit? ;)

PR #593, opened 2016-10-26
Add maguare logo

yohanboniface:
Maguaré is a nice site for children made by the Colombian ministry of Culture: http://maguare.gov.co/


ideascube/serveradmin/catalog.py
```diff
@@ -342,7 +342,7 @@ def theme(self):
         # For this we use a dummy function who do nothing.
         # As the function is named _, gettext will mark the strings.
         _ = lambda t: t
-        if '.map' in self.id:
+        if '.map' in self.id or 'maguare' in self.id:
```
bochecha: This is going to go out of hand quickly. :-/

mgautierfr: Yes, we have to find a way to solve this hack.
But for the context of this PR, everything seem ok.

Issue #466, opened 2016-09-05
Add a favicon


yohanboniface:
Browing uwsgi logs now:

[pid: 1116|app: 0|req: 11/20] 192.168.2.141 () {32 vars in 441 bytes} [Mon Sep  5 11:09:42 2016] GET /favicon.ico => generated 2276 bytes in 85 msecs (HTTP/1.1 404) 4 headers in 133 bytes (1 switches on core 0)

Having a favicon is user friendly.
And then we should make nginx serve it.

@barbayellow giving the issue to you to have a graphic designer work on this (for ideasbox, koombook and a default ideascube one I'd say).

Related #302

yohanboniface:
BTW, given we are only targeting modern browsers, I think we could add a `meta` tag so we could locate the favicon under /static/ so Nginx will serve if without the need for a dedicated rule. Something like:

    <link rel="Shortcut Icon" href="{% static 'favicon.ico' %}" type="image/x-icon">

barbayellow:
Deal! Our graphic designer is coming back from holidays in the coming days.

yohanboniface:
Cool. Let's try to have at least a first version this week: this may reduce a lot the connection handle by the server (by caching it on client side).

bochecha:
@barbayellow Did you graphic designer get any time to work on this?

For the record, right now we have the following favicons:

* Ideasbox: https://github.com/ideascube/ideascube/blob/master/ideascube/static/ideascube/img/favicons/ideasbox.png
* Koombook: https://github.com/ideascube/ideascube/blob/master/ideascube/static/ideascube/img/favicons/koombook.png

They aren't super pretty (you can see my bad graphics skills :wink: ), so your graphic designer probably could do better. :smile:

barbayellow:
Not yet sorry. But I'll put this one on the top of the stack (not sure if we can say this I'm English but you get the idea :)

Issue #302, opened 2016-02-26
Add a default "ideascube" logo


barbayellow:
So far we only have a "ideasbox" logo out of the box (meaning with the default conf file) which is a BSF' project. We'd like this project to be used not only by BSF. We need a "neutral" logo by default.

yohanboniface:
A favicon will also be needed, see #466

barbayellow:
It's here ! It's alive !!!

Logo + favicon

![favicon-ideasbox](https://cloud.githubusercontent.com/assets/4816690/20005751/3b90e0d4-a294-11e6-9507-66376506c0f1.png)
![favicon-ideascube](https://cloud.githubusercontent.com/assets/4816690/20005754/3b91c8be-a294-11e6-9de6-165157402ae4.png)
![favicon-koombook](https://cloud.githubusercontent.com/assets/4816690/20005755/3b920608-a294-11e6-8745-d936eabd1a04.png)
![ideascube-logo](https://cloud.githubusercontent.com/assets/4816690/20005753/3b911e82-a294-11e6-9126-80bf04689a51.png)

PR #607, opened 2016-10-27
Redo the way we handle branding

bochecha:
With this merged, Ideascube now allows deployments to override the default branding simply by installing two files in the right place:
-   `/usr/share/ideascube/static/branding/favicon.png`
-   `/usr/share/ideascube/static/branding/header-logo.png`

As a consequence, all custom branding (Ideasbox and Koombook) we had inside of Ideascube is now removed, in favour of a new default branding. (see #302 and #466)

The branding might be extended in the future, so that more things can be overridden by deployments.


bochecha:
I opened https://github.com/ideascube/ansiblecube/pull/46 for the corresponding work in ansiblecube. It isn't finished though.

PR #610, opened 2016-10-28
Import users

### LINKED ISSUE #485: User import/export
We should be able to import users by batch in ideascube. This is what our Colombian partner is asking. Attached is a sample csv from their app.

Once imported, we should be able to export the information created around the users by the box managers, meaning : 
- user qualification
- user activity (from the entries app)

[Usuarios.txt](https://github.com/ideascube/ideascube/files/470816/Usuarios.txt)

yohanboniface:
Fix #485

### DIFF
--- a/ideascube/forms.py
+++ b/ideascube/forms.py
@@ -1,6 +1,13 @@
+import csv
+
 from django import forms
 from django.contrib.auth import get_user_model
 from django.utils.translation import ugettext_lazy as _
+
+from ideascube.utils import TextIOWrapper
+
+
+User = get_user_model()
 
 
 class UserForm(forms.ModelForm):
@@ -14,7 +21,7 @@ class UserForm(forms.ModelForm):
                 field.widget = forms.DateInput(format='%Y-%m-%d')
 
     class Meta:
-        model = get_user_model()
+        model = User
         exclude = ['password', 'last_login', 'is_staff']
 
 
@@ -52,3 +59,28 @@ class CreateStaffForm(forms.ModelForm):
         user.is_staff = True
         user.save()
         return user
+
+
+class UserImportForm(forms.Form):
+    source = forms.FileField(required=True)
+
+    def save(self):
+        source = TextIOWrapper(self.cleaned_data['source'].file)
+        users = []
+        errors = []
+        for idx, row in enumerate(csv.DictReader(source)):
+            qs = User.objects.all()
+            try:
+                instance = qs.get(serial=row['serial'])
+            except (User.DoesNotExist, KeyError):
+                instance = None
+            form = UserForm(data=row, instance=instance)
+            if form.is_valid():
+                user = form.save()
+                users.append(user)
+            else:
+                reason = ', '.join('{}: {}'.format(k, v.as_text())
+                                   for k, v in form.errors.items())
+                errors.append(_('Invalid row at line {id}: {reason}'.format(
+                    id=idx + 1, reason=reason)))
+        return users, errors[:10]
--- a/ideascube/tests/test_import_users.py
+++ b/ideascube/tests/test_import_users.py
@@ -1,0 +1,222 @@
+import pytest
+
+from django.contrib.auth import get_user_model
+from django.core.urlresolvers import reverse
+from webtest import Upload
+
+from .factories import UserFactory
+
+pytestmark = pytest.mark.django_db
+User = get_user_model()
+
+
+def test_anonymous_user_should_not_access_import_users_page(app):
+    response = app.get(reverse('user_import'), status=302)
+    assert 'login' in response['Location']
+
+
+def test_normal_user_should_not_access_import_users_page(loggedapp, user):
+    response = loggedapp.get(reverse('user_import'), status=302)
+    assert 'login' in response['Location']
+
+
+def test_staff_should_access_import_users_page(staffapp):
+    assert staffapp.get(reverse('user_import'), status=200)
+
+
+def test_should_create_users(staffapp, monkeypatch):
+    data = (b'serial,full_name\n'
+            b'12345,Gabriel Garcia Marquez\n'
+            b'23456,Miguel de Cervantes')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response = response.follow()
+    assert User.objects.count() == 3  # Two users created plus the staff user.
+    response.mustcontain('Successfully processed 2 users.')
+    user = User.objects.order_by('pk')[1]
+    assert user.serial == '12345'
+    assert user.full_name == 'Gabriel Garcia Marquez'
+    user = User.objects.order_by('pk')[2]
+    assert user.serial == '23456'
+    assert user.full_name == 'Miguel de Cervantes'
+
+
+def test_does_not_fail_on_unknown_column(staffapp, monkeypatch):
+    data = (b'serial,full_name,unknown\n'
+            b'12345,Gabriel Garcia Marquez,xxxx')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response.follow()
+    assert User.objects.count() == 2
+    user = User.objects.order_by('pk').last()
+    assert user.serial == '12345'
+    assert user.full_name == 'Gabriel Garcia Marquez'
+
+
+def test_does_not_fail_on_readonly_column(staffapp, monkeypatch):
+    data = (b'serial,full_name,created_at\n'
+            b'12345,Gabriel Garcia Marquez,2016-10-04 09:07:23.443699+00:00')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response.follow()
+    assert User.objects.count() == 2
+    user = User.objects.order_by('pk').last()
+    assert user.serial == '12345'
+    assert user.full_name == 'Gabriel Garcia Marquez'
+
+
+def test_does_not_import_without_serial(staffapp, monkeypatch):
+    data = (b'full_name\n'
+            b'Gabriel Garcia Marquez')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response.follow()
+    assert User.objects.count() == 1  # The staff user only.
+
+
+def test_does_not_import_with_empty_serial(staffapp, monkeypatch):
+    data = (b'serial,full_name\n'
+            b',Gabriel Garcia Marquez')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response.follow()
+    assert User.objects.count() == 1  # The staff user only.
+
+
+def test_does_not_import_if_form_is_invalid(staffapp, monkeypatch):
+    data = (b'serial,full_name,school_level\n'
+            b'12345,Gabriel Garcia Marquez,invalid')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response = response.follow()
+    response.mustcontain('Invalid row at line 1')
+    assert User.objects.count() == 1  # The staff user only.
+
+
+def test_should_import_arabic(staffapp, monkeypatch):
+    data = ('serial,full_name\n'
+            '12345,جبران خليل جبران')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data.encode(), 'text/csv')
+    response = form.submit()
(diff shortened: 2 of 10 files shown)


ideascube/forms.py
```diff
@@ -52,3 +60,37 @@ def save(self, *args, **kwargs):
...
+
+    def save(self):
+        source = self.cleaned_data['source']
+        users = []
+        errors = []
+        rows = csv.DictReader(source.read().decode().splitlines())
+        for idx, row in enumerate(rows):
+            serial = row.get('serial')
+            if not serial:
+                continue
```
bochecha: Not even a warning?

ideascube/forms.py
```diff
@@ -52,3 +60,37 @@ def save(self, *args, **kwargs):
...
+            data = {}
+            for field in User._meta.fields:
+                try:
+                    data[field.name] = row[field.name]
+                except KeyError:
+                    continue
+                if isinstance(field, models.CharField):
+                    data[field.name] = to_unicode(data[field.name])
+
+            instance = User.objects.filter(serial=serial).first()
```
bochecha: That's a weird construction. `serial` has a unicity constraint, so `.filter` doesn't feel like the right thing to use (over `.get`)

Are you doing things this way just to take advantage of the fact that `.filter(...).first()` returns `None` when the filter doesn't match anything?

To me that's relying on a side effect, and it makes the code less obvious than, say:

```python
try:
    instance = User.objects.get(serial=serial)

except User.DoesNotExist:
    instance = None
```

The latter makes it explicit that you want to allow `None` as a valid value for `instance`, and in which case that happens.

ideascube/forms.py
```diff
@@ -52,3 +60,37 @@ def save(self, *args, **kwargs):
...
+            form = UserForm(data=data, instance=instance)
+            if form.is_valid():
+                user = form.save()
+                users.append(user)
+            else:
+                reason = ', '.join('{}: {}'.format(k, v.as_text())
+                                   for k, v in form.errors.items())
+                errors.append(_('Invalid row at line {id}: {reason}'.format(
+                    id=idx + 1, reason=reason)))
```
bochecha: This would all make more sense to me if you were passing the `row` directly as data to the `UserForm`, since it handles validating it anyway.

It would also give you error messages for when a row doesn't contain the `serial`, for example (which you're silently ignoring above).

In short, it would make this function shorter and better.

ideascube/tests/test_import_users.py
```diff
@@ -0,0 +1,180 @@
...
+def test_normal_user_should_not_access_import_users_page(loggedapp, user):
+    response = loggedapp.get(reverse('user_import'), status=302)
+    assert 'login' in response['Location']
+
+
+def test_staff_should_access_import_users_page(staffapp):
+    assert staffapp.get(reverse('user_import'), status=200)
+
+
+def test_should_create_users(staffapp, monkeypatch):
```
bochecha: Maybe mention in the name that this is about importing users?

ideascube/views.py
```diff
@@ -241,6 +241,25 @@ def get_row(self, user):
...
+    form_class = UserImportForm
+    template_name = 'ideascube/user_import.html'
+    success_url = reverse_lazy('user_import')
+
+    def form_valid(self, form):
+        users, errors = form.save()
+        if users:
+            msg = _('Successfully processed {count} users.')
+            msg = msg.format(count=len(users))
+            messages.add_message(self.request, messages.SUCCESS, msg)
```
bochecha: Why not `messages.success(...)` ?

ideascube/forms.py
```diff
@@ -52,3 +60,37 @@ def save(self, *args, **kwargs):
...
+
+
+class UserImportForm(forms.Form):
+    source = forms.FileField(required=True)
+
+    def save(self):
+        source = self.cleaned_data['source']
+        users = []
+        errors = []
+        rows = csv.DictReader(source.read().decode().splitlines())
```
bochecha: Why not just `csv.DictReader(source)` ?

ideascube/views.py
```diff
@@ -241,6 +241,25 @@ def get_row(self, user):
...
+        if errors:
+            for error in errors:
```
bochecha: You can drop the `if errors` line. If `errors` is empty, then the `for` loop just won't do anything.

ideascube/views.py
yohanboniface: Good point, I forgot this API :)

ideascube/views.py
yohanboniface: Meeeeh, good catch :)

ideascube/tests/test_import_users.py
yohanboniface: The file name itself is `test_import_users` :)

ideascube/forms.py
yohanboniface: Yeah, agreed.
I've been a bit disturbed by both the need to get the serial out of the data in order to query for the instance, and for the need to convert to unicode. But I've refactored now and it's much clearer and shorter I think. Thanks :)

ideascube/tests/test_import_users.py
bochecha: Indeed, I missed that. I assumed you were just adding new test functions to test_views.py.

ideascube/forms.py
bochecha: Yes, it's shorter, more explicit, and it doesn't silently swallow errors like the missing `serial`. :+1:

ideascube/forms.py
bochecha: What I don't like here is that your use of `read` means we take the whole file in memory. If the file is long, that could be a problem.

Working with an iterator would be IMHO better here, and that's why I suggested to pass the file-like object directly to `csv.DictReader`.

But you're right, that doesn't work because of `bytes` vs `str`.

In this case, you can do:

```python
from io import TextIOWrapper

source = TextIOWrapper(self.cleaned_data['source'].file, encoding='utf-8')
rows = csv.DictReader(source)
```

And that should consume the file line by line, without taking it all in memory. :smiley:

yohanboniface: I do like this approach, but we'll have the issue to convert to unicode. We may create a generic generator function that does that (calling TextIOWrapper, itering over it, and calling to_unicode for each row). What do you think?

ideascube/utils.py
```diff
@@ -34,3 +34,16 @@ def printerr(*args, **kwargs):
...
+    """Do its best to return an unicode string."""
+    text = text or ''
+    if isinstance(text, str):
+        return text
+    else:
+        try:
+            text = text.decode('utf-8')
+        except UnicodeDecodeError:
+            text = text.decode('latin-1')
+        return text
```
bochecha: Should you also move the test of this function?

ideascube/utils.py
```diff
@@ -47,3 +47,8 @@ def to_unicode(text):
         except UnicodeDecodeError:
             text = text.decode('latin-1')
         return text
+
+
+def iter_file(file):
+    for row in file:
+        yield to_unicode(row)
```
bochecha: Oh no, you did it.

Sorry, I should have said earlier that this wouldn't work, see [my previous comment](https://github.com/ideascube/ideascube/pull/610#discussion_r86303725)

Please move back to what you had before.

ideascube/tests/test_import_users.py
```diff
@@ -0,0 +1,180 @@
...
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response.follow()
+    user = User.objects.order_by('pk')[1]
+    assert user.serial == '12345'
+    assert user.full_name == 'Gabriel Garcia Marquez'
+    assert user.is_staff is False
+    user = User.objects.order_by('pk')[2]
+    assert user.is_staff is True
+    assert user.full_name == 'Miguel de Cervantes'
```
bochecha: Maybe add a quick test to ensure the `__system__` user is not exported?

yohanboniface: OK will do.

ideascube/utils.py
yohanboniface: Sorry, I don't get why it does not work. Aren't we decoding line by line?

ideascube/utils.py
yohanboniface: Oh, right, I'll do.

ideascube/utils.py
bochecha: > Sorry, I don't get why it does not work. Aren't we decoding line by line?

And that's exactly the problem, as I mentioned in my previous comment.

Say the file is in the Latin-1 encoding, but the first line decodes fine as UTF-8.

Your iterator will yield the first line decoded as UTF-8, then the next line as Latin-1, then...

But the fact that a Latin-1 line decodes fine as UTF-8 doesn't mean that you get the same unicode string in both cases.

Therefore, the iterator might very well end up yielding wrong data for the line that decode fine as UTF-8.

Guessing the encoding of a file must be done for the whole file at once.

ideascube/forms.py
```diff
@@ -52,3 +59,27 @@ def save(self, *args, **kwargs):
...
+                instance = None
+            form = UserForm(data=row, instance=instance)
+            if form.is_valid():
+                user = form.save()
+                users.append(user)
+            else:
+                reason = ', '.join('{}: {}'.format(k, v.as_text())
+                                   for k, v in form.errors.items())
+                errors.append(_('Invalid row at line {id}: {reason}'.format(
+                    id=idx + 1, reason=reason)))
```
bochecha: > yohanboniface 5 days ago ideascube member
>
> I suggest to export only the, say, 10 errors: if a same error apply to all lines, we don't want to display one error per line; and it's unlikely that a user will find useful to have, say, 100 errors displayed at a time, even if they are all different.

:-°

ideascube/tests/test_import_users.py
bochecha: > is there any damage that we could do?

Change the password for the `__system__` user, making it possible for someone to login with its credentials?

Also, the way you match the serial when importing... Would that match the existing `__system__` user since it's filtered out of the default queryset?

If not, and the file contains a `__system__` user, would that just fail the import with a DB integrity error?

mgautierfr: > Also, the way you match the serial when importing... Would that match the existing __system__ user since it's filtered out of the default queryset?
> 
> If not, and the file contains a __system__ user, would that just fail the import with a DB integrity error?

Yes, it will fail. As the __system__ user is filtered, it tries to create a new __system__ user and fails with a duplicate primary key error.

yohanboniface: > Change the password for the __system__ user, making it possible for someone to login with its credentials?

We are not taking password into account, so I think we are safe on this side (just added some unittests to cover the password scenario). We may want at some point take the password into account to be able to import even the staff users, but that's out of current scope.

> Yes, it will fail. As the system user is filtered, it tries to create a new system user and fails with a duplicate primary key error.

Fixed and added a test to cover the scenario.

ideascube/forms.py
```diff
@@ -52,3 +59,29 @@ def save(self, *args, **kwargs):
...
+
+
+class UserImportForm(forms.Form):
+    source = forms.FileField(required=True)
+
+    def save(self):
+        source = self.cleaned_data['source']
+        users = []
+        errors = []
+        source = to_unicode(source.read().decode())
```
bochecha: Wait, why are you decoding the bytes before passing the result to `to_unicode`?

yohanboniface: bah, I was certainly tired yesterday night :p
Anyway, I'm waiting for your branch to be merged so I can use the Automagic thing :)

ideascube/tests/test_import_users.py
```diff
@@ -0,0 +1,222 @@
...
+    response.follow()
+    user = User.objects.order_by('pk')[1]
+    assert user.serial == '12345'
+    assert user.full_name == 'Gabriel Garcia Marquez'
+    assert user.password == before.password
+
+
+@pytest.mark.usefixtures('systemuser')
+def test_does_not_try_to_create_system_user(staffapp, monkeypatch):
+    qs = User.objects.all(include_system_user=True)
```
bochecha: Aren't we removing the `.all(include_system_user=True)` API in favour of the `.get_unfiltered_queryset()` one?

https://github.com/ideascube/ideascube/pull/616#discussion_r85548778

ideascube/tests/test_import_users.py
```diff
@@ -0,0 +1,222 @@
...
+    assert qs.count() == 2
+    data = (b'serial,full_name,unknown\n'
+            b'__system__,ANewFunkyName,xxxx')
+    form = staffapp.get(reverse('user_import')).forms['import']
+    form['source'] = Upload('users.csv', data, 'text/csv')
+    response = form.submit()
+    response.follow()
+    assert qs.count() == 2
+    user = qs.get(serial='__system__')
+    assert user.full_name == 'ANewFunkyName'
```
bochecha: I strongly suggest we **do not** change the `__system__` user at import.

We should completely ignore any `__system__` serial in the import source, either silently or with a warning message.

yohanboniface: OK, I'll do that.

ideascube/tests/test_import_users.py
yohanboniface: Ekkkkk, good catch! :)

PR #615, opened 2016-10-28
Export and import stock items

bochecha:


yohanboniface:
Quite important coverage decrease, you checked why? May still be one of those percent side effects, but this seems a bit high.

ideascube/monitoring/forms.py
```diff
@@ -149,3 +151,37 @@ def clean_loan(self):
...
             loan = Loan.objects.due().filter(specimen__barcode=barcode).first()
         return loan
+
+
+class StockImportForm(forms.Form):
+    source = forms.FileField(label=_('CSV File'), required=True)
+
+    def save(self):
+        source = self.cleaned_data['source']
+        source = TextIOWrapper(source.file, encoding='utf-8')
```
yohanboniface: Watch out that we can still have non UTF-8 CSV. `to_unicode` helpers should be here to handle the most common cases (latin1…).

ideascube/monitoring/views.py
```diff
@@ -217,6 +218,46 @@ class StockItemDelete(DeleteView):
...
+    template_name = 'monitoring/stock_import.html'
+    success_url = reverse_lazy('monitoring:stock')
+
+    def form_valid(self, form):
+        items, errors = form.save()
+
+        if items:
+            messages.success(
+                self.request,
+                _('Successfully import {} elements').format(len(items)))
```
yohanboniface: Let's use item, like we do in other user facing strings.

ideascube/monitoring/views.py
```diff
@@ -217,6 +218,46 @@ class StockItemDelete(DeleteView):
...
 
 
+class StockExport(CSVExportMixin, View):
+    prefix = 'stock'
+
+    def get_headers(self):
+        return ['module', 'name', 'description']
+
+    def get_items(self):
+        qs = StockItem.objects.all()
```
yohanboniface: `return StockeItem.objects.filter(xxx)` does not fit in one line?

ideascube/monitoring/urls.py
```diff
@@ -9,7 +9,7 @@
     url(r'^stock/inventory/(?P<pk>[\d]+)/$', views.inventory, name='inventory'),  # noqa
     url(r'^stock/inventory/(?P<pk>[\d]+)/edit/$', views.inventory_update, name='inventory_update'),  # noqa
     url(r'^stock/inventory/(?P<pk>[\d]+)/export/$', views.inventory_export, name='inventory_export'),  # noqa
-    url(r'^inventory/(?P<pk>[\d]+)/delete/$', views.inventory_delete, name='inventory_delete'),  # noqa
+    url(r'^stock/inventory/(?P<pk>[\d]+)/delete/$', views.inventory_delete, name='inventory_delete'),  # noqa
```
yohanboniface: Good catch :)

ideascube/mixins.py
```diff
@@ -117,7 +117,7 @@ def render_to_csv(self):
         filename = self.get_filename()
         attachment = 'attachment; filename="{name}.csv"'.format(name=filename)
         response['Content-Disposition'] = attachment
-        response['Content-Type'] = 'text/csv'
+        response['Content-Type'] = 'text/csv; charset=utf-8'
```
yohanboniface: Nice!

ideascube/monitoring/tests/test_views.py
```diff
@@ -88,6 +89,35 @@ def test_staff_should_access_stock_page(staffapp):
...
+
+    assert response.content_type == 'text/csv'
+    assert response.content_disposition.startswith(
+        'attachment; filename="stock_')
+    lines = response.unicode_body.strip().split('\r\n')
+    assert len(lines) == 2
+    assert lines[0] == 'module,name,description'
+    assert lines[1].startswith('digital,Stock item ')
+
+
```
yohanboniface: Are we missing import tests?

ideascube/monitoring/forms.py
```diff
@@ -149,3 +151,37 @@ def clean_loan(self):
...
+                item = StockItem(**data)
+                item.save()
+                items.append(item)
+
+            except Exception as e:
+                errors.append(_('Could not import line {}: {}').format(
+                    index + 1, e))
+                continue
+
+        return items, errors
```
yohanboniface: I suggest to export only the, say, 10 errors: if a same error apply to all lines, we don't want to display one error per line; and it's unlikely that a user will find useful to have, say, 100 errors displayed at a time, even if they are all different.

ideascube/monitoring/forms.py
```diff
@@ -149,3 +151,37 @@ def clean_loan(self):
...
+
+            except KeyError as e:
+                errors.append(_('Missing column "{}" on line {}').format(
+                    e.args[0], index + 1))
+                continue
+
+            try:
+                item = StockItem(**data)
+                item.save()
+                items.append(item)
```
yohanboniface: I'd move this line to an `else` block, as it's supposed to raise any exception we want to catch here.

ideascube/monitoring/tests/test_views.py
bochecha: Yes, we were. Added.

ideascube/utils.py
```diff
@@ -1,8 +1,27 @@
+from io import TextIOWrapper
 import sys
 
 from django.conf import locale
 
 
+class AutomagicTextIOWrapper(TextIOWrapper):
```
yohanboniface: Maybe add a comment explaining why we use this wrapper and we not simply read the whole file as unicode then splitlines on it?

bochecha: That comment would fit where we use the `AutomagicTextIOWrapper` (by the way, I dislike this name and would welcome a suggestion for a better one).

But then we'd write that comment everywhere we use it?

yohanboniface: This is why I suggest to just put it on the class itself: curious people will end there if they want to understand why we use it :)
About the name: not sure we need to keep the reference to TextIoWrapper in it. I'd go for something like IterDecode, because that's what it does for us. Thoughts?

bochecha: > This is why I suggest to just put it on the class itself: curious people will end there if they want to understand why we use it :)

We use it for the exact same reasons we'd use a `TextIOWrapper`. Also, it's not a very specific class, and (if moved to use cchardet) could even be sent to upstream cchardet.

In that scenario, where it gets accepted in cchardet and we use it from them, where would you put the comment?

> About the name: not sure we need to keep the reference to TextIoWrapper in it.

It really is a TextIOWrapper though, with all its bells and whistles. We use it only to decode as we iterate, but we could use it for something else. (e.g call its `.read()` method)

In fact, I'm tempted to name it just `TextIOWrapper`, since we'd just use it everywhere in Ideascube instead of Python's `TextIOWrapper`. :smile:

ideascube/monitoring/views.py
```diff
@@ -159,9 +159,9 @@ class InventoryDelete(DeleteView):
 class InventoryExport(CSVExportMixin, DetailView):
     model = Inventory
 
-    def get(self, request, *args, **kwargs):
+    def get(self, *args, **kwargs):
         self.object = self.get_object()
```
mgautierfr: Do we need to keep this assignation ?
We can use directly `self.get_object()` in `get_row` and `get_filename` and not redefined at all the `get` method.

bochecha: There's more I want to do to the `CSVExportMixin` to simplify it even more.

I had started, but hurt myself to this `InventoryExport` view being a bit weird: its model is the `Inventory` but the items it exports are `Specimen` instances. Other views deriving from the `CSVExportMixin` don't have this peculiarity, so I couldn't easily just do the changes I had in mind.

But I'll get back to it eventually. :wink: 

Therefore, I'd request we leave this view as it is for now. Works for you?

ideascube/utils.py
mgautierfr: AutoEncodingTextIoWrapper ?
But it will be a long name...

bochecha: >> This is why I suggest to just put it on the class itself: curious people will end there if they want to understand why we use it :)

> We use it for the exact same reasons we'd use a TextIOWrapper. Also, it's not a very specific class, and (if moved to use cchardet) could even be sent to upstream cchardet.

Coming back on this, because I just realized we might be approaching this from a different angle.

When I say « We use it for the exact same reasons we'd use a `TextIOWrapper` », I'm looking at it from the perspective that using a `TextIOWrapper`-like solution is the obvious thing to do.

It seems to me that what you're implying is that, for you, the obvious thing to do would have been to `splitlines` and use an array, which makes using a `TextIOWrapper` here a sort of special case in your view. Am I right?

If that is indeed your view, then I'm going to have to respectfully disagree with it, and here is why.

The right way to read files line by line (and we all know that TOOWTDI :stuck_out_tongue_winking_eye:) is:

```python
with open('path', 'r') as f:
    for line in f:
        ...
```

I would consider the following to be a mistake because it reads the whole file (which might be very big) at once and loads it all in memory, iterates over the content once to split it, then iterate again over the list of lines:

```python
with open('path', 'r') as f:
    lines = f.read().splitlines()

    for line in lines:
        ...
```

To me the above example naturally leads to using a generator-like solution to decoding the lines. `TextIOWrapper` is such a solution, and it is builtin with Python. Does that not make it the right choice?

All this to say that I find nothing surprising in using a `TextIOWrapper`-like solution here, and really don't see the need to specifically comment on its usage.

I would have instead found a solution based on `.read().splitlines()` more surprising and deserving of a comment. (« we can't just iterate over the file as we don't know its encoding » or something)

In fact, if you remember, [I did find it surprising and asked for details](https://github.com/ideascube/ideascube/pull/610#discussion_r85533166). :smile:

yohanboniface: > It seems to me that what you're implying is that, for you, the obvious thing to do would have been to splitlines and use an array, which makes using a TextIOWrapper here a sort of special case in your view. Am I right?

Nope. The obvious thing to do would have been to just give the file to DictReader, but cannot because… and this is where the comment comes in ;)

yohanboniface: (And I'm totally fine using TextIOWrapper.)

bochecha: >> It seems to me that what you're implying is that, for you, the obvious thing to do would have been to splitlines and use an array, which makes using a TextIOWrapper here a sort of special case in your view. Am I right?

> Nope. The obvious thing to do would have been to just give the file to DictReader, but we cannot because… and this is where the comment comes in ;)

Oh, so you want a comment explaining that we need to decode the `bytes` to `str`?

From your first comment, I thought you wanted a comment explaining why we use an iterator-like solution instead of `to_unicode(source.read()).splitlines()`.

I guess this whole conversation was based on a misunderstanding, then.

ideascube/monitoring/views.py
mgautierfr: Yes, we can go with this version.

PR #616, opened 2016-10-28
Test with migration

mgautierfr:
Be able to run test with migration or not.

The default (run without migration) is not changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 test:
 	py.test
 testcov:
-	py.test --cov=ideascube/ --cov-report=term-missing
+	py.test --cov=ideascube/ --cov-report=term-missing --migrations
 develop:
 	pip3 install -r requirements-dev.txt
 doc:
--- a/conftest.py
+++ b/conftest.py
@@ -22,12 +22,17 @@ def cleansearch():
 
 @pytest.fixture()
 def user(db):
-    return UserFactory(short_name="Hello", password='password')
+    return UserFactory(serial='123456',
+                       short_name="Hello",
+                       password='password')
 
 
 @pytest.fixture()
 def staffuser(db):
-    return UserFactory(short_name="Hello", password='password', is_staff=True)
+    return UserFactory(serial='123456staff',
+                       short_name="Hello",
+                       password='password',
+                       is_staff=True)
 
 
 @pytest.fixture()
--- a/conftest.py
+++ b/conftest.py
@@ -37,8 +37,15 @@ def staffuser(db):
 
 @pytest.fixture()
 def systemuser(db):
+    from django.contrib.auth import get_user_model
     # Create it the same way the migration does it
-    return UserFactory(serial='__system__', full_name='System', password='!!')
+    User = get_user_model()
+    try:
+        return User.objects.get_system_user()
+    except User.DoesNotExist:
+        return UserFactory(serial='__system__',
+                           full_name='System',
+                           password='!!')
 
 
 @pytest.fixture
@@ -175,4 +182,3 @@ def catalog():
     mocker = CatalogMocker()
     with mocker:
         yield mocker
-
--- a/ideascube/db_router.py
+++ b/ideascube/db_router.py
@@ -1,30 +1,53 @@
+
+_database_to_use = {
+    ('search', 'search'): 'transient'
+}
+
+
 class DatabaseRouter(object):
     """
     A Database router.
     Some models do not need (and must not be backuped).
-    Those models will (must) have a _dbname class attribute set to 'transiant'.
-    This attribute will be use as database name.
+    So we route them to the transient base.
     """
     def db_for_read(self, model, **hints):
         """
-        If the model as a _dbname set, use it as db for read
+        Return the database from which this model can be read,
+        or None to use the default database.
         """
-        return getattr(model, '_dbname', None)
+        app_model = model._meta.app_label, model._meta.model_name
+        return _database_to_use.get(app_model)
 
     def db_for_write(self, model, **hints):
         """
-        If the model as a _dbname set, use it as db for read
+        Return database for which this model can be write,
+        or None to use the default database.
         """
-        return getattr(model, '_dbname', None)
+        app_model = model._meta.app_label, model._meta.model_name
+        return _database_to_use.get(app_model)
 
     def allow_relation(self, obj1, obj2, **hints):
+        """Return whether a relation is allowed between these two models
+
+        We only allow relations between models stored in the same database.
         """
-        Allow relations if both object are in the same database.
-        """
-        return getattr(obj1, '_dbname', None) == getattr(obj2, '_dbname', None)
+        # I'm not sure of how to correctly implement this method.
+        # It was the same for allow_migrate and I made it wrong.
+
+        # However, there is no JOIN relation between two tables of two
+        # different db for now. So it seems safe to me to simply allow relation
+        # if the tables are in the same db.
+
+        app_model1 = obj1._meta.app_label, obj1._meta.model_name
+        app_model2 = obj2._meta.app_label, obj2._meta.model_name
+        return _database_to_use.get(app_model1) \
+            == _database_to_use.get(app_model2)
 
     def allow_migrate(self, db, app_label, model_name=None, **hints):
         """
-        All database need to be migrated.
+        Allow migration only if the migration is applying to the right db.
         """
-        return None
+        db_to_use = _database_to_use.get((app_label, model_name), 'default')
+        if db_to_use != db:
+            return False
+        return True
--- a/ideascube/db_router.py
+++ b/ideascube/db_router.py
@@ -50,4 +50,7 @@ class DatabaseRouter(object):
         db_to_use = _database_to_use.get((app_label, model_name), 'default')
         if db_to_use != db:
             return False
+        using = hints.get('using')
+        if using:
+            return using == db
         return True
--- a/ideascube/library/migrations/0008_library_stock.py
+++ b/ideascube/library/migrations/0008_library_stock.py
@@ -163,7 +163,7 @@ class Migration(migrations.Migration):
             },
             bases=('monitoring.specimen', models.Model),
         ),
-        migrations.RunPython(migrate_data),
+        migrations.RunPython(migrate_data, hints={'using': 'default'}),
         migrations.DeleteModel('OldBookSpecimen'),
         migrations.DeleteModel('OldBook'),
     ]
--- a/ideascube/migrations/0009_add_a_system_user.py
+++ b/ideascube/migrations/0009_add_a_system_user.py
@@ -6,7 +6,7 @@ from django.db import migrations
 
 def add_user(apps, *args):
     # The apps.get_model() method returns us a 'fake' model with a
-    # default manager and the 'save' method is not overwrited
+    # default manager and the 'save' method is not overwritten
     # and the user is not indexed at creation.
     # This is not a problem here as we have post-migrate hook who reindex
     # everything at end of migration. Moreover, it is better has we do not
--- a/ideascube/migrations/0009_add_a_system_user.py
+++ b/ideascube/migrations/0009_add_a_system_user.py
@@ -5,6 +5,13 @@ from django.db import migrations
 
 
 def add_user(apps, *args):
+    # The apps.get_model() method returns us a 'fake' model with a
+    # default manager and the 'save' method is not overwrited
+    # and the user is not indexed at creation.
+    # This is not a problem here as we have post-migrate hook who reindex
+    # everything at end of migration. Moreover, it is better has we do not
+    # have to care about if the 'transient' db is existing or not when doing
+    # this migration.
     User = apps.get_model('ideascube', 'User')
     User(serial='__system__', full_name='System', password='!!').save()
 
@@ -17,5 +24,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(add_user, None),
+        migrations.RunPython(add_user, None, hints={'using': 'default'}),
     ]
--- a/ideascube/models.py
+++ b/ideascube/models.py
@@ -64,13 +64,6 @@ class UserManager(BaseUserManager.from_queryset(UserQuerySet)):
         # Use the parent's queryset, as it is not filtered
         return super().get_queryset()
 
-    def all(self, include_system_user=False):
-        if include_system_user:
-            return self.get_queryset_unfiltered().all()
-
-        else:
-            return super().all()
-
     def get_system_user(self):
         return self.get_queryset_unfiltered().get(serial='__system__')
 
@@ -131,7 +124,7 @@ class User(SearchMixin, TimeStampedModel, AbstractBaseUser):
         if self.id is None:
             # We are creating a new user.
             # Include system user in the check for uniqueness.
-            all_users = User.objects.all(include_system_user=True)
+            all_users = User.objects.get_queryset_unfiltered()
 
             if all_users.filter(serial=self.serial).exists():
                 raise ValidationError({
--- a/ideascube/search/migrations/0001_initial.py
+++ b/ideascube/search/migrations/0001_initial.py
@@ -6,9 +6,11 @@ from ideascube.search.utils import create_index_table
 
 
 class CreateSearchModel(migrations.CreateModel):
-    def database_forwards(self, *_):
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
         # Don't run the parent method, we create the table our own way
-        create_index_table()
+        model = to_state.apps.get_model(app_label, self.name)
+        if self.allow_migrate_model(schema_editor.connection.alias, model):
+            create_index_table()
 
 
 class Migration(migrations.Migration):
--- a/ideascube/search/migrations/0002_drop_old_index_table.py
+++ b/ideascube/search/migrations/0002_drop_old_index_table.py
@@ -14,5 +14,5 @@ class Migration(migrations.Migration):
     dependencies = [('search', '0001_initial')]
 
     operations = [
-        migrations.RunPython(migrate),
+        migrations.RunPython(migrate, hints={'using': 'default'}),
     ]
--- a/ideascube/tests/test_models.py
+++ b/ideascube/tests/test_models.py
@@ -61,36 +61,3 @@ def test_user_data_fields_should_return_labels_and_values(settings):
     assert str(fields['ar_level']['value']) == 'Understood, Spoken'
     assert str(fields['ar_level']['label']) == 'Arabic knowledge'
 
-
-class JSONModel(models.Model):
-    class Meta:
-        app_label = 'ideascube'
-
-    data = JSONField()
-
-
-@pytest.mark.parametrize(
-    'value',
-    [
-        True,
-        42,
-        None,
-        'A string',
-        [1, 2, 3],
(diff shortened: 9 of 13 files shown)


ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -6,8 +6,9 @@
 
 
 def add_user(*args):
-    User = get_user_model()
-    User(serial='__system__', full_name='System', password='!!').save()
+    queryset = get_user_model().objects.get_queryset_unfiltered()
+    queryset.update_or_create(serial='__system__',
```
bochecha: I've been going back and forth on whether we should just create or `update_or_create` here.

At the point we run this migration, the `__system__` user really shouldn't exist, so the migration can't fail on the serial already being taken. It if does, it means somebody actually created a user with that serial, and I'm not sure we should silently change their password. :smile: 

However, failing in the migration means (at the moment) failing the Debian package update, which really must not happen, so I guess `update_or_create` is marginally better here.

ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -6,8 +6,9 @@
...
+                              defaults={'full_name':'System', 'password':'!!'})
```
bochecha: I have the same remark about `get_queryset_unfiltered` here.

However, thinking about it some more, I don't like this any better:

```python
    get_user_model().objects.all(include_system_user=True).update_or_create(...)
```

So maybe we should just drop the `.all(include_system_user=True)` method, and just use `get_queryset_unfiltered()` everywhere. (it's currently used in exactly one place in the code and in one test)

Not something for this pull request, but something to keep in mind.

ideascube/tests/test_models.py
```diff
@@ -37,13 +37,17 @@ def test_create_superuser():
 def test_list_users():
     model = get_user_model()
     model.objects.create_user('123456')
-    model.objects.create_user('__system__')
+
+    unfiltered_queryset = model.objects.get_queryset_unfiltered()
+    if not len(unfiltered_queryset.filter(serial='__system__')):
```
bochecha: I prefer:

```python
if not unfiltered_queryset.filter(serial='__system__').exists():
```

It's much more explicit what you're actually checking.

However, why not just use the `systemuser` fixture? :)

ideascube/tests/test_models.py
```diff
@@ -72,10 +72,19 @@ def test_user_data_fields_should_return_labels_and_values(settings):
...
 
     data = JSONField()
 
 
+@pytest.fixture()
+def json_table():
+    from django.db import connections
+    cursor = connections['default'].cursor()
+    cursor.execute("CREATE TABLE ideascube_jsonmodel "
+                   "(id, data)")
```
bochecha: *Putting my comment here even though it is not related to this line, but there's no place comment on a commit message.*

> If we run tests with migration, django fail to correctly migrate the json table.

What? Why would it even try to migrate it?

conftest.py
```diff
@@ -32,8 +32,12 @@ def staffuser(db):
 
 @pytest.fixture()
 def systemuser(db):
+    from django.contrib.auth import get_user_model
     # Create it the same way the migration does it
-    return UserFactory(serial='__system__', full_name='System', password='!!')
+    unfiltered_queryset = get_user_model().objects.get_queryset_unfiltered()
```
mgautierfr: You're right. Your solution is better.

ideascube/migrations/0009_add_a_system_user.py
mgautierfr: I have a slicy preference to keep `get_queryset_unfiltered()` but it is not so important.
Whatever what we keep we should move to keep only one way to do is better.

ideascube/tests/test_models.py
mgautierfr: That's a good question : Why haven't you use the fixture ? :)
I will change this  to use the fixture.

ideascube/tests/test_models.py
mgautierfr: That a good question.
I wonder if in fact, it just do not create the table as there is no migration that create the model.
So it fail to use it. If we run test without migration, django create all table from the defined model (and doesn't care of the migration script) and tests pass.

ideascube/migrations/0009_add_a_system_user.py
bochecha: Since we agree, want to do it in this pull request? :stuck_out_tongue:

ideascube/tests/test_models.py
bochecha: I think we (not surewhich one of us it was) added the fixture after I had written these test methods.

ideascube/tests/test_models.py
bochecha: So, I created this model just for the purpose of testing the `JSONField`. I don't actually care about this model.

Your solution is fine for now, but what I'd like to do eventually is just drop this model entirely, and just test the field on its own. We just need to figure out how. :smile:

ideascube/tests/test_models.py
```diff
@@ -34,16 +34,13 @@ def test_create_superuser():
...
     model = get_user_model()
-    model.objects.create_user('123456')
-    model.objects.create_user('__system__')
-
     users = model.objects.all()
     assert [u.serial for u in users] == ['123456']
 
     users = model.objects.all(include_system_user=True)
-    assert [u.serial for u in users] == ['__system__', '123456']
+    assert set([u.serial for u in users]) == set(['__system__', '123456'])
```
bochecha: Why do you move to a set here?

On the contrary, I'd argue we really want to not use a set and allow for duplicates, because that would be an error which should fail the tests.

ideascube/models.py
```diff
@@ -131,7 +124,7 @@ def validate_unique(self, exclude=None):
         if self.id is None:
             # We are creating a new user.
             # Include system user in the check for uniqueness.
-            all_users = User.objects.all(include_system_user=True)
+            all_users = User.objects.get_queryset_unfiltered().all()
```
bochecha: Minor nit: you don't need the `.all()` here.

ideascube/tests/test_models.py
mgautierfr: Hum you're right, I don't catch any duplicate here.
But I move to a set to be independent of the order of the list.

ideascube/tests/test_models.py
```diff
@@ -34,16 +34,13 @@ def test_create_superuser():
...
     model = get_user_model()
-    model.objects.create_user('123456')
-    model.objects.create_user('__system__')
-
     users = model.objects.all()
     assert [u.serial for u in users] == ['123456']
 
     users = model.objects.all(include_system_user=True)
-    assert [u.serial for u in users] == ['__system__', '123456']
+    assert sorted([u.serial for u in users]) == sorted(['__system__', '123456'])
```
bochecha: I would have preferred ordering the query directly (`.order_by('serial')`) and not using `sorted` on the list on the right-hand side of the equality.

But I won't block on this.

bochecha:
Just looked at the changes in your last push, I maintain my approval of this pull request. :smiley:

Makefile
```diff
@@ -1,7 +1,7 @@
 test:
 	py.test
 testcov:
-	py.test --cov=ideascube/ --cov-report=term-missing
+	py.test --cov=ideascube/ --cov-report=term-missing --migrations
```
yohanboniface: We may want to remove the explicit run of `migrate` from travis, then, no?

https://github.com/ideascube/ideascube/blob/master/.travis.yml#L12

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
+
+_databaseToUse = {
+    ('search', 'search') : 'transient'
+}
```
bochecha: So you're adding a mapping here, in addition to the one on the model?

Why not simply get the model from the `app_label` and `model_name`?

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
+
+_databaseToUse = {
+    ('search', 'search') : 'transient'
```
bochecha: Using a tuple as a key works, but is not super elegant IMHO.

Why not model names as keys of a dict for the app?

```python
_databaseToUse = {
    'search': {
        'search': 'transient',
    }
}
```

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
+
+_databaseToUse = {
```
bochecha: camelCase?

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
...
+        Route the table'read to the right base or the default one (None)
         """
-        return getattr(model, '_dbname', None)
+        app_model = model._meta.app_label, model._meta.model_name
+        return _databaseToUse.get(app_model)
 
     def db_for_write(self, model, **hints):
         """
-        If the model as a _dbname set, use it as db for read
+        Route the table'write to the right base or the default one (None)
```
bochecha: `table'write`?

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
...
     A Database router.
     Some models do not need (and must not be backuped).
-    Those models will (must) have a _dbname class attribute set to 'transiant'.
+    So we route them to the transient base.
     This attribute will be use as database name.
     """
     def db_for_read(self, model, **hints):
         """
-        If the model as a _dbname set, use it as db for read
+        Route the table'read to the right base or the default one (None)
```
bochecha: This `(None)` is too few information to make sense.

Something like `Return the right database for this model, or None to use the default database.` would be easier to understand in 6 months when we find a new problem with the way testing works in Django and we look at this code again.

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
...
+        Route the table'write to the right base or the default one (None)
         """
-        return getattr(model, '_dbname', None)
+        app_model = model._meta.app_label, model._meta.model_name
+        return _databaseToUse.get(app_model)
 
     def allow_relation(self, obj1, obj2, **hints):
         """
         Allow relations if both object are in the same database.
+        This works here cause the only database in transient is the index
```
bochecha: s/cause/because/

Oral slang doesn't look pretty in docstrings. :smile:

bochecha: I think you meant the only **table** in the transient db?

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
...
+        and it doesn't have 'real' relation.
```
bochecha: What's a "real" relation?

More generally, what is this sentence here for? I don't understand what it's trying to tell me. :confused:

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
...
 
     def allow_migrate(self, db, app_label, model_name=None, **hints):
         """
-        All database need to be migrated.
+        Allow migration only if the migration is applying to the right db.
         """
+        dbToUse = _databaseToUse.get((app_label, model_name), 'default')
+        if dbToUse != db:
+            return False
         return None
```
bochecha: Maybe return `True` explicitly here.

I mean, now that you dived into the darkness and explained to us that `None` ends up meaning `True`, I know it's the same thing.

But might that change in a future Django version?

And really, returning either `False` or `None` makes it a bizarre boolean.

ideascube/db_router.py
mgautierfr: > Why not simply get the model from the app_label and model_name?

I didn't get the model because we need a apps to use `apps.get_config()` and I think it was simpler to move the information here instead of found how to get the apps.
After investigation, it seems that get the apps is no complex at all (`from django.app import apps`)

However, the `apps` is the present apps and a model of migration may not exists at the time we apply the migration (the model name changed). If we attach the information to the model, we may loose the information.
We can default the database to use to 'default' and it will work most of the time (for now) but
it could lead to problem in the future (if we rename the search table in the transient db).

So I suggest to keep move with a separate dict of which db to use depending of a `app_label` and `model_name`

ideascube/db_router.py
mgautierfr: I disagree with you here :)

I don't see any reason to have two nested dictionaries.
And it forces us to have two lookup instead of one.

ideascube/db_router.py
mgautierfr: Yes, `_database_to_use` is better.

ideascube/db_router.py
```diff
@@ -1,30 +1,44 @@
...
+    ('search', 'search') : 'transient'
+}
+
 class DatabaseRouter(object):
     """
     A Database router.
     Some models do not need (and must not be backuped).
-    Those models will (must) have a _dbname class attribute set to 'transiant'.
+    So we route them to the transient base.
     This attribute will be use as database name.
```
mgautierfr: Right

ideascube/db_router.py
mgautierfr: "the read of the table". I will change it.

mgautierfr: Well it make sense to me. Hopefully because I wrote it. :)
If you think your version is more explicit, I don't see any reason to not change this sentence.

ideascube/db_router.py
mgautierfr: > I think you meant the only table in the transient db?

Right

ideascube/db_router.py
mgautierfr: `return None` means "This router takes no decision here".
This could be useful if we add another router to take decision based on another heuristics.

But we have only one router for now and we don't plan to add one in next future, so it's not so important.
As another commit already changes this method to potentially return True, I think it better to actually change the `return None` to `return True`

ideascube/search/migrations/0002_drop_old_index_table.py
```diff
@@ -14,5 +14,5 @@ class Migration(migrations.Migration):
     dependencies = [('search', '0001_initial')]
 
     operations = [
-        migrations.RunPython(migrate),
+        migrations.RunPython(migrate, hints={'using': 'default'}),
```
bochecha: This all looks good as far as the code is concerned.

Since there's nowhere to comment on the commit message I'll put it here:

> django doesn't know on witch

I think you meant "django doesn't know any witch". Or was it « which » ? 😜

Makefile
bochecha: Just to not leave this hanging: this is now done in the latest push to this branch.

ideascube/tests/test_models.py
```diff
@@ -61,36 +61,3 @@ def test_user_data_fields_should_return_labels_and_values(settings):
...
-        'list',
-        'dict',
-    ])
-def test_json_field(value):
-    obj = JSONModel(data=value)
-    obj.save()
-    assert JSONModel.objects.count() == 1
-
-    obj = JSONModel.objects.first()
-    assert obj.data == value
```
bochecha: Again, this is all code code-wise, but there's nowhere in Github to comment on the commit messages:

> Having a model speciajly

I think you meant "specially".

> the table is no created

Should be "the table is not created"

> It has been agree that

First, "agreed". But passive voice is too heavy here, just use "We agreed that"

ideascube/search/migrations/0001_initial.py
```diff
@@ -6,9 +6,10 @@
 
 
 class CreateSearchModel(migrations.CreateModel):
-    def database_forwards(self, *_):
-        # Don't run the parent method, we create the table our own way
-        create_index_table()
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
+        model = to_state.apps.get_model(app_label, self.name)
+        if self.allow_migrate_model(schema_editor.connection.alias, model):
+            create_index_table()
```
bochecha: Once more, commit message...

> we don't totally mastering

"master"

> I prefer to keep

"I prefer keeping"

ideascube/search/migrations/0001_initial.py
```diff
@@ -6,9 +6,10 @@
 
 
 class CreateSearchModel(migrations.CreateModel):
-    def database_forwards(self, *_):
-        # Don't run the parent method, we create the table our own way
```
bochecha: I wish you had kept the comment here, the goal was to explicitly state that we don't want anything `super()`.

ideascube/search/migrations/0001_initial.py
```diff
@@ -6,9 +6,10 @@
...
-        create_index_table()
+    def database_forwards(self, app_label, schema_editor, from_state, to_state):
+        model = to_state.apps.get_model(app_label, self.name)
+        if self.allow_migrate_model(schema_editor.connection.alias, model):
```
bochecha: Now I wonder... If we monkeypatched `schema_editor.create_model`, then we could use `super().database_forwards` couldn't we?

(no, that would not necessarily be a good idea, but it's been a long day)

Well done in any case here.

ideascube/db_router.py
```diff
@@ -1,30 +1,60 @@
...
     def allow_migrate(self, db, app_label, model_name=None, **hints):
         """
-        All database need to be migrated.
+        Allow migration only if the migration is applying to the right db.
         """
-        return None
+        dbToUse = _database_to_use.get((app_label, model_name), 'default')
+        if dbToUse != db:
+            return False
+        using = hints.get('using')
```
bochecha: For a nice symmetry with the previous branch, I'd have written:

```python
        using = hints.get('using', 'default')

        if using != db:
            return False
```

ideascube/library/migrations/0008_library_stock.py
```diff
@@ -163,7 +163,7 @@ class Migration(migrations.Migration):
             },
             bases=('monitoring.specimen', models.Model),
         ),
-        migrations.RunPython(migrate_data),
+        migrations.RunPython(migrate_data, hints={'using': 'default'}),
```
bochecha: Wasn't there a commit which made all the `RunPython` migrations reversible at some point in this branch?

ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -5,6 +5,13 @@
 
 
 def add_user(apps, *args):
+    # The apps.get_model() method returns us a 'fake' model with a
+    # default manager and the 'save' method is not overwrited
```
bochecha: "overwritten"

ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -5,6 +5,13 @@
...
+    # and the user is not indexed at creation.
+    # This is not a problem here as we have post-migrate hook who reindex
+    # everything at end of migration. Moreover, it is better has we do not
+    # have to care about if the 'transient' db is existing or not when doing
+    # this migration.
```
bochecha: I'm curious, what made you feel this comment was necessary?

To me it just says « something unexpected happens underneath us, but in the end everything's fine ».

ideascube/library/migrations/0008_library_stock.py
mgautierfr: A migration have to be reversible to allow migration tests to test this migration or migration before it.
As this migration already exists and we do not test it or one of this dependencies we do not need to make it reversible.

ideascube/migrations/0009_add_a_system_user.py
mgautierfr: > I'm curious, what made you feel this comment was necessary?

The time I spent to understand all this stuff and the pity I've got for the next developer who will try to understand it again.

mgautierfr:
Last `push --force` fix all the typos detected by @bochecha even if I didn't answer them individually.

PR #617, opened 2016-10-29
Add User.disabilities

### LINKED ISSUE #608: add User.disability
Possible values:
- visual
- auditive
- physical
- cognitive
- mental

Of course the field is nullable, and null by default.
Needed for Colombian project.

yohanboniface:
Fix #608


ideascube/migrations/0009_add_a_system_user.py
```diff
@@ -1,12 +1,11 @@
...
 from __future__ import unicode_literals
 
-from django.contrib.auth import get_user_model
 from django.db import migrations
 
 
-def add_user(*args):
-    User = get_user_model()
+def add_user(apps, *args):
+    User = apps.get_model('ideascube', 'User')
```
bochecha: Oops, good catch.

ideascube/models.py
```diff
@@ -244,6 +244,14 @@ def index_strings(self):
         ('female', _('Female')),
     )
 
+    DISABILITY_CHOICES = (
+        ('visual', _('Visual')),
+        ('auditive', _('Auditive')),
+        ('physical', _('Physical')),
+        ('cognitive', _('Cognitive')),
+        ('mental', _('Mental')),
```
bochecha: My only worry with this is whether we're not using the wrong categories here, or missing some.

Of course we can always add more in the future, but for someone in a specific situation, it can be painful to have your disability missing from a list of choices, as if you didn't exist. :disappointed: 

Of course, from the height of my abled priviledge, I have no idea whether this list is correct or not. So I'm not against, just worried that we're missing something important to someone, and hoping such an omission won't cause too much pain.

Or should this be a free-form `CharField` instead?

yohanboniface: I don't really know either. This list comes from Sergio, BSF project manager in Colombia.
They are supposed to be generic categories, not specific disabilities.
I suppose, the categories may vary according to the specialist answering, and I'm not sure we have a better person than Sergio to arbitrate.
The immediate need is to match with Llave del saber (which does not know about "mental", by the way; this has been added by Sergio). I think we are good to go, given we can still add more in the future.

bochecha: Searching a bit, "intellectual disability" seems to be the preferred term in English, over "cognitive disability" and "mental disability" being the least preferred one.

These 3 seem to mostly be the same though, so maybe that's why Llave del saber didn't have the "mental" one?

In addition, the various categorizations seem to vary from one region to another, with sometimes "cognitive disability" being the same as "intellectual disability", but sometimes the former is a superset of the latter.

The more I research about this, the more I'm leaning towards a free-form `CharField`. It removes any assumption we could make about people with disabilities (assumptions are bad), and it makes sure we get it right by not excluding anybody.

It also removes the need to make annoying migrations in the future. For example, using a fixed set of categories, say a deployment wanted to split a category in two, we wouldn't be able to reassign people's disability to one or the other during the migration.

barbayellow: This is a specific need for this project. In BSF' context, Yohan's list in #608 is exhaustive enough. If we want to be sure not to leave anyone on the side, adding a "other" might be an acceptable alternative.

bochecha:
I still think a fixed list is a bad idea here.

But since the majority opinion goes towards a list, and since the code correctly implements that, I'm going to merge it.

Issue #598, opened 2016-10-26
Add colors to book sections


yohanboniface:
Book sections are used to organize books in the physical library.
Each section should have it's own color, so the librarian can add colored stickers on the books with same colors.
What we want is to display the color near the section name in the dropdown select and when seeing the book page.

Fun fact: this is a request from the Colombian project librarians, but we used to have this in the very first version of the ideasbox.

Colors:
- novels: #4183D7
- poetry: #9B59B6
- theatre: #6C7A89
- documentary, digital, other, game: white
- comics: #D91E18
- cartoons: #e9d460
- myths and legends: #F27935
- tales: #4DAF7C
- graphic novels: #674172

mgautierfr:
Those colors do not correspond to the color in #597.
Which one are the best ?

yohanboniface:
The ones here.

PR #619, opened 2016-11-03
Library sections

### LINKED ISSUE #597: New book sections
We want new book sections:
- children - myths and legends
- children - tales
- adult - myths and legends
- adult - tales

mgautierfr:
fix #597 and #598.

This PR is depending of #616 

The two last commits of this PR are optional.
They are useful but may be discussed separately.
I add them in this PR by laziness but if there is no consensus on them, I will move them on another PR to no block this one.


ideascube/tests/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+from django.db.migrations.executor import MigrationExecutor
+from django.db import connection
+import pytest
+
+needMigration = pytest.mark.skipif(
+    pytest.config.getoption("--nomigrations"),
+    reason="need migrations to run tests on migrations"
+)
+
+@needMigration
```
bochecha: Since you're using the decorator only once, why introduce this intermediary variable instead of directly using the `pytest.mark.skipif`?

Also... camelCase? o_O

ideascube/library/tests/test_migrations.py
```diff
@@ -0,0 +1,35 @@
...
+               11: 'children-theatre',
+               4: 'children-documentary',
+               5: 'children-comics',
+               6: 'adults-novels',
+               7: 'adults-documentary',
+               8: 'adults-comics',
+               12: 'adults-poetry',
+               13: 'adults-theatre',
+               9: 'game',
+               99: 'OTHER'}
```
bochecha: Indentation?

Generally in other places we use:

```python
meh = {
    1: ...,
    2: ...,
}
```

bochecha: You're duplicating the values from the migration file, here. Any way to DRY-ify this a bit?

ideascube/library/tests/test_migrations.py
```diff
@@ -0,0 +1,35 @@
...
+
+class SectionTypeMigration(MigrationTestCase):
+    migrate_from = [('library', '0009_auto_20161027_0801')]
+    migrate_to   = [('library', '0010_section_type')]
+
+
```
bochecha: One newline too many.

ideascube/library/tests/test_migrations.py
```diff
@@ -0,0 +1,35 @@
...
+        Book = self.old_apps.get_model('library', 'Book')
+        for i in range(14):
+            #There is no section 0, use 99 instead
+            i = i or 99
+            Book.objects.create(name="{}".format(i), section=i)
+
+    def test_section_migration_change_from_int_to_string(self):
+        Book = self.new_apps.get_model('library', 'Book')
+        for book in Book.objects.all():
+            old_section = int(book.name)
```
bochecha: Clever. =]

ideascube/library/tests/test_migrations.py
```diff
@@ -0,0 +1,35 @@
...
+    migrate_from = [('library', '0009_auto_20161027_0801')]
+    migrate_to   = [('library', '0010_section_type')]
+
+
+    def setUpDataBeforeMigration(self):
+        Book = self.old_apps.get_model('library', 'Book')
+        for i in range(14):
+            #There is no section 0, use 99 instead
+            i = i or 99
+            Book.objects.create(name="{}".format(i), section=i)
```
bochecha: `str(i)` ?

ideascube/tests/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+
+        self.executor = MigrationExecutor(connection)
+        self.executor.loader.project_state(self.migrate_from).apps
+        self.executor.loader.build_graph()  # reload.
+        self.executor.migrate(self.migrate_from)
+
+        self.setUpDataBeforeMigration()
+
+        self.executor.loader.build_graph()  # reload.
+        self.executor.migrate(self.migrate_to)
```
bochecha: So if I get this right, the way this works is that if I want to test a migration I would do this:

```python
class SomeTest(MigrationTestCase):

    migrate_from = ...
    migrate_to = ...

    def setUpDataBeforeMigration(self):
        # Add some data
        ...

    def test_something(self):
        # Just check that the migration went fine
        ...
```

But from what I understand, if I want to run multiple tests on the same migration, with different pre-loaded data, I'd have to do:

```python
class SomeTest(MigrationTestCase):

    migrate_from = ...
    migrate_to = ...

    def setUpDataBeforeMigration(self):
        # Add some data
        ...

    def test_something(self):
        # Just check that the migration went fine
        ...


class SomeOtherTest(MigrationTestCase):

    migrate_from = ...  # Same as previous class
    migrate_to = ...  # Same as previous class

    def setUpDataBeforeMigration(self):
        # Add some different data
        ...

    def test_something(self):
        # Just check that the migration went fine
        ...


...
```

That is, once test case class per test.

I'd prefer being able to write only one test case class for **all the tests on a given migration**:

```python
class MigrationXXXTestCase(MigrationtestCase):

    migrate_from = ...
    migrate_to = ...

    def test_something(self):
        # Set up the data here
        ...

        # Run the migration here
        ...

    def test_something_else(self):
        ...

    def test_a_third_thing(self):
        ...
```

That does imply a slight change on the API exposed by your `MigrationTestCase` class:

* the `setupDataBeforeMigration` method becomes unnecessary, each test function is responsible for loading its data;
* you need a new `run_migration` method which each test function can call (it would probably just run the last two `build_graph` and `migrate` lines)

What do you think?

ideascube/library/migrations/0010_section_type.py
```diff
@@ -0,0 +1,48 @@
...
+               11: 'children-theatre',
+               4: 'children-documentary',
+               5: 'children-comics',
+               6: 'adults-novels',
+               7: 'adults-documentary',
+               8: 'adults-comics',
+               12: 'adults-poetry',
+               13: 'adults-theatre',
+               9: 'game',
+               99: 'OTHER'}
```
mgautierfr: This is the same order that the one in the model.
We may reorder the one in the model also.
However, field's choice in the model will evolve (as new sections are added) and not this one.
I don't have special idea on this point.

ideascube/library/migrations/0010_section_type.py
```diff
@@ -0,0 +1,48 @@
...
+
+    dependencies = [
+        ('library', '0009_auto_20161027_0801'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='book',
+            name='section_named',
+            field=models.CharField(choices=[('digital', 'digital'), ('children-cartoons', 'children - cartoons'), ('children-novels', 'children - novels'), ('children-poetry', 'children - poetry'), ('children-theatre', 'children - theatre'), ('children-documentary', 'children - documentary'), ('children-comics', 'children - comics'), ('adults-novels', 'adults - novels'), ('adults-poetry', 'adults - poetry'), ('adults-theatre', 'adults - theatre'), ('adults-documentary', 'adults - documentary'), ('adults-comics', 'adults - comics'), ('game', 'game'), ('OTHER', 'other')], default='OTHER', max_length=50, verbose_name='section'),
```
mgautierfr: Hum, there should be no default at all for section (there was no default before)
I add a default here cause we need something to put in this field during the migration.

So, a `preserve_default=False` is missing in the `AddField` call and I will remove the default in the model definition.

ideascube/library/models.py
```diff
@@ -18,22 +18,22 @@ def available(self):
 
 class Book(StockItem, SearchMixin, TimeStampedModel):
 
-    OTHER = 99
+    OTHER = 'OTHER'
```
mgautierfr: Good catch. I will change it to 'other'

ideascube/library/models.py
```diff
@@ -49,7 +53,7 @@ class Book(StockItem, SearchMixin, TimeStampedModel):
     publisher = models.CharField(verbose_name=_('publisher'), max_length=100,
                                  blank=True)
     section = models.CharField(verbose_name=_('section'), max_length=50,
-                               choices=SECTION_CHOICES)
+                               choices=SECTION_CHOICES, default=OTHER)
```
mgautierfr: > What's going on?

Mess :)

I will remove it.

ideascube/library/static/library/library.css
```diff
@@ -39,4 +39,145 @@ img.cover {
...
+  padding: 0 5px;
+  color: #fff;
+}
+
+.card.children-myths,
+.card.adults-myths,
+.cover.children-myths,
+.cover.adults-myths {
+  border: 4px solid #F27935;
+}
```
mgautierfr: There are white.

I consider the white as the default and I do not explicit the color here.

ideascube/tests/helpers.py
mgautierfr: > What do you think?

You're right :)

I will push a new version with your API.

ideascube/library/tests/test_migrations.py
mgautierfr: Those values are mostly the heart of the migration. If we reuse the migration data here, I'm fear we are testing almost nothing. (Even if, I concede, those values are mainly copy/paste)

ideascube/library/migrations/0010_section_type.py
bochecha: Indeed, I didn't know about `preserve_default=False`, but that seems to address exactly what I was worried about.

ideascube/library/static/library/library.css
bochecha: Sure, but are they supposed to be white?

mgautierfr: Yes, at least from (#598)

ideascube/library/migrations/0010_section_type.py
```diff
@@ -40,7 +40,11 @@ class Migration(migrations.Migration):
             field=models.CharField(choices=[('digital', 'digital'), ('children-cartoons', 'children - cartoons'), ('children-novels', 'children - novels'), ('children-poetry', 'children - poetry'), ('children-theatre', 'children - theatre'), ('children-documentary', 'children - documentary'), ('children-comics', 'children - comics'), ('adults-novels', 'adults - novels'), ('adults-poetry', 'adults - poetry'), ('adults-theatre', 'adults - theatre'), ('adults-documentary', 'adults - documentary'), ('adults-comics', 'adults - comics'), ('game', 'game'), ('other', 'other')], default='other', max_length=50, verbose_name='section'),
             preserve_default=False
         ),
-        migrations.RunPython(migrate_data),
+        # We need to make this migration revertible for tests.
+        # We do not support real backward migration, test will revert this
+        # migration on an empty database, so we do not have to do a real
+        # backward data migration. So just use a noop a backward migration.
+        migrations.RunPython(migrate_data, migrations.RunPython.noop),
```
bochecha: I wouldn't bother with the comment here.

Explicitly making all migrations backwards, even if there's no actual backward operation and it's just a noop seems like a good thing to do anyway.

ideascube/library/tests/test_migrations.py
bochecha: > If we reuse the migration data here, I'm fear we are testing almost nothing.

I disagree.

Look at #626 for example, where we realized the migration wasn't good. I'm going to reuse your helper to test things with various cases of data loaded into the db before the migration.

To me, the interesting part of the test is the data loaded into the db before the test runs the migration. That's what you want to verify after the migration. How the migration achieved that end result doesn't matter (from the point of view of the tests), as long as for all the cases you test the end data is good.

As such, if you feel you're « testing almost nothing » with the imported mapping, then you should feel the same if you're copy-pasting the mapping into the test. 😉 

But that doesn't matter much, I'm not going to block on this, I was merely wondering.

ideascube/library/static/library/library.css
```diff
@@ -39,4 +39,145 @@ img.cover {
...
+
+
+
+/* ************************************************* */
+/* ************ SECTION COLORS ********************* */
+/* ************************************************* */
+
+/************* Cartoons *************/
+#id_section > option[value=children-cartoons] {
```
yohanboniface: I'd rather put a small square of the given color close to the option text, it's a bit flashy as it is ;)

ideascube/tests/helpers.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db.migrations.executor import MigrationExecutor
+from django.db import connection
+import pytest
+
+
+@pytest.mark.skipif(
+    pytest.config.getoption("--nomigrations"),
+    reason="skipping due to --nomigrations"
+)
+class MigrationTestCase(TransactionTestCase):
```
yohanboniface: I'd have used pytest fixture instead of going for unittest classes here, see http://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code for managing setup/teardown (more readable and maintainable *imho*, and certainly more consistent with the way we usually do in other tests).

Is there a reason why you didn't go that way?

ideascube/library/static/library/library.css
mgautierfr: I'm not sure we can put html node into a option.
Maybe it can work with a :after/:before css rule.
I'll try.

ideascube/tests/helpers.py
mgautierfr: I don't know how to create py.test fixtures parametrized by the value known at test definition.
Fixtures can be parametrized or use another fixtures but this is defined at fixture creation.

The `migrate_from` and `migrate_to` are relative to test, and so it's not possible to create a generic fixture here. (And I agree, it's a pity :/ )

yohanboniface: I think that's something you can do with request fixture and indirect params: http://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

ideascube/tests/helpers.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+
+def migration_test(migrate_from, migrate_to):
```
bochecha: That decorator looks like a scary beast to maintain in the long run.

However, the API it provides (decorate test functions) is pretty great!

ideascube/library/static/library/library.css
bochecha: To avoid leaving this thread hanging, I have verified that this was done in the latest push to this branch. (and it looks beautiful)

bochecha:
Still approving this. I especially love the new API provided by that (hairy) decorator for migration tests.

Can't wait to use it in #626 !

ideascube/library/tests/test_migrations.py
```diff
@@ -0,0 +1,37 @@
...
+            #There is no section 0, use 99 instead
+            i = i or 99
+            Book.objects.create(name=str(i), section=i)
+
+        migration.run_migration()
+
+        Book = migration.new_apps.get_model('library', 'Book')
+        for book in Book.objects.all():
+            old_section = int(book.name)
+            assert book.section == section_int_to_name[old_section]
```
bochecha: Care to fix the indentation before we merge, though? :wink:

mgautierfr: Good catch, indentation is from when the code was using Class API.

PR #621, opened 2016-11-03
Section search

### LINKED ISSUE #599: Add "section" filter in library
This means also indexing it, but we don't have a column for that.

mgautierfr:
This PR is base on the PR #619 as we need the new section field.

Fix #599.

### DIFF
--- a/ideascube/library/tests/test_views.py
+++ b/ideascube/library/tests/test_views.py
@@ -409,6 +409,14 @@ def test_by_tag_page_should_be_filtered_by_tag(app):
     assert boat.item.name not in response.content.decode()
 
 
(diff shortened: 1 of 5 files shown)


ideascube/library/tests/test_views.py
```diff
@@ -409,6 +409,14 @@ def test_by_tag_page_should_be_filtered_by_tag(app):
     assert boat.item.name not in response.content.decode()
 
 
+def test_by_tag_page_should_be_filtered_by_section_kind(app):
```
bochecha: `test_by_kind_...` ?

bochecha:
My minor nitpick is fixed, and it's the only change since I had approved this pull request.

PR #622, opened 2016-11-04
Add a warning to staff user if there is no specimen for a book.

### LINKED ISSUE #573: Add a warning when a book has no specimen
When a book has no specimen associated with, it does not appear in the library application for the users. Which makes sense. However, it appears in the library application when for the staff. It's confusing. 

We should add a warning for the staff when a book has no specimen : "Please add a specimen, or the book won't be available for the users".

mgautierfr:
fix #573

### DIFF
--- a/ideascube/library/tests/test_views.py
+++ b/ideascube/library/tests/test_views.py
@@ -24,6 +24,12 @@ def test_only_books_with_specimen_should_be_in_index(app, book, specimen):
     assert book.name not in response.content.decode()
 
 
+def test_all_books_should_be_in_index_for_staff(staffapp, book, specimen):
+    response = staffapp.get(reverse('library:index'))
+    assert specimen.item.name in response.content.decode()
+    assert book.name in response.content.decode()
+
+
 def test_index_page_is_paginated(app, monkeypatch):
     monkeypatch.setattr(Index, 'paginate_by', 2)
     BookSpecimenFactory.create_batch(size=4)
@@ -106,6 +112,28 @@ def test_book_detail_page_does_not_list_specimens_to_non_staff(app, book):
     DigitalBookSpecimenFactory(item=book, file__filename='book.epub')
     resp = app.get(reverse('library:book_detail', kwargs={'pk': book.pk}))
     resp.mustcontain(no=[specimen.barcode])
+
+
+def test_book_detail_page_shows_warning_message_to_staff_if_no_specimens(staffapp, book):
+    resp = staffapp.get(reverse('library:book_detail', kwargs={'pk': book.pk}))
+    resp.mustcontain("Please add a specimen, or the book won&#39;t be available for the users")
+
+
+def test_book_detail_page_does_not_show_warning_message_to_user_if_no_specimens(app, book):
+    resp = app.get(reverse('library:book_detail', kwargs={'pk': book.pk}))
+    resp.mustcontain(no="Please add a specimen, or the book won&#39;t be available for the users")
+
+
+def test_book_detail_page_does_not_show_warning_message_if_specimens(staffapp, book):
+    BookSpecimenFactory(item=book)
+    resp = staffapp.get(reverse('library:book_detail', kwargs={'pk': book.pk}))
+    resp.mustcontain(no="Please add a specimen, or the book won&#39;t be available for the users")
+
+
+def test_book_detail_page_does_not_show_warning_message_if_digital_specimens(staffapp, book):
+    DigitalBookSpecimenFactory(item=book)
+    resp = staffapp.get(reverse('library:book_detail', kwargs={'pk': book.pk}))
+    resp.mustcontain(no="Please add a specimen, or the book won&#39;t be available for the users")
 
 
 def test_book_detail_page_list_physical_specimens_to_staff(staffapp, book):
--- a/ideascube/library/views.py
+++ b/ideascube/library/views.py
@@ -66,6 +66,14 @@ index = Index.as_view()
 
 class BookDetail(DetailView):
     model = Book
+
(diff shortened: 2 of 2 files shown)


ideascube/library/views.py
```diff
@@ -66,6 +66,13 @@ def get_context_data(self, **kwargs):
 
 class BookDetail(DetailView):
     model = Book
+    def get(self, request, *args, **kwargs):
+        ret = super().get(request, *args, **kwargs)
+        user = request.user
+        if user.is_authenticated() and user.is_staff and not self.object.physical:
```
yohanboniface: `not self.object.physical` will add the warning even for a book that has only digital specimens, which is not what we want AFAIK.

bochecha:
The way I understand #573 is that the warning should be on the page listing all books, since that's the page showing differences for staff and non-staff users.

However, having it on the book-detail page is certainly a good idea as well, since that's the page a staff admin sees right after creating a book.

Maybe add it on both?

ideascube/library/tests/test_views.py
```diff
@@ -24,6 +24,12 @@ def test_only_books_with_specimen_should_be_in_index(app, book, specimen):
     assert book.name not in response.content.decode()
 
 
+def test_all_books_should_be_in_index_for_staff(staffapp, book, specimen):
+    response = staffapp.get(reverse('library:index'))
+    assert specimen.item.name in response.content.decode()
+    assert book.name in response.content.decode()
```
bochecha: Good catch, this was an important missing test.

ideascube/library/views.py
```diff
@@ -66,6 +66,13 @@ def get_context_data(self, **kwargs):
...
+            messages.warning(request, _("Please add a specimen, or the book "
+                                        "won't be available for the users"))
```
bochecha: I wouldn't have used `messages.warning(...)` here, I'd have put the message right where the book specimens would normally be listed.

yohanboniface: I think a warning is nice: it's already the way we display a message to the user. No need to invent a new one imho.

yohanboniface:
> The way I understand #573 is that the warning should be on the page listing all books, since that's the page showing differences for staff and non-staff users.

I think the idea was to have a specimen counter on the list (from recent discussion, not from the issue description).

ideascube/library/views.py
mgautierfr: You're right. Changed in last push force

mgautierfr:
>> The way I understand #573 is that the warning should be on the page listing all books, since that's the page showing differences for staff and non-staff users.

> I think the idea was to have a specimen counter on the list (from recent discussion, not from the issue description).

Yes, and it is not easy to provide a smart warning on the index pages. If there is a lot of book, the user will have difficulties to find the book without specimens.

We could add a specimen counter for each book but I don't see a how to integrate it nicely on the UI side. Maybe change the background of the card for book without specimen ?

bochecha:
> I think the idea was to have a specimen counter on the list (from recent discussion, not from the issue description).

Indeed, that would work.

ideascube/library/tests/test_views.py
```diff
@@ -108,6 +114,26 @@ def test_book_detail_page_does_not_list_specimens_to_non_staff(app, book):
     resp.mustcontain(no=[specimen.barcode])
 
 
+def test_book_detail_page_shows_warning_message_to_staff_if_no_specimens(staffapp, book):
+    resp = staffapp.get(reverse('library:book_detail', kwargs={'pk': book.pk}))
+    resp.mustcontain("Please add a specimen, or the book won&#39;t be available for the users")
+
+
+def test_book_detail_page_does_not_show_warning_message_to_user_if_no_specimens(app, book):
```
bochecha: So users can reach the book page, even though they can't see the book in the list?

Is that what we want? Or should the detail page for a book without specimens instead be a 404 if the user isn't staff?

ideascube/library/views.py
```diff
@@ -66,6 +66,15 @@ def get_context_data(self, **kwargs):
 
 class BookDetail(DetailView):
     model = Book
+    def get(self, request, *args, **kwargs):
+        ret = super().get(request, *args, **kwargs)
+        user = request.user
+        if ( user.is_authenticated()
+         and user.is_staff
```
bochecha: `user.is_staff` implies `user.is_authenticated()`, so you don't need to test both.

ideascube/library/tests/test_views.py
mgautierfr: Test `test_everyone_should_access_book_detail_page` makes me think that it's what we want.

But I agree it could be changed. (For a 403 ? As the book is existing but we don't want to show it to user)

@yohanboniface, what do you think about this ?

ideascube/library/views.py
mgautierfr: Right.
Fixed in last push --force

ideascube/library/tests/test_views.py
yohanboniface: Not a big deal imho.
We do not show the books without specimen in the list to normal users, because we don't want to polute their list, and we do not want them to think they can read those books, because we actually don't have them.
But if they manage to get on the detail page anyway, that's not a "problem" (even if this is something that should not happen, i.e. we should not have links pointing here).

ideascube/library/views.py
```diff
@@ -66,6 +66,14 @@ def get_context_data(self, **kwargs):
 
 class BookDetail(DetailView):
     model = Book
+    def get(self, request, *args, **kwargs):
+        ret = super().get(request, *args, **kwargs)
+        user = request.user
+        if ( user.is_staff
```
bochecha: Space after the parentheses?

PR #624, opened 2016-11-08
Clean kb conf

yohanboniface:
This simplifies the KB confs.
The target is to use only generic confs (kb, kb_bsfcampus, maybe one or two more) as much as possible.
Some confs still remain because of the IDEASCUBE_NAME, that needs to be updated before (through Ansiblecube). That will be another step.
Plus, we may tackle in another thread the LANGUAGE_CODE (using the server lang?), for now we can proceed, as it's not used for the UI (only for scripts and such; on the UI we default on browser language).

@bochecha @mgautierfr one of you can have a quick safety look? I'd like to release tomorrow morning before going to work :)

Thankss!

### DIFF
--- a/ideascube/conf/kb.py
+++ b/ideascube/conf/kb.py
@@ -22,7 +22,4 @@ HOME_CARDS = STAFF_HOME_CARDS + [
     {
         'id': 'khanacademy',
     },
-    {
-        'id': 'software',
-    }
 ]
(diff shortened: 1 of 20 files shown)


yohanboniface:
Note: this has been done together with @fheslouin when I was in the office today :)

bochecha:
A bit late to the party, but apart from my minor comment, this looks like it should work.

PR #627, opened 2016-11-15
add vhost in from of requests in access_log

mbaechler:
this needs real-life testing, can I let you try on your ideasbox ?

### DIFF
--- a/extras/nginx/ideascube
+++ b/extras/nginx/ideascube
@@ -3,6 +3,13 @@ upstream ideascube {
     server unix:///tmp/ideascube.sock;
 }
 
+#add server_name at line start
+log_format vhost_combined '$server_name $remote_addr - $remote_user [$time_local] '
+                          '"$request" $status $body_bytes_sent '
+                          '"$http_referer" "$http_user_agent"';
+                    
+access_log /var/log/nginx/access.log vhost_combined;
+                    
 # configuration of the server
 server {
     # the port your site will be served on


extras/nginx/ideascube
```diff
@@ -3,6 +3,13 @@ upstream ideascube {
     server unix:///tmp/ideascube.sock;
 }
 
+#add host in from of request to track server_name in logs
```
bochecha: I don't understand this comment.

extras/nginx/ideascube
```diff
@@ -14,6 +21,8 @@ server {
     server_name ideasbox.lan www.ideasbox.lan ideasbox.local www.ideasbox.local koombook.lan www.koombook.lan koombook.local www.koombook.local bibliotecamovil.lan www.bibliotecamovil.lan;
     charset     utf-8;
 
+    
+
```
bochecha: Why those new lines?

mbaechler: it's noise, I first expected to put configuration there and forgot to remove them.

extras/nginx/ideascube
bochecha: Ah, that makes more sense now, thanks.

Can you merge both commits?

letompouce:
Thinking a bit more about it, we want this log format for all logs, so it should be configured right into the main nginx configuration file and not only for this vhost. Somewhere in https://github.com/ideascube/ansiblecube/blob/oneUpdateFile/roles/nginx/files/nginx.conf

mbaechler:
of course, feel free to close this PR and do the modification wherever it best fits.

Toniob:
I think you should use "$host" instead of "$server_name" in the log format

letompouce:
Fixed by https://github.com/ideascube/ansiblecube/commit/9b046ee10490fa9696427a74018dc75a813f9cbc

PR #628, opened 2016-11-17
Upgrade our deps

bochecha:


bochecha:
WTH? Why was this closed?

PR #629, opened 2016-11-21
Import and export book specimens

### LINKED ISSUE #595: Import book specimens
In the project "biblioteca movil", we'll have 20 box that will share both the book metadatas AND the specimen bar codes.
Now it's only possible to export/import metadatas.

We need to add both an export of specimens and an import.

When importing, specimens will be attached to existing book matching the original title (`name`) and ISBN (that can be empty).

When existing, we need to export the specimen files (epub, pdf…).

The format should be a zip containing a CSV for the metadatas, plus any optional linked files.

bochecha:
With quite a bit of refactoring of the `CSVExportMixin` first.

Fixes #595


yohanboniface:
Travis is complaining, you've seen that? :)

ideascube/library/views.py
```diff
@@ -184,3 +184,34 @@ def get_row(self, book):
...
+
+    def get_headers(self):
+        return self.fields
+
+    def get_row(self, bookspecimen):
+        row = {}
+
+        for field in self.fields:
+            if field == 'item':
+                row[field] = bookspecimen.item.name
```
yohanboniface: Having just "name" to match the book may generate some issues, I'm tempted to add also the book ISBN (that may be null) to reduce the problematic cases.

ideascube/library/forms.py
```diff
@@ -34,6 +38,93 @@ class Meta:
...
+                            'Missing column "{}" on line {}').format(
+                                e.args[0], index + 1))
+                        continue
+
+                    # Get the related book item
+                    try:
+                        title = data.pop('item')
+                        data['item'] = Book.objects.get(name=title).pk
+
+                    except Book.DoesNotExist:
```
yohanboniface: We should also catch "MultipleObject".

ideascube/library/templates/library/specimen_import.html
```diff
@@ -0,0 +1,23 @@
...
+{% block content %}
+
+    <div class="row">
+        <h2>{% trans "Import Book Specimens" %}</h2>
+        {% if form.errors %}
+            <div class="error">{{ form.errors }}</div>
+        {% endif %}
+    </div>
+
+    <div class="row">
```
yohanboniface: Do we still have "row" class, now that we use purecss?

ideascube/library/templates/library/specimen_import.html
```diff
@@ -0,0 +1,23 @@
...
+        <form method="POST" id="import"  enctype="multipart/form-data">
```
yohanboniface: Two spaces before "enctype" :p

ideascube/library/tests/test_views.py
```diff
@@ -514,3 +514,88 @@ def test_export_book_specimens(staffapp, monkeypatch):
...
+    book2 = BookFactory(isbn='234567', name='ﺎﻠﻨﺒﻳ (ﻚﺗﺎﺑ)')
+
+    zip_path = tmpdir.join('specimens.zip')
+    csv_content = '\n'.join([
+        'item,barcode,serial,comments,location,file',
+        'my book title,1234,,,,',
+        'my book title,,,,,1.epub',  # file is missing from the archive
+        'ﺎﻠﻨﺒﻳ (ﻚﺗﺎﺑ),,,,,2.epub',
+        'ﺎﻠﻨﺒﻳ (ﻚﺗﺎﺑ),1234,,,,',     # this barcode already exists
+        'no such book,2345,,,,',     # no item with this title
```
yohanboniface: I'd add a case with two books matching: twice same title, no ISBN.

ideascube/library/tests/test_views.py
```diff
@@ -514,3 +514,88 @@ def test_export_book_specimens(staffapp, monkeypatch):
...
+    assert specimen.item.pk == book2.pk
+    assert specimen.barcode is None
+    assert specimen.serial is None
+    assert specimen.comments == ''
+    assert specimen.location == ''
+    assert specimen.file.read() == b'Trust me, this is an epub'
+
+    # Errors
+    assert (
+        'Could not import line 2: file 1.epub is missing from the archive'
```
yohanboniface: Are we sure the pk "1" is reliable?

ideascube/library/forms.py
bochecha: Wait, `name` does not have a unicity constraint? 😮 

How is that going to work then?

ideascube/library/tests/test_views.py
bochecha: I don't see why it wouldn't be.

This isn't even a private key at all, it is just the file name in the tarball.

It just so happens that when **exporting** the file is called after the private key of the exported model instance.

But when **importing**, the new model instance will have its own private key (if successfully imported), which might or might not end up being the same integer as the one used for the file name.

(but if it were a private key, then yes, it would be reliable, since the test gets a clean database with no specimens in it, the first one will therefore be `1`)

ideascube/library/views.py
bochecha: Must we match **both** name and ISBN to consider that we found the right `Book`?

yohanboniface: I was about to say "Yes!", but then I thought: what if we find the same ISBN but with a different title?
We already know different books share one ISBN, so this can happen. But more realistically if the title is somewhat different (case, missing "the", typo…) ?
I'd say:
- if we have an ISBN, just match on it, and we can't handle the case of different books sharing the ISBN, that will be a human responsibility (human creates the shit, human fixes it, no? :p)
- else match on the name, and if we have more than one, raise an error, human matter again

bochecha: > We already know different books share one ISBN, so this can happen.

Not with our model:

```python
    # We allow ISBN to be null, but when it is set it needs to be unique.
    isbn = models.CharField(verbose_name=_('isbn'), max_length=40,
                            unique=True, null=True, blank=True)
```

Should I fix that too?

bochecha: I don't understand how you can conclude this:

> Nope, ISBN definitely needs to be unique.

If then you accept this:

> I meant in real life, we know that it happens that two different books share the same ISBN.

yohanboniface: > I meant in real life

Real life, like out of the context of our servers :)

bochecha: Oh, ok, I thought our servers were real? 😕 

I'm now extremely confused.

ideascube/library/templates/library/specimen_import.html
bochecha: Since I'm creating this new template, I'd rather create it the right new way.

So what's the right new way now with PureCSS to do what `row` used to do?

ideascube/library/templates/library/specimen_import.html
bochecha: Fixed, good eye.

ideascube/library/tests/test_views.py
bochecha: I've added quite a few more tests in the latest push to this branch.

bochecha:
Since @yohanboniface is unfortunately unavailable these days, let me try and summarize where we are with this.

I'm trying to write this as neutrally as possible, in the hope that I won't misrepresent or belittle @yohanboniface's view.

---

@yohanboniface said on IRC he thinks this can be merged, after the fixes I made since his last review.

I disagree because it considers ISBNs are unique. They (unfortunately) aren't, and this pull request would be a good occasion to get back on this decision and fix it. Instead, we enshrine that decision by making even more things depend on it, and it might end up being harder to fix it later on.

However, since we are pressed on time and need to merge this, I created #645 to track the ISBN reuse issue.

I continue thinking it would have been preferable to fix the issue before merging this, but I'll accept if I don't "win" this argument.

With all that said, feel free to merge this @mgautierfr

mgautierfr:
If ISBN are not unique, then I think we should handler it, this is a problem we will encounter. (Especially if we start to use ideascube to handle "real" libraries)

However, this PR seems already big enough to not try to add an "unrelated" fix.
This PR is to add import and export of book specimens and it does it pretty well (and more thank to the @bochecha CSV mixin refactoring).

We need to fix #645, but as this is a pretty rare case (hopefully) and the fix is not so easy to done.
Moreover this feature is needed for this release, so I will merge this PR now.

PR #632, opened 2016-11-23
Kb rwa giz

fheslouin:
Can you review and merge ? 

Thanks

### DIFF
--- a/ideascube/conf/kb_rwa_giz.py
+++ b/ideascube/conf/kb_rwa_giz.py
@@ -1,0 +1,13 @@
+"""KoomBook conf"""
+from .kb import *  # noqa
+
+LANGUAGE_CODE = 'en'
+IDEASCUBE_NAME = 'GIZ'
+HOME_CARDS = STAFF_HOME_CARDS + [
+    {
+        'id': 'blog',
+    },
+    {
+        'id': 'mediacenter',
+    },
+]


yohanboniface:
Seems the statics in the package are still loaded from the original domain. Here is what I have when I test it:

    <link rel="stylesheet" type="text/css" media="all" href="http://www.24hdansuneredaction.com/wp-content/themes/24hdansuneredaction/base.css">

ideascube/conf/kb_rwa_giz.py
```diff
@@ -0,0 +1,13 @@
+"""KoomBook conf"""
+from .kb import *  # noqa
+
+LANGUAGE_CODE = 'en'
+IDEASCUBE_NAME = 'GIZ'
```
bochecha: Did you know we don't need to specify the `IDEASCUBE_NAME` in the settings files any more? 😄

fheslouin: Ok, but what is the default name then ?

fheslouin: I'm using set server name currently in the playbook, however the name picked up to set the var is the project name which is not really nice...I'll remove IDEASCUBE_NAME

People will have to find their way to the admin page to change the displayed name...

bochecha: > I'm using set server name currently in the playbook, however the name picked up to set the var is the project name which is not really nice...

Well, but that could be made nicer with a separate variable, set to a prettier name, no?

fheslouin: Yep but we are not always aware of the name wished by the end-user :/
See #634

PR #640, opened 2016-11-28
Workaround the bad Koombook battery driver

### LINKED ISSUE #446: Update KoomBook battery page
Can we update the battery info page for the KoomBook
- Power connected or disctonnected 
  - `/sys/power/axp_pmu/ac/connected` (Return 0 or 1)
- Battery charging or not 
  - `/sys/power/axp_pmu/battery/charging` (Return 0 or 1)
- Battery connected to the board 
  - `/sys/power/axp_pmu/battery/connected`  (Return 0 or 1)
- Battery capacity 
  - `/sys/power/axp_pmu/battery/capacity` (in %)

We can get info from the battery charger as well, it can be interesting for admin purpose : 
- Power Management Unit Voltage :
  - `/sys/power/axp_pmu/pmu/voltage` (in Volt)
- Power Management Unit temperature :
  - `/sys/power/axp_pmu/pmu/temp` (in °C)

bochecha:
This is pretty ugly, but the Koombook battery driver doesn't respect the
Linux convention for reporting the power status to sysfs.

As a result, batinfo doesn't work with it.

Since this is quite specific, I don't think it makes sense to push that
upstream to batinfo, so I opted for a workaround in our code.

Fixes #446

### DIFF
--- a/ideascube/serveradmin/views.py
+++ b/ideascube/serveradmin/views.py
@@ -1,6 +1,5 @@
 from subprocess import call
 
-import batinfo
 from django.conf import settings
 from django.contrib import messages
 from django.http import StreamingHttpResponse
@@ -12,6 +11,7 @@ from ideascube.decorators import staff_member_required
 from ideascube.utils import get_all_languages
 
 from .backup import Backup
+from .battery import get_batteries
 from .systemd import Manager, NoSuchUnit, UnitManagementError
 from .wifi import (
     AvailableWifiNetwork, KnownWifiConnection, enable_wifi, WifiError)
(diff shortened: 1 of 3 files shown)


ideascube/serveradmin/views.py
```diff
@@ -166,7 +167,7 @@ def backup(request):
 @staff_member_required
 def battery(request):
     return render(request, 'serveradmin/battery.html',
-                  {'batteries': batinfo.batteries()})
```
mgautierfr: Also remove the `import batinfo`

bochecha: Good catch! Done.

PR #642, opened 2016-12-01
Better handle the library to stock migration

### LINKED ISSUE #626: bookspecimen to specimen migration edge case
The migration failed because we tried to migrate a bookspecimen with a serial already used by a specimen.

I think we need to consolidate such cases.

bochecha:
One of the things happening during this migration is that the serial
attribute of the old BookSpecimen model is becoming the barcode
attribute of the Specimen model.

Before this migration, people had to enter their books and specimens
both in the stock (StockItem/Specimen) and in the library
(Book/BookSpecimen).

Because they had to do it in both places, it is very possible that they
had a Specimen with a certain barcode, and a BookSpecimen with the same
serial. (see the first paragraph for how this field maps)

This crashes the migration, because Specimen.barcode has a unicity
constraint.

This commit ensures that the migration still works, removing the
existing StockItem and Specimen, effectively merging them with the new
Book and BookSpecimen.

Fixes #626

---

Note that this is based on @mgautierfr's #619, so let's merge that one first.

### DIFF
--- a/ideascube/library/migrations/0008_library_stock.py
+++ b/ideascube/library/migrations/0008_library_stock.py
@@ -11,26 +11,62 @@ def migrate_data(apps, schema_editor):
     OldBook = apps.get_model('library', 'OldBook')
     Book = apps.get_model('library', 'Book')
     BookSpecimen = apps.get_model('library', 'BookSpecimen')
+    Specimen = apps.get_model('monitoring', 'Specimen')
 
     db_alias = schema_editor.connection.alias
+
+    specimens = Specimen.objects.filter(barcode__isnull=False)
+    specimens = {s.barcode: s for s in specimens}
 
     for old_book in OldBook.objects.using(db_alias).order_by('created_at'):
         new_book = Book(
             module='library', name=old_book.title,
-            description=old_book.summary,
-            # And now the stuff which hasn't changed
+            description=old_book.summary, created_at=old_book.created_at,
             isbn=old_book.isbn, authors=old_book.authors, serie=old_book.serie,
             subtitle=old_book.subtitle, publisher=old_book.publisher,
             section=old_book.section, lang=old_book.lang, cover=old_book.cover,
             tags=old_book.tags)
+
+        old_bookspecimens = list(old_book.specimens.all())
+
+        # First try and find whether we need to relate this new Book to an
+        # existing StockItem
+        for old_bookspecimen in old_bookspecimens:
+            try:
+                old_specimen = specimens[old_bookspecimen.serial]
+
+            except KeyError:
+                continue
+
+            new_book.stockitem_ptr = old_specimen.item
+            new_book.id = old_specimen.item.id
+            break
+
         new_book.save()
 
-        for old_bookspecimen in old_book.specimens.all():
+        # Now migrate the BookSpecimen objects
+        for old_bookspecimen in old_bookspecimens:
+            try:
+                old_specimen = specimens[old_bookspecimen.serial]
+
+            except KeyError:
+                # There was no corresponding specimen in the stock
+                count = 1
+                specimen_ptr = None
+
+            else:
+                # There was a corresponding specimen in the stock
+                count = old_specimen.count
+                specimen_ptr = old_specimen
+
             new_bookspecimen = BookSpecimen(
-                barcode=old_bookspecimen.serial, item=new_book, count=1,
+                item=new_book,
+                count=count,
+                specimen_ptr=specimen_ptr,
+                barcode=old_bookspecimen.serial,
                 comments=old_bookspecimen.remarks,
-                # And now the stuff which hasn't changed
-                location=old_bookspecimen.location, file=old_bookspecimen.file)
+                location=old_bookspecimen.location,
+                file=old_bookspecimen.file)
             new_bookspecimen.save()
 
 
--- a/ideascube/library/migrations/0008_library_stock.py
+++ b/ideascube/library/migrations/0008_library_stock.py
@@ -163,7 +163,9 @@ class Migration(migrations.Migration):
             },
             bases=('monitoring.specimen', models.Model),
         ),
-        migrations.RunPython(migrate_data, hints={'using': 'default'}),
+        migrations.RunPython(
+            migrate_data, migrations.RunPython.noop,
+            hints={'using': 'default'}),
         migrations.DeleteModel('OldBookSpecimen'),
         migrations.DeleteModel('OldBook'),
     ]
--- a/ideascube/library/migrations/0008_library_stock.py
+++ b/ideascube/library/migrations/0008_library_stock.py
@@ -25,13 +25,13 @@ def migrate_data(apps, schema_editor):
             tags=old_book.tags)
         new_book.save()
 
-        for old_specimen in old_book.specimens.all():
-            new_specimen = BookSpecimen(
-                barcode=old_specimen.serial, item=new_book, count=1,
-                comments=old_specimen.remarks,
+        for old_bookspecimen in old_book.specimens.all():
+            new_bookspecimen = BookSpecimen(
+                barcode=old_bookspecimen.serial, item=new_book, count=1,
+                comments=old_bookspecimen.remarks,
                 # And now the stuff which hasn't changed
-                location=old_specimen.location, file=old_specimen.file)
-            new_specimen.save()
+                location=old_bookspecimen.location, file=old_bookspecimen.file)
+            new_bookspecimen.save()
 
 
 class Migration(migrations.Migration):
--- a/ideascube/library/tests/test_migrations.py
+++ b/ideascube/library/tests/test_migrations.py
@@ -18,6 +18,231 @@ section_int_to_name = {
 }
 
 
+@migration_test(
+    migrate_from=[
+        ('library', '0007_auto_20161007_1220'),
+        ('monitoring', '0004_auto_20161007_1240'),
+        ('taggit', '0002_auto_20150616_2121'),
+    ], migrate_to=[('library', '0008_library_stock')])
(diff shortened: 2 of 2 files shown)


ideascube/library/migrations/0008_library_stock.py
```diff
@@ -26,6 +30,13 @@ def migrate_data(apps, schema_editor):
         new_book.save()
 
         for old_bookspecimen in old_book.specimens.all():
+            if old_bookspecimen.serial in specimens:
```
mgautierfr: Why not using a request in the db here ?

ideascube/library/migrations/0008_library_stock.py
```diff
@@ -26,6 +30,13 @@ def migrate_data(apps, schema_editor):
...
+                # If "old serial" equals "new barcode" then we are dealing with
+                # the same specimen
+                old_specimen = specimens.pop(old_bookspecimen.serial)
+                old_specimen.item.delete()
+                old_specimen.delete()
```
mgautierfr: Should we delete the old_specimen or update it ?
I see at least InventorySpecimen that has a ForeignKey on the Specimen. What happen if we delete a specimen referenced by a InventorySpecimen ?

ideascube/library/migrations/0008_library_stock.py
bochecha: > Why not using a request in the db here ?

Because doing a request in the DB for every iteration would be way too costly.

mgautierfr: Hum, not sure :)
Getting almost all the specimen, generating a dict and do a __contains__ on this dict may be longer than a request per old_bookspecimen.

It depends mainly of the number of specimens, old_books and old_books.specimens.
It's also depends on a lots of code optimization I don't know about :)

And premature optimization is the root of all evil.;)
Especially if this migration script will be run only once.

However, code is written then no need to rewrite it.

bochecha:
Ok, so I've just pushed something which **I think** fixes [the problem @mgautierfr found](https://github.com/ideascube/ideascube/pull/642#discussion_r90891001)

However, please don't merge this yet.

The code is very complex, it took me a lot of trial and error to get here.

I want to add a bunch of tests before we merge this, because I'm not confident in it **at all**.

bochecha:
I added some tests, the migration seems to hold up.

It's quite hairy, so don't hesitate to ask for comments in the complex parts.

If you have ideas for special data cases I should have tested, do say, because I don't think it's possible to ever get too confident on this one.

ideascube/library/migrations/0008_library_stock.py
```diff
@@ -23,16 +27,45 @@ def migrate_data(apps, schema_editor):
...
+                # the same specimen
+                old_specimen = specimens.pop(old_bookspecimen.serial)
+
+            if old_specimen is not None:
+                # Either we just got old_specimen in the previous if condition,
+                # or we got it on a previous iteration
+
+                if new_book.id is None:
+                    new_book.stockitem_ptr = old_specimen.item
+                    new_book.id = old_specimen.item.id
```
mgautierfr: Hum, I understand why you do this two lines. You make the Book inherit from the StockItem you've just found from the old_specimen.

But, when you create the new Book (l.22), django should have create a (temporary) StockItem has the Book inherit from it. So the given `name` and `description` are stored in a this (temporary) StockItem. If you change the stockitem here, the `name` and `description` will(/may?) be lost.

ideascube/library/migrations/0008_library_stock.py
```diff
@@ -23,16 +27,45 @@ def migrate_data(apps, schema_editor):
...
+
+                new_bookspecimen = BookSpecimen(
+                    specimen_ptr=old_specimen,
+                    # And now the stuff which hasn't changed
+                    location=old_bookspecimen.location,
+                    file=old_bookspecimen.file)
```
mgautierfr: Do we've lost the (BookSpeciment) comments / (oldBookSpeciment) remarks ?
The remarks become the comments in the migration. You take this into account when you create the BookSpecimen line 55, but not here.

ideascube/library/migrations/0008_library_stock.py
bochecha: Nope. They aren't lost.

Django doesn't create anything until we actually run `save()`.

That's why I run `save()` for books at the end of the iteration.

ideascube/library/tests/test_migrations.py
```diff
@@ -20,6 +20,126 @@
...
+    assert bookspecimen.serial is None
+    assert bookspecimen.item.book.id == book.id
+
+    specimen = specimens[2]
+    assert specimen.id == old_specimen_id
+    assert specimen.bookspecimen.id == bookspecimen.id
+    assert specimen.item.id == item.id
+
+    bookspecimen = bookspecimens[3]
+    assert bookspecimen.barcode == '4567'
```
mgautierfr: `assert bookspeciment.id != old_specimen_id`
(bookspeciment 4567 has no already existing specimen)

ideascube/library/tests/test_migrations.py
```diff
@@ -20,6 +20,126 @@
 }
 
 
+@migration_test(
+    migrate_from=[
+        ('library', '0007_auto_20161007_1220'),
+        ('monitoring', '0004_auto_20161007_1240'),
+        ('taggit', '0002_auto_20150616_2121'),
+    ], migrate_to=[('library', '0008_library_stock')])
+def test_library_to_monitoring_with_duplicated_data(migration):
```
mgautierfr: A global comment for this test.

- Please use more explicit variable name than `book`, `item`, `specimen` when you're testing that migration had run correctly.
- Please test the books in the same order you've create them (this migration is far complex enough without that)

ideascube/library/migrations/0008_library_stock.py
```diff
@@ -23,16 +27,45 @@ def migrate_data(apps, schema_editor):
...
-                location=old_bookspecimen.location, file=old_bookspecimen.file)
-            new_bookspecimen.save()
+            if old_bookspecimen.serial in specimens:
+                # If "old serial" equals "new barcode" then we are dealing with
+                # the same specimen
+                old_specimen = specimens.pop(old_bookspecimen.serial)
+
+            if old_specimen is not None:
+                # Either we just got old_specimen in the previous if condition,
+                # or we got it on a previous iteration
```
mgautierfr: > That means we can only associate a BookSpecimen with the corresponding Specimen, but we have nothing to associate a Book to its corresponding StockItem.

Yes, but this is already done at the first iteration, no need to keep it to the second iteration.
Moreover, by keeping the `old_specimen` you may associate a `new_bookspecimen` without corresponding `specimen` to the `old_specimen`. And so having two `bookspecimen` inherit from the same `specimen`

ideascube/library/migrations/0008_library_stock.py
mgautierfr: That's ok then :)

ideascube/library/migrations/0008_library_stock.py
bochecha: > Yes, but this is already done at the first iteration

Not necessarily, no.

If the book has two bookspecimens, and only the second how has a corresponding specimens.

> Moreover, by keeping the old_specimen you may associate a new_bookspecimen without corresponding specimen to the old_specimen.

If I did that it would indeed be a pretty bad bug. Am I doing it?

mgautierfr: Yes, but this is already done at the first iteration **where there is a bookspecimen with a corresponding specimen**

So no need to keep the old_specimen for another iteration (with corresponding specimen or not)

ideascube/library/migrations/0008_library_stock.py
bochecha: I think we were, yes. Fixed.

ideascube/library/tests/test_migrations.py
bochecha: > Please use more explicit variable name than book, item, specimen when you're testing that migration had run correctly.

Like what?

Since those variables are reused, I don't really see what could be more explicit than the type being tested.

> Please test the books in the same order you've create them (this migration is far complex enough without that)

I'm testing them in the order I did, because I need to `order_by` them to fix the order.

So I've tested them in the order returned by `order_by`.

Do you suggest I test them like so?

```python
# First, the book without specimens
book = books[1]
...

# Next, the book with two specimens
book = books[0]
...

# And now, the book with specimens and relation to the stock
book = books[2]
...
```

I don't mind, just trying to understand if that's what you're asking for.

ideascube/library/tests/test_migrations.py
bochecha: Good idea, especially with regards to https://github.com/ideascube/ideascube/pull/642#discussion_r91484847

bochecha:
Now with the fixes for the issues you found, and more tests for cases we talked about.

bochecha:
FYI, I just force-pushed a change in this migration: explicitly setting the `Book.created_at` attribute. Without this, `new_book.save()` was failing with: `django.db.utils.IntegrityError: NOT NULL constraint failed: library_book.created_at`.

I'm not sure why that happened, since `Book` inherits from `TimeStampedModel` which sets `created_at` as `auto_now_add=True`.

My guess is that this is (once again) due to the fact that in migrations we don't have access to the actual model, but to those weird "proxy" objects, and that `auto_now_add=True` somehow doesn't work for these.

ideascube/library/tests/test_migrations.py
mgautierfr: This is tested line 179.
So, this is ok for me.

mgautierfr:
This looks good now.
The migration is passing correctly on several databases (especially the one revealing the buggy migration).

It's time to merge this PR !

PR #643, opened 2016-12-02
Csv2pkg update

mgautierfr:


ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -13,8 +16,29 @@
 Optional columns are "lang", "preview", "kind", "tags".
 Attributes "path" and "preview" (if provided) must be path
 (relative to csv directory) to existing files.
+
+csv2pkg also generate a yaml to add to the catalog.
```
letompouce: generateS

ideascube/serveradmin/package.py
```diff
@@ -8,26 +8,37 @@
...
             if not media.get(required_metadata):
+                if self.report:
+                    self.report.warning("Metadata missing, ignoring the corresponding media",
+                                        "{} at line {}".format(required_metadata, line))
                 return False
-        if not op.exists(op.join(self.working_dir, media['path'])):
+        full_path = op.join(self.working_dir, media['path'])
+        if not op.exists(full_path):
+            if self.report:
+                self.report.warning("Path does not exists, ignoring the corresponding media",
```
letompouce: s/exists/exist/

ideascube/serveradmin/package.py
```diff
@@ -8,26 +8,37 @@
...
+                                    "{} used at line {}".format(full_path, line))
             return False
         preview_path = media.get('preview')
         if (preview_path
-         and not op.exists(op.join(self.working_dir, preview_path))):
-            print("Warning : preview_path {} defined but file does not exists"
```
letompouce: s/exists/exist/

ideascube/serveradmin/package.py
```diff
@@ -8,26 +8,37 @@
...
-                  .format(preview_path))
+           and not op.exists(op.join(self.working_dir, preview_path))):
+            if self.report:
+                self.report.warning("Preview path does not exists.",
```
letompouce: (guess what)

ideascube/serveradmin/management/commands/csv2pkg.py
letompouce: generateS

barbayellow: kind is NOT optionnal

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -25,6 +49,18 @@ def add_arguments(self, parser):
...
+        group.add_argument('--package-id', default="<package_id>",
+                           help="Identifier of the package.")
+        group.add_argument('--name', default="<Package name>",
+                           help="Name of the package.")
+        group.add_argument('--description', default="<Package description>",
+                           help="Description of the package.")
+        group.add_argument('--language', default="<lang>",
+                           help="Language of the package.")
+        group.add_argument('--url', default="<http://...>",
+                           help="URL where the package we will be available")
```
bochecha: I wouldn't bother with the URL...

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -59,3 +95,15 @@ def handle(self, *args, **options):
...
+
+        if not options['dry_run']:
+            print("\nYaml to use is: \n")
+            yaml_metadata = generate_catalog_yaml_metadata(
+                package_path,
+                'zipped-medias',
+                options['package_id'],
+                options['name'],
+                options['description'],
+                options['language']
```
bochecha: ... but if you add the URL as a cli argument, then you probably should put it in the YAML. 😉 

(but again, I just wouldn't bother)

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -13,8 +16,29 @@
...
 Attributes "path" and "preview" (if provided) must be path
 (relative to csv directory) to existing files.
+
+csv2pkg also generates a yaml to add to the catalog.
+Some attributes are automically calculated from the generated package.
+Other attributes may be provided as command arguments.
 """
 
+def generate_catalog_yaml_metadata(path, _type, package_id, name,
+                                   description, language):
```
bochecha: A function for the only purpose of creating a dict seems like a bit much.

```python
print(yaml.safe_dump(
    {
        options['package_id']: {
            'name': options['name'],
            'description': options['description'],
            ...
        }
    }, default_flow_style=False)
```

?

ideascube/serveradmin/package.py
```diff
@@ -36,7 +47,7 @@ def from_csv(cls, csv_path):
             except csv.Error:
                 dialect = csv.unix_dialect()
             medias = list(csv.DictReader(content.splitlines(), dialect=dialect))
-            return cls(working_dir, medias)
+            return cls(working_dir, medias, report)
```
bochecha: `medias` and `report` are keyword arguments, please pass them as such.

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -21,18 +48,54 @@ def add_arguments(self, parser):
...
+        group.add_argument('--package-id', default="<package_id>",
+                           help="Identifier of the package.")
+        group.add_argument('--name', default="<Package name>",
+                           help="Name of the package.")
+        group.add_argument('--description', default="<Package description>",
+                           help="Description of the package.")
+        group.add_argument('--language', default="<lang>",
+                           help="Language of the package.")
+        group.add_argument('--url', default="<http://...>",
+                           help="URL where the package we will be available")
```
bochecha: I think @letompouce meant you had mixed `we` and `will`.

That is, you should have replaced `we` by `will`, not added `will`. 😄

ideascube/utils.py
```diff
@@ -66,3 +68,22 @@ def to_unicode(text):
...
+    with open(path, 'rb') as f:
+        while True:
+            data = f.read(8388608)
+
+            if not data:
+                break
+
+            sha.update(data)
+
+    return sha.hexdigest()
```
bochecha: I see this is done in the latest push to this branch.

ideascube/serveradmin/management/commands/csv2pkg.py
mgautierfr: The function doesn't just create a dictionary.
It also make some operation on the file to get its sha and  size.

And I like to separate code in different function when the code is about different stuff.
The problem here come from the function definition who take to much arguments and so seems to just put all the given arguments in a dict. (And seems to be a code duplication with the caller code that extract information from the options dict)

I will rewrite the function to take only the `path` and the `options` arguments.

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -21,18 +22,37 @@ def add_arguments(self, parser):
...
 
-        package = MediaCenterPackage.from_csv(csv_path)
-        if not options['dry_run']:
+        package = MediaCenterPackage.from_csv(csv_path, report=self.report)
+        print(self.report)
+
+        if not package.medias:
+            print("There is no (valid) media to create the package.\n"
+                  "No package will be created")
+            self.abort()
```
mgautierfr: I would be of bad faith, I would say to print in stdout instead of stderr.

But the truth is that I forget the `self.abort` and add it later. I will pass the message to `abort`

ideascube/serveradmin/management/commands/csv2pkg.py
mgautierfr: Good catch, I will add it

ideascube/serveradmin/package.py
```diff
@@ -8,26 +8,37 @@
 
 
 class MediaCenterPackage:
-    def __init__(self, working_dir, medias=None):
+    def __init__(self, working_dir, medias=None, report=None):
         self.working_dir = working_dir
-        self.medias = list(filter(self.check_media, medias or []))
+        self.report = report
+        filtered = filter(self.check_media, enumerate(medias or [], 1))
+        self.medias = [elem[1] for elem in filtered]
```
mgautierfr: Yes. There where already the filter construction and I stick with it. A for loop is far better.

ideascube/serveradmin/package.py
mgautierfr: The same old discussion here. For me they are not a keyword arguments. They are arguments with default value.

ideascube/utils.py
```diff
@@ -66,3 +68,22 @@ def to_unicode(text):
...
+
+def get_file_size(path):
+    with open(path, 'rb') as f:
+        size = f.seek(0, os.SEEK_END)
```
mgautierfr: > Why not simply use os.stat(path).st_size?

Old habits die hard.

ideascube/serveradmin/management/commands/csv2pkg.py
mgautierfr: huhu. Monday morning... :)

letompouce:
I spotted a minor bug/enhancement into the YAML generation.

A catalog starts by `all:` and all other keys are then indented by two more spaces ([example](http://catalog.ideascube.org/kiwix.yml.html)).

In the current state, `csv2pkg` displays a YAML which first key is not indented.

Please make the whole generated YAML indented by two more spaces, so we can just copy/paste it into our catalog file.

mgautierfr:
I add the `all:` entry at top of the generated YAML.

This make a full YAML that can be used out of the box in a catalog.yaml file (assuming the this catalog will have only one package and that you provide all the metadata information to the csv2pkg command)

But you're free to copy only the package metadata :)

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -13,8 +16,30 @@
...
+csv2pkg also generates a yaml to add to the catalog.
+Some attributes are automically calculated from the generated package.
+Other attributes may be provided as command arguments.
 """
 
+
+def generate_catalog_yaml_metadata(path, metadata, _type='zipped-medias'):
+    return {
+        'all' : {
+            metadata['package_id']: {
```
letompouce: These fields are displayed th their alphabetical order. Is it possible to fix an arbitrary order?
name, description, language, version, url, size, sha256sume, type.

bochecha: Just a pedantic comment on this:

> These fields are displayed in their alphabetical order.

There is no ordering guarantee for Python dictionaries.

Therefore, you saw the fields displayed in their alphabetical order, but you can't assume they always will be.

Even if you ran it 100 times and it always gave you the alphabetical ordering, it's only what happened in your experience, but there is no guaranteed that it will always happen.

Consider the current ordering to be random, and potentially changing every time.

---

As for the possibility to fix a given order, it should be doable, yes: http://stackoverflow.com/a/31609484

mgautierfr: I've added a fixed order for the metadata.
It should be ok now.

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -23,19 +24,29 @@
 """
 
 
+def ordered_representer(dumper, data):
+    return dumper.represent_mapping('tag:yaml.org,2002:map', data.items())
+
+yaml.add_representer(OrderedDict, ordered_representer, Dumper=yaml.SafeDumper)
```
bochecha: This might end up being useful elsewhere, so eventually we might want to move it to some kind of `utils` module.

Not asking you to do it now, just writing it down because it helps me remember. 😄

ideascube/serveradmin/management/commands/csv2pkg.py
```diff
@@ -23,19 +24,29 @@
...
+        if k == 'version':
+            package_metadata[k] = date.today().isoformat()
+        elif k == 'sha256sum':
+            package_metadata[k] = get_file_sha256(path)
+        elif k == 'size':
+            package_metadata[k] = get_file_size(path)
+        elif k == 'type':
+            package_metadata[k] = _type
+        else:
+            package_metadata[k] = metadata[k]
```
bochecha: That's, huh, a bit baroque.

What's wrong with:

```python
    package_metadata = OrderedDict([
        ('name', metadata['name']),
        ('description', metadata['description']),
        ('language', metadata['language']),
        ('version', date.today().isoformat()),
        ('url', metadata['url']),
        ('size', get_file_size(path)),
        ('sha256sum', get_file_sha256(path)),
        ('type', _type),
    ])
```

It's 3 less lines, no `for` loop full of `if`s.

If it were a complicated structure and we were receiving the list of keys from elsewhere then sure, a more systematic approach like you have would be nicer.

But this is such a simple thing, that your way just looks like needless complexity.

ideascube/serveradmin/tests/test_csv2pkg_command.py
```diff
@@ -41,6 +41,11 @@ def test_csv2pkg_command(package_path, csv_writer, capsys):
                            }
 
     assert "Package {} has been created with 3 medias.".format(package_path) in out
+    #Check the order of the output
+    for k in ('name', 'description', 'language', 'version', 'url', 'size', 'sha256sum', 'type'):
+        index = out.find('    '+k)
+        assert index != -1, '{} is not in the right order'.format(k)
+        out = out[index:]
```
bochecha: You could use `index = out.find('    '+k, index)` and not need to modify `out`.

ideascube/serveradmin/management/commands/csv2pkg.py
mgautierfr: Huhu, I haven't think about this construction. I was thinking of `OrderedDict(name=..., description=...)` which, of course, doesn't keep the order.

I've written the `for` loop telling me it was pretty complex for nothing.

ideascube/serveradmin/tests/test_csv2pkg_command.py
mgautierfr: As `out` is not used elsewhere it's more a matter of style.
But I don't especially care, I will change it :)

PR #646, opened 2016-12-05
users: Add an 'extra' field

bochecha:
This is a free-form text field, which can be used for anything at all, depending on the needs of each deployment.

However, it should only be used for things that are too specific to put in a dedicated field shared by all deployments.


bochecha:
> Also, there is a migration script conflict, probably because of a merge with another branch.

Good catch, I missed that when I rebased after merging #617

PR #648, opened 2016-12-07
Conf bibliotecamovil

letompouce:


ideascube/conf/idb_col_llavedelsaber.py
```diff
@@ -4,3 +4,9 @@
 LANGUAGE_CODE = 'es'
 DOMAIN = 'bibliotecamovil.lan'
 ALLOWED_HOSTS = ['.bibliotecamovil.lan', 'localhost']
+
+USER_FORM_FIELDS = USER_FORM_FIELDS + (
+        (_('Personal informations'), ['extra'], ['disabilities']),
```
bochecha: The two fields should be in a single list.

PR #651, opened 2016-12-08
Document our release process

bochecha:
This is based on notes from @yohanboniface.

It's certainly not perfect, but it's a pretty decent beginning.

### DIFF
--- a/docs/release_process.md
+++ b/docs/release_process.md
@@ -1,0 +1,79 @@
+# Release Process
+
+So you want to release a new version, great!
+
+First, make sure all the desired pull requests have been merged.
+
+Next, pull [the latest translations](translate.md).
+
+Finally, here are the steps to follow.
+
+## Get a clean environment
+
+Create a new virtual environment (`pew` makes this trivial), then get the code:
+
+    $ pew mktmpenv -p python3
+    $ git clone ssh://git@github.com/ideascube/ideascube
+    $ cd ideascube
+    $ git checkout master
+
+Now install all the dependencies:
+
+    $ make develop
+
+## Sanity check
+
+Run the tests:
+
+    $ py.test
+
+And run the migrations:
+
+    $ python manage.py migrate --database=default
+    $ python manage.py migrate --database=transient
+
+Let's make sure there are no migration conflicts:
+
+    $ rm -fr storage
+    $ git checkout $TAG_FOR_LATEST_RELEASE
+    $ python manage.py migrate --database=default
+    $ python manage.py migrate --database=transient
+    $ make dummydata
+    $ git checkout master
+    $ python manage.py migrate --database=default
+    $ python manage.py migrate --database=transient
+
+Ideally, in the above steps, it would be even better to use a real-world
+database (from a server running the latest release) instead of the dummy data.
+
+## Release
+
+First, decide what the new version will be. Is it a new major, minor or patch
+release?
+
+At the moment, since we are still in the pre-1.0 days, we usually follow this
+convention:
+
+* if the release introduces new features, bump the **minor** version;
+* if the release only fixes bugs, adds config files, then bump the **patch**
+    version.
+
+Write the `debian/changelog` entry for the new version. Review all the commits
+in master since the last release to make sure you don't forget anything.
+
+Bump the version in `ideascube/__init__.py`.
+
+Commit, using the new version as the commit message:
+
+    $ git commit -a -m "x.x.x"
+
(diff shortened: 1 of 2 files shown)


docs/release_process.md
```diff
@@ -0,0 +1,82 @@
...
+
+First, make sure all the desired pull requests have been merged.
+
+## Update the translations
+
+**TODO**
+
+## Get a clean environment
+
+Create a new virtual environment (`pew` makes this trivial), then get the code:
```
letompouce: How does one install pew?

letompouce: Fine with me, thanks!

docs/release_process.md
```diff
@@ -0,0 +1,82 @@
...
+
+    $ pew mktempenv -p python3
```
letompouce: s/mktempenv/mktmpenv/

docs/release_process.md
```diff
@@ -0,0 +1,82 @@
...
+    $ git clone https://github.com/ideascube.ideascube
```
letompouce: s@/@.@

bochecha: I think you meant `s@.@/@`. :P 

Good catch though, fixed.

docs/release_process.md
bochecha: Fixed.

docs/release_process.md
```diff
@@ -0,0 +1,82 @@
...
+## Update the translations
+
+**TODO**
+
+## Get a clean environment
+
+Create a new virtual environment (`pew` makes this trivial), then get the code:
+
+    $ pew mktmpenv -p python3
+    $ git clone https://github.com/ideascube/ideascube
```
letompouce: Better use a git:// protocol, or you'll have a hard time trying to push over HTTP.

mgautierfr: Git push over http is available. It's even easier as git:// is sometime blocked by proxy/firewall depending of your connection.

bochecha:
Just added a bit about translation management.

docs/release_process.md
```diff
@@ -0,0 +1,80 @@
...
+
+And run the migrations:
+
+    $ python manage.py migrate --database=default
+    $ python manage.py migrate --database=transient
+
+Let's make sure there are no migration conflicts:
+
+    $ rm -fr storage
+    $ git checkout $TAG
```
mgautierfr: `$TAG_OF_LAST_RELEASE` instead of `$TAG` ?

docs/release_process.md
```diff
@@ -0,0 +1,80 @@
...
+    $ git commit -a -m “x.x.x”
+
+Create a new tag:
+
+    $ git tag x.x.x
+
+Push the release:
+
+    $ git push origin master
+    $ git push --tags
```
mgautierfr: `$ git push origin master x.x.x`

This way, we are sure that only the release tag is push, not all tags.
If user uses a clean env, there should be no extra tags to push, but lets be sure.

docs/release_process.md
```diff
@@ -0,0 +1,80 @@
+# Release Process
+
+So you want to release a new version, great!
+
+First, make sure all the desired pull requests have been merged.
+
+Next, pull [the latest translations](translate.md).
```
mgautierfr: We don't have to just pull the latest translations. We have to push them to transiflex, be sure they are translated and pull them.

All those steps are explain in the corresponding section, but let be more clear.

bochecha: > We don't have to just pull the latest translations. We have to push them to transiflex, be sure they are translated and pull them.

I don't think the release is the right moment to push the translations and wait for translators to do their magic.

We should do that much earlier.

When comes the time to make a release, we can only pull what has been translated up to now.

docs/release_process.md
bochecha: I agree with you.

However, our package building process would be very unhappy if we pushed a tag which isn't for a release.

That's certainly one more reason to use the syntax you suggested, but I felt I needed to say it, because we really shouldn't push tags for anything but releases, unless we start thinking hard about our package build process.

docs/release_process.md
bochecha: Thanks, fixed.

docs/release_process.md
bochecha: And fixed.

docs/release_process.md
mgautierfr: Yes, you're right.

PR #656, opened 2016-12-15
Properly import users from a CSV export

### LINKED ISSUE #655: User import /export consistency
In the users app, it's possible to import users from a csv file or to export users.
an export gives us a file with this column : 
date de création, |identifiant|extra|Disabilities (depending on the configuration of ideascube).
BUT when importing a file with the same columns and the same format, we get an error : "Invalid row at line 1: serial: * Ce champ est obligatoire". I've gtried with a column named "serial" and it works.
Shouldn't we have the same format in import and exit data ?

bochecha:
We export users using the translated verbose name, so that the CSV is
easier to handle by actual humans in their favourite spreadsheet
software.

That means we need to get back to the canonical field name when
importing.

Fixes #655

### DIFF
--- a/ideascube/forms.py
+++ b/ideascube/forms.py
@@ -69,11 +69,24 @@ class CreateStaffForm(forms.ModelForm):
 class UserImportForm(forms.Form):
     source = forms.FileField(required=True)
 
+    def canonicalize_row(self, row):
+        result = {}
+        data_fields = User.get_data_fields()
+        data_fields = {str(f.verbose_name): f.name for f in data_fields}
+
+        for field in row:
+            # TODO: What to do on KeyError?
+            canonical_name = data_fields[field]
+            result[canonical_name] = row[field]
+
+        return result
+
     def save(self):
         source = TextIOWrapper(self.cleaned_data['source'].file)
         users = []
         errors = []
         for idx, row in enumerate(csv.DictReader(source)):
+            row = self.canonicalize_row(row)
             qs = User.objects.all()
             try:
                 instance = qs.get(serial=row['serial'])


ideascube/forms.py
```diff
@@ -69,11 +69,24 @@ def save(self, *args, **kwargs):
 class UserImportForm(forms.Form):
     source = forms.FileField(required=True)
 
+    def canonicalize_row(self, row):
+        result = {}
+        data_fields = User.get_data_fields()
+        data_fields = {str(f.verbose_name): f.name for f in data_fields}
```
bochecha: I started with the assumption that if we used the `verbose_name` for export, then there must be a good reason, and as such I made the import capable of getting back to the canonical field name.

But let's challenge this assumption, maybe it is the export which should always use the canonical field name?

ideascube/forms.py
```diff
@@ -69,11 +69,24 @@ def save(self, *args, **kwargs):
...
+
+        for field in row:
+            # TODO: What to do on KeyError?
```
mgautierfr: The KeyError could also be raised in case of a translation change.
If we update the field translation between the export and the import, names will not match with the field.

I'm the feeling that what is wrong is the header translation at the export itself.

@barbayellow what is the purpose of the export ?
Are we exporting to backup/exchange with another software or are we exporting for a human reading/editing ?

ideascube/forms.py
barbayellow: As far as I know, the export has been used to backup data or to load backuped data within ideascube. So far, BSF never received an export from the users from an ideasbox.
However, the aimed goal is to have insight of who is using the box (age, sex, educationnal level, etc.).

mgautierfr:
This PR is superseded by #657.

PR #657, opened 2016-12-15
users: Use the field names when exporting

### LINKED ISSUE #655: User import /export consistency

bochecha:
Using the translated verbose names leads to nicer exports which humans
can better understand and process with their favourite spreadsheet
software.

However, it also means that an Ideascube export can't be imported back
into Ideascube without first renaming the column headers to the field
names.

@barbayellow says user exports have never been used by actual humans,
but only as backups.

Therefore, it seems to make sense to use the field names when exporting.

Fixes #655
Supercedes #656

### DIFF
--- a/ideascube/tests/test_views.py
+++ b/ideascube/tests/test_views.py
@@ -410,8 +410,8 @@ def test_export_users_should_return_csv_with_users(staffapp, settings):
     user2 = UserFactory(short_name="user2", full_name=u"I'm user2 with é")
     resp = staffapp.get(reverse('user_export'), status=200)
     resp.mustcontain(
-        'identifier', user1.serial, user2.serial, no=[
-            'usual name', user1.short_name, user2.short_name,
+        'serial', user1.serial, user2.serial, no=[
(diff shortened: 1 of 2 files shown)


bochecha:
This is an alternative take on fixing #655.

It is pretty much the opposite of #656, as per my conversation with @barbayellow this morning.

The decision to use the `verbose_name`s when exporting users seems to have been a conscious design decision, given there are unit tests to verify it.

However, that's from before I joined the project, so I can't say why it was done this way, or whether things changed and we could come back on that decision.

In any case, we should either pursue with this or #656, but not both.

Issue #659, opened 2016-12-16
Add the serial in the main view of the stock app


barbayellow:
For logistics reasons our partner in Colombia has asked us to add the serial (when any) in the main view of the stock app. It appears to e a good idea.

![serial](https://cloud.githubusercontent.com/assets/4816690/21273143/417a242a-c390-11e6-8ba4-527db84df355.png)

PR #672, opened 2017-01-05
User can now scan the serial to make inventory or handle loans and returns.

### LINKED ISSUE #660: Inventory should be made with the barcode OR the serial
Sometimes it is very useful to scan only the serial number rather than a barcode because some items don't have barcode but they have a serial number. That will avoid to stick a barcode sticker when there is already a serial. Ex : computers, tablets, ereaders.
Serial and barcode are both unique numbers so that makes sense.

mgautierfr:
Fixes #660


ideascube/monitoring/tests/test_views.py
```diff
@@ -424,7 +521,23 @@ def test_cannot_loan_twice_same_item(staffapp, user):
     assert Loan.objects.due().count() == 1
 
 
-def test_can_loan_twice_same_item_when_first_is_returned(staffapp, user):
+def test_cannot_loan_twice_same_item_from_serial(staffapp, user):
```
bochecha: How about an item loaned by barcode, can we then loan it a second time by serial? 😜

ideascube/monitoring/models.py
```diff
@@ -127,6 +127,10 @@ class Specimen(models.Model):
 
     objects = SpecimenQuerySet.as_manager()
 
+    @property
+    def user_identifier(self):
```
bochecha: `user_identifier` is a pretty poor name for this, since `user` generally refers to the user model, the logged in user, etc...

I generally see things like `display_*` or `pretty_*` being commonly used in such cases.

ideascube/monitoring/tests/test_views.py
```diff
@@ -460,6 +609,18 @@ def test_can_return_loan(staffapp, user):
     assert not Loan.objects.due().count()
 
 
+def test_can_return_loan_from_serial(staffapp, user):
```
bochecha: Can we return a loan from serial if the loan had been made with the barcode? And the opposite?

ideascube/monitoring/tests/test_views.py
```diff
@@ -325,6 +358,48 @@ def test_cant_create_inventoryspecimen_by_barcode_twice(staffapp):
                                             specimen=specimen).count() == 1
 
 
+def test_staff_can_create_inventoryspecimen_by_serial(staffapp):
```
bochecha: What about trying to create an inventory barcode by serial if it previously was created by serial? 😜

ideascube/monitoring/forms.py
```diff
@@ -143,14 +155,13 @@ class ReturnForm(forms.Form):
...
-        try:
-            loan = Loan.objects.due().get(specimen__barcode=barcode)
-        except Loan.DoesNotExist:
-            msg = _('Item with barcode {barcode} is not loaned.')
+        loan = Loan.objects.due().filter(specimen__barcode=barcode).first()
+        if loan is None:
+            loan = Loan.objects.due().filter(specimen__serial=barcode).first()
+        if loan is None:
+            msg = _('Item with barcode or serial number {barcode} is not '
+                    'loaned.')
```
mgautierfr: Because this (previous) `try/except` except for `Loan.DoesNotExist` but also `Loan.MultipleObjectsReturned`.
Adding the same code (one try and two except) in the `Loan.DoesNotExist` execpt clause seems a bit to big for me.

bochecha: > Because this (previous) `try`/`except` except for `Loan.DoesNotExist` but also `Loan.MultipleObjectsReturned`.

Right, but given the comment I think we could just remove that case, provided all boxes are upgraded past that version. (see my next comment)

ideascube/monitoring/forms.py
```diff
@@ -143,14 +155,13 @@ class ReturnForm(forms.Form):
...
             raise forms.ValidationError(msg.format(barcode=barcode))
-        except Loan.MultipleObjectsReturned:
-            # Should araise only after migration 0.3.2 to 0.3.0.
-            loan = Loan.objects.due().filter(specimen__barcode=barcode).first()
```
mgautierfr: This code is not removed.
As `filter(...).first()` "includes" `get(...)`, I always use the more global version.

ideascube/monitoring/models.py
mgautierfr: I named it like that because this is the identifier from the user point of view.
But yes, maybe `pretty_indetifier` is better.

ideascube/monitoring/templates/monitoring/loan.html
```diff
@@ -28,7 +28,7 @@
         </tr>
         {% for loan in loans %}
             <tr>
-                <td>{{ loan.specimen.item }} - {{ loan.specimen.barcode }}</td>
+                <td>{{ loan.specimen.item }} - {{ loan.specimen.user_identifier }}</td>
```
mgautierfr: If user start to scan the serial and do not print a specific ideascube barcode there will be no barcode and so, nothing displayed.
I think it's better to display the serial if there is no barcode as this is what the user use.

ideascube/monitoring/tests/test_views.py
```diff
@@ -206,6 +206,16 @@ def test_staff_can_create_specimen(staffapp):
...
 
 
+def test_staff_can_create_specimen_with_serial(staffapp):
+    item = StockItemFactory()
+    url = reverse('monitoring:specimen_create', kwargs={'item_pk': item.pk})
+    form = staffapp.get(url).forms['model_form']
+    assert not item.specimens.count()
+    form['serial'] = '23135321'
+    form.submit().follow()
+    assert item.specimens.count()
```
mgautierfr: Right

ideascube/monitoring/tests/test_views.py
```diff
@@ -216,6 +226,16 @@ def test_staff_can_create_specimen_with_letters_and_ints_in_barcode(staffapp):
     assert item.specimens.filter(barcode='123abc').count()
 
 
+def test_staff_can_create_specimen_with_letters_and_ints_in_serial(staffapp):
```
mgautierfr: > Functionally, can a serial contain letters and ints?

I see no reason why it couldn't.

> If so, how is this related to this commit?

I've used the same tests that for the specimen creation with barcode.

ideascube/monitoring/templates/monitoring/loan.html
bochecha: In this case maybe we should do the same as #659 then?

ideascube/monitoring/tests/test_views.py
mgautierfr: Ho, yes. Missing this test.

In a more general way, it's missing some tests mixing barcode and serial.
However, I'm not sure of what we must do in this case.
The original need is to be able to make inventory using serial to not have to print a barcode.
So, do we must handle the case of a user mixing the use-cases ? @barbayellow  ?

ideascube/monitoring/tests/test_views.py
bochecha: > I see no reason why it couldn't.

Me neither, but that might just be that you and I aren't aware of how things are expected to work in the field.

Currently, without this test, then we just don't know what the serial string can contain (which probably means it can contain anything). Adding this test enshrines a given format, documents it as the expected norm.

Basically, since we don't know that this is the expected norm, then we shouldn't add this test, because it might be wrong.

However we should try to obtain details on how serials are actually formed, and validate **that** with a test. (along with the required validation on the form side of things)

But that would probably be a concern for a very different pull request.

ideascube/monitoring/forms.py
```diff
@@ -110,24 +118,28 @@ class LoanForm(forms.ModelForm):
...
+        return specimen
 
     def clean_user(self):
         serial = self.cleaned_data['user']
         try:
             user = user_model.objects.get(serial=serial)
         except user_model.DoesNotExist:
             raise forms.ValidationError(
-                _('Identifier {serial} not found'.format(serial=serial)))
+                _('Identifier {serial} not found').format(serial=serial))
```
bochecha: I see you've split it out, thanks, closing this thread.

ideascube/monitoring/tests/test_views.py
bochecha: > The original need is to be able to make inventory using serial to not have to print a barcode.

Not have to print it, yes.

But then it still might be printed at a later date, because somebody had a bit too much time.

I think these mixed tests would be very useful.

ideascube/monitoring/forms.py
```diff
@@ -110,24 +113,28 @@ class LoanForm(forms.ModelForm):
 
     def clean_specimen(self):
         barcode = self.cleaned_data['specimen']
-        if Loan.objects.due().filter(specimen__barcode=barcode).exists():
+        if (Loan.objects.due().filter(specimen__barcode=barcode).exists()
+         or Loan.objects.due().filter(specimen__serial=barcode).exists()):
```
bochecha: I know you disagree with pep8 on this specific matter of alignment of conditions.

For now I think the pep8 way is better than yours, but I'm curious: how would you have written the above two lines with a `and` instead of the `or`?

ideascube/monitoring/templates/monitoring/loan.html
```diff
@@ -28,7 +28,7 @@
         </tr>
         {% for loan in loans %}
             <tr>
-                <td>{{ loan.specimen.item }} - {{ loan.specimen.barcode }}</td>
+                <td>{{ loan.specimen.item }} - {{ loan.specimen.pretty_identifier }}</td>
```
bochecha: I really think it would be better to display both the `barcode` and the `serial` here, just like in #659.

ideascube/monitoring/templates/monitoring/specimen_confirm_delete.html
```diff
@@ -1,4 +1,4 @@
 {% extends 'ideascube/form_confirm_delete.html' %}
 {% load i18n %}
 
-{% block confirm_message %}{% blocktrans with specimen=specimen.barcode item=specimen.item %}Are you sure you want to delete the specimen "{{ specimen }}" of "{{ item }}"?{% endblocktrans %}{% endblock %}
+{% block confirm_message %}{% blocktrans with specimen=specimen.pretty_identifier item=specimen.item %}Are you sure you want to delete the specimen "{{ specimen }}" of "{{ item }}"?{% endblocktrans %}{% endblock %}
```
bochecha: I really think it would be better to display both the `barcode` and the `serial` here, just like in #659.

ideascube/monitoring/templates/monitoring/stock.html
```diff
@@ -15,7 +15,7 @@
         {% for item in module.objects  %}
             <tr class="stockitem" id="stockitem-{{ item.pk }}"><td>{{ item }} <em>{{ item.description|truncatewords:10 }}</em></td><td>{% trans 'Total:' %} {{ item.specimens.physical|length }}</td><td class="actions"><a href="{% url 'monitoring:stockitem_update' pk=item.pk %}" title="{% trans 'Edit' %}">{% fa 'pencil-square' %}</a> <a href="{% url 'monitoring:specimen_create' item_pk=item.pk %}" title="{% trans 'Add new specimen of this item' %}">{% fa 'plus-square' %}</a></td></tr>
             {% for specimen in item.specimens.physical %}
-                <tr class="specimen"><td>{{ specimen.barcode|default:'—' }}</td><td>{{ specimen.count }}</td><td class="actions"><a href="{% url 'monitoring:specimen_update' pk=specimen.pk %}" title="{% trans 'Edit this specimen' %}">{% fa 'pencil-square' %}</a></td></tr>
+                <tr class="specimen"><td>{{ specimen.pretty_identifier|default:'—' }}</td><td>{{ specimen.count }}</td><td class="actions"><a href="{% url 'monitoring:specimen_update' pk=specimen.pk %}" title="{% trans 'Edit this specimen' %}">{% fa 'pencil-square' %}</a></td></tr>
```
bochecha: I really think it would be better to display both the `barcode` and the `serial` here, just like in #659.

ideascube/monitoring/templates/monitoring/loan.html
mgautierfr: Agree.

ideascube/monitoring/templates/monitoring/specimen_confirm_delete.html
mgautierfr: There is no column here. Do you think we should display both values ?

ideascube/monitoring/templates/monitoring/stock.html
mgautierfr: This is not easy here. There is a table but the rows are shared between the item and the specimens.

- We could split the column (and span the item description cell on two columns). But we have to find nice header's text(s).
- We can just write in the cell something link "Barcode : {{ specimen.barcode|default:'—' }}, Serial number : {{ specimen.serial|default:'—' }}" But it would lead to a lot of duplication of "Barcode :" and "Serial : " and reduce readability.

ideascube/monitoring/templates/monitoring/specimen_confirm_delete.html
bochecha: Indeed, the fact it is a sentence makes it more tricky.

But let's take a step back.

Currently, if someone creates a specimen without a barcode, then when trying to delete it they would see:

screenshot from 2017-01-12 18-18-57

That is extremely unhelpful, to say the least. Basically, that sentence is just as useful as saying:

> Are you sure you want to delete this specimen of "Blah" ?

So I guess we could just say this. 😄 

But then, in the case where there actually is a barcode for the specimen, showing it to the user can reassure them that they didn't misclick. It can also be useful for someone who clicks on the delete link, gets called by someone else, then comes back to that confirmation dialog.

So it is interesting to keep it, maybe.

Now, it is entirely possible to create a specimen with neither a `barcode` nor a `serial`. This surprises me quite a lot to be honest, but that's how the code currently works. Which means that even with the above change, that sentence could still read as unhelpful as above in some cases.

I think this confirmation dialog needs to be redone:

* it needs a cancel/no button
* the title should be something more generic like "Are you sure you want to delete this specimen?"
* the details should be displayed separately, and we should display all the details which can help the user confirm: barcode, serial, quantity, whether it has an associated file, stockitem name, module, description, ...

I also think this is a different change from this one (I'll be happy to open a ticket if you want me to).

Therefore, I'd argue for you to leave this sentence unchanged from master.

ideascube/monitoring/templates/monitoring/stock.html
bochecha: I agree with you the former would read easier.

ideascube/monitoring/forms.py
mgautierfr: WARNING: A long digression not really related to this PR

I would add one or two space between the `if` and the `(` to keep `Loan...` aligned and keep an indentation between `if` and `and` :

```python
if  (A.very.long.line.test()
 and D.very.long.line.test()
 and (B.very.long.line.test()
   or C.very.long.line.test())):
    foo()
```

I agree that is not perfect but I'm not convinced neither by the pep8 solution :

```python
if (A.very.long.line.test()
    and D.very.long.line.test()
    and (B.very.long.line.test()
         or C.very.long.line.test())):
    foo())
```

In this context, I would have prefer the prefix notation, but that's not python :
```python
if and(A.very.long.line.test(),
       D.very.long.line.test(),
       or(B.very.long.line.test(),
          C.very.long.line.test())):
    foo()
```

or a python version (but less readable) :
```python
if all([A.very.long.line.test(),
        D.very.long.line.test(),
        any([B.very.long.line.test(),
             C.very.long.line.test()])]):
    print('a')
```

Maybe the solution is to use intermediate variables and write the `if` condition in one line...
... what I will do :)

bochecha: > WARNING: A long digression not really related to this PR

Oh, absolutely.

And until we collectively agree on a specific style and run the linter in the CI, I won't block a pull request on this.

However, it's already late, I'm curious, and your pull request was a perfect occasion to ask how you prefer aliging long `if`s. 😃 

> I would add one or two space between the `if` and the `(` to keep `Loan...` aligned and keep an indentation between `if` and `and`

My problem with this is that adding a new condition (e.g an `and` clause to the conditional from the code we're are talking under) means modifying the existing lines:

```diff
-        if (Loan.objects.due().filter(specimen__barcode=barcode).exists()
-         or Loan.objects.due().filter(specimen__serial=barcode).exists()):
+        if  (Loan.objects.due().filter(specimen__barcode=barcode).exists()
+         and ...
+          or Loan.objects.due().filter(specimen__serial=barcode).exists()):
```

That's a bigger diff than necessary, making the review that much harder.

> I'm not convinced neither by the pep8 solution

Your example is only one of the solutions suggested in PEP8. However, you're right that's it's extremely unreadable (the conditions being aligned with the inner block), and in fact the `pep8` tool dislikes that one too.

Another solution suggested in PEP8 (and which the `pep8` tool likes) is:

```python
# Add some extra indentation on the conditional continuation line.
if (this_is_one_thing
        and that_is_another_thing):
    do_something()
```

That's the solution I favour.

ideascube/monitoring/templates/monitoring/stock.html
```diff
@@ -11,11 +11,34 @@
...
+                <td>{% trans 'Total:' %} {{ item.specimens.physical|length }}</td>
+                <td class="actions">
+                    <a href="{% url 'monitoring:stockitem_update' pk=item.pk %}" title="{% trans 'Edit' %}">
+                        {% fa 'pencil-square' %}
+                    </a>
+                    <a href="{% url 'monitoring:specimen_create' item_pk=item.pk %}" title="{% trans 'Add new specimen of this item' %}">
+                        {% fa 'plus-square' %}
+                    </a>
+                </td>
+            </tr>
```
bochecha: Oh wow, that was an awfully long line! 😮

ideascube/monitoring/forms.py
```diff
@@ -83,13 +84,14 @@ class InventorySpecimenForm(forms.ModelForm):
 
     def clean_specimen(self):
         barcode = clean_barcode(self.cleaned_data['specimen'])
+        specimen_query = Q(barcode=barcode)|Q(serial=barcode)
         try:
-            specimen = Specimen.objects.get(barcode=barcode)
+            specimen = Specimen.objects.get(specimen_query)
```
bochecha: What if a specimen has its serial set to the same value as another specimen's barcode?

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,71 @@
...
+@migration_test(
+    migrate_from=[('monitoring', '0005_auto_20161027_0801')],
+    migrate_to=[('monitoring', '0006_unique_serial')]
+)
+def test_do_no_break_already_unique_serial(migration):
+    Specimen = migration.old_apps.get_model('monitoring', 'Specimen')
+    StockItem = migration.old_apps.get_model('monitoring', 'StockItem')
+
+    stockItem = StockItem()
+    stockItem.save()
```
bochecha: Dromedary case! 😜

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,71 @@
+from ideascube.tests.helpers import migration_test
+
+
+@migration_test(
+    migrate_from=[('monitoring', '0005_auto_20161027_0801')],
+    migrate_to=[('monitoring', '0006_unique_serial')]
+)
+def test_do_no_break_already_unique_serial(migration):
```
bochecha: Did you mean "do not break"? 😉

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,71 @@
...
+    Specimen = migration.old_apps.get_model('monitoring', 'Specimen')
+    StockItem = migration.old_apps.get_model('monitoring', 'StockItem')
+
+    stockItem = StockItem()
+    stockItem.save()
+
+    for i in range(5):
+        specimen = Specimen(item_id=stockItem.id, barcode=str(i), serial=str(i)*5)
```
bochecha: The `*5` seems a bit gratuitous. `str(i)` would have been just fine.

Not asking you to change it now, it doesn't matter much. It's just that it makes it feel "special". I had to go back to the model to verify whether there was a minimum length. 😃

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,71 @@
...
+)
+def test_create_unique_serial(migration):
+    Specimen = migration.old_apps.get_model('monitoring', 'Specimen')
+    StockItem = migration.old_apps.get_model('monitoring', 'StockItem')
+
+    stockItem = StockItem()
+    stockItem.save()
+
+    for i in range(5):
+        specimen = Specimen(item_id=stockItem.id, barcode=str(i), serial='9'*5)
```
bochecha: `'9'*5` = `99999`. I knows math! 😆

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,71 @@
...
+        specimen.save()
+
+    migration.run_migration()
+
+    Specimen = migration.new_apps.get_model('monitoring', 'Specimen')
+    specimens = Specimen.objects.order_by('barcode')
+    for specimen in specimens:
+        assert specimen.serial.startswith('99999')
```
bochecha: Why not test the whole `serial` including the tilde and the added suffix?

ideascube/monitoring/forms.py
```diff
@@ -112,11 +112,15 @@ class LoanForm(forms.ModelForm):
 
     def clean_specimen(self):
         barcode = self.cleaned_data['specimen']
-        if Loan.objects.due().filter(specimen__barcode=barcode).exists():
+        already_loaned_query = (Q(specimen__barcode=barcode)
+                               |Q(specimen__serial=barcode))
```
bochecha: ~~Same remark as previously about how this might return 2 results.~~ Nope, this is fine here, you use it to test for existence, no problem.

ideascube/monitoring/templates/monitoring/loan.html
```diff
@@ -21,14 +21,18 @@
...
             <th>{% trans "Due date" %}</th>
             <th>{% trans "Comments" %}</th>
             <th>{% trans "By" %}</th>
         </tr>
         {% for loan in loans %}
             <tr>
-                <td>{{ loan.specimen.item }} - {{ loan.specimen.barcode }}</td>
+                <td>{{ loan.specimen.item }}</td>
+                <td>{{ loan.specimen.barcode }}</td>
+                <td>{{ loan.specimen.serial }}</td>
```
bochecha: :+1: on displaying all relevant info here.

ideascube/monitoring/tests/test_views.py
```diff
@@ -525,6 +599,52 @@ def test_can_return_loan(staffapp, user):
...
+    form['specimen'] = specimen.serial
+    form['user'] = user.serial
+    form.submit('do_loan')
+    assert Loan.objects.due().count() == 1
+
+    # Return it.
+    form = staffapp.get(url).forms['return_form']
+    form['loan'] = specimen.barcode
+    form.submit('do_return')
+    assert Loan.objects.due().count() == 0
```
bochecha: Those two are nice functional tests, fully covering corner cases. I like!

ideascube/monitoring/forms.py
```diff
@@ -110,11 +112,15 @@ class LoanForm(forms.ModelForm):
...
+        if Loan.objects.due().filter(already_loaned_query).exists():
             msg = _('Item with barcode {barcode} is already loaned.')
             raise forms.ValidationError(msg.format(barcode=barcode))
+
+        specimen_query = Q(barcode=barcode)|Q(serial=barcode)
```
bochecha: Same remark as previously about how this might return 2 results.

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,71 @@
...
+        for j in range(5):
+            specimen = Specimen(item_id=stockItem.id, barcode=str(i)+str(j), serial=str(i)*5)
+            specimen.save()
+
+    migration.run_migration()
+
+    Specimen = migration.new_apps.get_model('monitoring', 'Specimen')
+    specimens = Specimen.objects.order_by('barcode')
+    for specimen in specimens:
+        assert specimen.serial.startswith(specimen.item.name*5)
```
bochecha: Same question about testing the whole thing.

ideascube/monitoring/tests/test_views.py
bochecha: Mixed tests still missing.

ideascube/monitoring/tests/test_views.py
bochecha: Mixed tests still missing.

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,67 @@
...
+
+@migration_test(
+    migrate_from=[('monitoring', '0005_auto_20161027_0801')],
+    migrate_to=[('monitoring', '0006_unique_serial')]
+)
+def test_create_unique_serial(migration):
+    Specimen = migration.old_apps.get_model('monitoring', 'Specimen')
+    StockItem = migration.old_apps.get_model('monitoring', 'StockItem')
+
+    stockItem = StockItem()
```
bochecha: In an earlier version you were saving the `stockItem` here. Why don't you do it any more?

More importantly, if the `stockItem` is not saved, then `stockItem.id` below iw `None`.

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,67 @@
...
+
+    for i in range(5):
+        specimen = Specimen(item_id=stockItem.id, barcode=str(i), serial='9')
```
bochecha: If you don't save the `specimen` here, then is the migration really doing anything at all?

ideascube/monitoring/tests/test_migration_unique_serial.py
```diff
@@ -0,0 +1,67 @@
...
+
+@migration_test(
+    migrate_from=[('monitoring', '0005_auto_20161027_0801')],
+    migrate_to=[('monitoring', '0006_unique_serial')]
+)
+def test_do_not_break_already_unique_serial(migration):
+    Specimen = migration.old_apps.get_model('monitoring', 'Specimen')
+    StockItem = migration.old_apps.get_model('monitoring', 'StockItem')
+
+    stockItem = StockItem()
```
bochecha: Same questions about saving the `stockItem` and `specimen` as in the other function.

ideascube/monitoring/tests/test_migration_unique_serial.py
mgautierfr: For some strange reason, my brain haven't make the connection between you `Dromedary case` and `camel case`. And for another strange reason, I thought you were saying me that the save was not needed. And as tests passed, it was ok for me.

I've readded the `.save()` and an extra (missing) `assert specimens.count() == 5` to be sure that we are really testing the value of the (saved) specimens.

bochecha: > For some strange reason, my brain haven't make the connection between you Dromedary case and camel case.

Hahahahaha

That should teach me not to try and be funny. 😆 

I thought "CamelCase has two humps, dromedaryCase has only one", and figured it was obvious for everybody. 😜

ideascube/monitoring/tests/test_views.py
bochecha: I see you've now added them, closing this subthread.

PR #676, opened 2017-01-06
Become, and stay, flakes-clean

bochecha:
Let's say this is a new year resolution for the Ideascube project. 😆

### DIFF
--- a/ideascube/conf/dev.py
+++ b/ideascube/conf/dev.py
@@ -4,8 +4,8 @@ from .base import *  # noqa
 
 
 try:
-    os.makedirs(os.path.join(STORAGE_ROOT, 'main'))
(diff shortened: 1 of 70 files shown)


bochecha:
The coverage decrease is due to the fact that this branch removes 2 covered lines: unused imports.

I'm fine with this.

mgautierfr:
Cannot comment en commit message, so I comment here :

```One could even see this as a feature: we (and the CI) can run the tests
with pyflakes to keep a higher code quality, and we can still find
those code smells with the pyflakes command.```

This sentence seems wrong. Are you sure of the second "pyflakes" ?

bochecha:
The problem was the first `pyflakes`.

Fixed.

PR #682, opened 2017-01-12
Update mkdocs to the new yml syntax.

mgautierfr:
Just to fix a warning.

There are other warnings but they come from the not up to date `yeti` theme.
There is no new version of it so we can be OK with the warnings for now or change the theme.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,11 +1,11 @@
 site_name: Ideascube
 pages:
-- [index.md, Home]
-- [install.md, Installation]
-- [config.md, Configuration]
-- [contributing.md, Contributing]
-- [i18n.md, Translations]
-- [release_process.md, Make a Release]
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -1,11 +1,11 @@
...
-- [config.md, Configuration]
-- [contributing.md, Contributing]
-- [i18n.md, Translations]
-- [release_process.md, Make a Release]
+- Home: index.md
+- Installation: install.md
+- Configuration: config.md
+- Contributing: contributing.md
+- Translations: i18n.md
+- Make a Release: release_process.md
```
bochecha: That does look nicer!

However, how am I supposed to know that this is the right syntax when reviewing this pull request?

Please add a link to the relevant documentation page in the commit message, as that's always useful in the future.

PR #683, opened 2017-01-12
Use the `=` sign for the progress bars

bochecha:
That unicode black square makes for very pretty progress bars.

Until you change your font, and the square suddenly has space on its
left and right sides, and these spaces are unaccounted for when
calculating the width, and the line gets broken and it looks terrible.

In addition, we regularly have encoding errors because of it when
running the commands through an SSH connexion, with different locales
installed on the client and the server.

So let's go back to simplicity.

### DIFF
--- a/ideascube/mediacenter/management/commands/import_medias.py
+++ b/ideascube/mediacenter/management/commands/import_medias.py
@@ -19,7 +19,6 @@ from ideascube.management.utils import Reporter
 
 class Bar(ProgressBar):
     template = 'Import: {percent} |{animation}| {done}/{total} | ETA: {eta}'
-    done_char = '⬛'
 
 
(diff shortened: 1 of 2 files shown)


bochecha:
Here's how it looks like on my machine, with the default monospace font in Fedora 25:

screenshot from 2017-01-12 15-00-50

Byoutiphoule !

mgautierfr:
'=' is the default character for the ProgressBar.
We should just remove the two lines instead of redefine the char to the default one.

bochecha:
> '=' is the default character for the ProgressBar.

Indeed, good catch.

And it is documented upstream, so I assume this is something we can count on not changing arbitrarily.

Fixed.

Issue #457, opened 2016-08-31
Delete previous files when reimporting medias


yohanboniface:
At the moment we have the default Django behaviour that just rename the new file if it has the same name, but always keep the first.
The rationale for Django is to say that it's not able to be sure not other object is use this given file.
But in our case we control the flow, so I think it's safe to delete the previous file before uploading a new one for a given media.

PR #686, opened 2017-01-13
Delete the file on the fs when removing the document from media center.

### LINKED ISSUE #684: Removing medias from media center does not erase the file one the dd
When removing a media from the media center, the file is not removed from the hard drive. This is a problem since we have storage issues, especially with the koombook.

mgautierfr:
Fixes #684.

### DIFF
--- a/ideascube/mediacenter/tests/test_models.py
+++ b/ideascube/mediacenter/tests/test_models.py
@@ -1,0 +1,103 @@
+import pytest
+
+from .factories import DocumentFactory
+
+import os.path
+import stat
+
+@pytest.mark.django_db(transaction=True)
+def test_deleting_document_should_delete_file_also():
+    from django.conf import settings
+    file_name = 'foo.img'
+    file_path = os.path.join(settings.MEDIA_ROOT, file_name)
+    with open(file_path, 'w'): pass
+    document = DocumentFactory(original=file_name)
+    document.delete()
+    assert not os.path.exists(file_path)
+
+
+@pytest.mark.django_db(transaction=True)
+def test_deleting_document_should_delete_preview_also():
+    from django.conf import settings
+    file_name = 'preview_foo.img'
+    file_path = os.path.join(settings.MEDIA_ROOT, file_name)
+    with open(file_path, 'w'): pass
+    document = DocumentFactory(preview=file_name)
+    document.delete()
+    assert not os.path.exists(file_path)
+
+
+@pytest.mark.django_db(transaction=True)
+def test_deleting_document_should_delete_file_and_preview_also():
+    from django.conf import settings
+    file_name = 'foo.img'
+    preview_name = 'preview_foo.img'
+    file_path = os.path.join(settings.MEDIA_ROOT, file_name)
+    with open(file_path, 'w'): pass
+    preview_path = os.path.join(settings.MEDIA_ROOT, preview_name)
+    with open(preview_path, 'w'): pass
+    document = DocumentFactory(original=file_name, preview=preview_name)
+    document.delete()
+    assert not os.path.exists(file_path)
+    assert not os.path.exists(preview_path)
+
+
+@pytest.mark.django_db(transaction=True)
+def test_deleting_document_should_not_fail_if_file_already_deleted():
+    from django.conf import settings
+    file_name = 'foo.img'
+    preview_name = 'preview_foo.img'
+    file_path = os.path.join(settings.MEDIA_ROOT, file_name)
+    preview_path = os.path.join(settings.MEDIA_ROOT, preview_name)
+    with open(preview_path, 'w'): pass
+    document = DocumentFactory(original=file_name, preview=preview_name)
+    document.delete()
+    assert not os.path.exists(file_path)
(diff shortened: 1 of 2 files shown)


ideascube/mediacenter/tests/test_models.py
```diff
@@ -0,0 +1,38 @@
...
+    assert remove_mock.call_count == 1
+
+
+@pytest.mark.django_db(transaction=True)
+def test_deleting_document_should_delete_file_and_preview_also(mocker):
+    file_name = 'oo.img'
+    preview_name = 'preview_foo.img'
+    exist_mock = mocker.patch('os.path.exists')
+    exist_mock.side_effect = lambda p: basename(p) in (file_name, preview_name)
+    remove_mock = mocker.patch('os.remove')
```
bochecha: Should you really mock these two?

Mocking them prevents you from testing what happens if an `OSError` gets raised while removing the files.

And in fact, did you test it? What happens if the file to be removed doesn't exist any more, for example? (e.g the admin `rm -fr`-ed it)

Basically, given its place in the transaction (at `on_commit` time), `real_delete` just cannot fail, or else the transaction will be botched.

mgautierfr: > And in fact, did you test it?

Of course I've tested it.

I've checked the code of django's FileField.delete() and it is globaly:
```python
if os.path.exists(path):
    os.remove(path)
```
(https://github.com/django/django/blob/1.10.5/django/core/files/storage.py#L379)
So we have to mock both to be sure that, at end, `os.remove` is called.

However, checking on github, I've just seen that on `master` they use a more pythonist way with a `try/catch` (https://github.com/django/django/blob/master/django/core/files/storage.py#L379)

So we need to have a more complex mock of os.remove than just mocking with a magicmock and we need a second test to be sure that every thing is ok if the file doesn't exist.

bochecha:
> I've checked the code of django's FileField.delete() and it is globaly:
> 
> ```python
> if os.path.exists(path):
>     os.remove(path)

Great, so it's racy. :(

> However, checking on github, I've just seen that on master they use a more pythonist way with a try/catch

It's not just more pythonic, it's the only safe way to do it without a race condition. I'm glad they fixed this.

In this case, I wonder if mocking is the right thing to do.

Why not just actually delete a real file?

I see at least 3 important test cases:

* the file actually exists on the file system, the code deletes it
* the file somehow disappeared from the file system (you remove it manually before the test), the code skips it without error
* the file exists but is a directory (`os.unlink` raises `IsADirectoryError`), the code must still skip it without error

As I said above:

> Basically, given its place in the transaction (at on_commit time), real_delete just cannot fail, or else the transaction will be botched.

I think this is one of the rare cases where a catch-all `except Exception as e` would make sense. Wrap the whole inner code of `real_delete` in a `try` statement, and log the exception, mentioning the file might not have been deleted because of it.

mgautierfr:
> Why not just actually delete a real file?

As the inner code will use different API at next django release, you're right, delete a real file is probably the safer way.

>I see at least 3 important test cases:
>
>  - the file actually exists on the file system, the code deletes it
>  - the file somehow disappeared from the file system (you remove it manually before the test), the code skips it without error
>  - the file exists but is a directory (os.unlink raises IsADirectoryError), the code must still skip it without error

This is highly improbable here. If we especially want to handle this case we should also handle a lot of strange things impossible to handle.

> I think this is one of the rare cases where a catch-all except Exception as e would make sense. Wrap the whole inner code of real_delete in a try statement, and log the exception, mentioning the file might not have been deleted because of it.

Whatever I just said above, I agree that we can be defensive and wrap every thing in a try/except to catch all.
When you speak about mentioning, is it mentioning in the log or add a message to end user ?

bochecha:
> When you speak about mentioning, is it mentioning in the log or add a message to end user ?

Mentioning it to the user could be nice, but it needs to be a non-technical message they can understand, something simple like "The file could not be removed from the disk", or maybe even simpler.

In the logs though, it would be nice to log the full exception traceback, as well as the full path to the file.

This way, when staff try to reclaim some space on their disk by deleting media files and it fails, they'd know something went wrong and they would (let us hope) contact us to do something. We'd have all the infos in the logs to 1. understand what went wrong and fix the code for the future, and 2. remove the file manually from their disk.

bochecha:
We just discussed this, and this seems like too much trouble.

First of all, the `on_delete` signal doesn't handle the update case (#457).

Second, we can't display an error message to the user in the web app if the file can't be deleted, because we're in the model, which mean we don't have access to the `Request` object and thus to the `django.messages` framework.

Which means that in case of errors, Ideascube would silently leave the file on disk, even though we'd have told users that from now on deleting the medias would get the files deleted. 😞 

Finally, this wouldn't handle old boxes which have already had medias removed but the associated files are still on disk.

Therefore, we're now leaning towards a completely different approach: a command which finds stale files and deletes them.

We can then have ansiblecube set up a recurring job of some sort to run that command, and that solves all the above problems.

It's also much simpler.

bochecha:
Superceded by #704, closing.

PR #687, opened 2017-01-16
Removing blog on idb_rwa_mahama

swbsf:
Correct me if needed :)
We may completely remove the blog from ideascube. We need to talk about that with @barbayellow .


barbayellow:
We'll talk about this on Wednesday :)

PR #688, opened 2017-01-16
Setting category for "bouquineux" as "READ".

swbsf:
Pull-me.


bochecha:
@swbsf Looks good.

I assume you tested it and it does the job, merging.

PR #691, opened 2017-01-18
Import from Llave del Saber

### LINKED ISSUE #662: Import data from Llave del Saber
The National Colombian Library gave us a sample of the data we must import. I can upload it here because it's anonymized.
 
We must be able to import these data in ideascube and specifically in the biblioteca movil (ola Hombre) project. 

In the sample uploaded here, we need to import:

* Llave
* Sexo
* Nivel Educativo
* Edad
* Discapacidades
* Ocupaciones

Those we don't need:
* Etnia
* Estrato
* Correo Electrónico
* Recibir Correo

There is a mapping on the 2nd tab of the file.
[llavedelsaber.xlsx](https://github.com/ideascube/ideascube/files/658877/llavedelsaber.xlsx)

This should replace #609

bochecha:
I had to take a bit of a detour in order to do this the way we wanted.

First, this branch introduces a new widget, the ComboBoxEntry, which is inexplicably absent from HTML.

Second, the `current_occupation` and `school_level` fields of the `User` model are changed to accept any value, but use the ComboBoxEntry as a widget so that we still show a default list to try and guide users towards always using the same values. Custom values get added to the list the first time they are used.

Third, a few of these values are added or renamed, as we found the current lists weren't perfect.

Finally, the user import process is made to optionally accept data in the Llave del Saber format, and try mapping their values to ours when it makes sense.

Relates to #662

### DIFF
--- a/ideascube/forms.py
+++ b/ideascube/forms.py
@@ -89,6 +89,15 @@ class UserImportForm(forms.Form):
     def _get_ideascube_mapping(self, data):
         return data
 
+    def _get_llavedelsaber_reader(self, source):
+        return csv.DictReader(source, delimiter=';', quoting=csv.QUOTE_ALL)
+
+    def _get_llavedelsaber_mapping(self, data):
+        # TODO: Implement the rest, once we hear back from Sergio
+        return {
+            'serial': data['serial'],
+        }
+
     def save(self):
         format = self.cleaned_data['format']
         source = TextIOWrapper(self.cleaned_data['source'].file)
--- a/ideascube/forms.py
+++ b/ideascube/forms.py
@@ -108,13 +108,13 @@ class UserImportForm(forms.Form):
         users = []
         errors = []
 
-        for idx, row in enumerate(reader(source)):
+        for idx, row in enumerate(reader(source), start=1):
             try:
                 row = mapper(row)
 
             except KeyError as e:
                 msg = _('Invalid row at line {id}: {field} missing')
-                errors.append(msg.format(id=idx + 1, field=e.args[0]))
+                errors.append(msg.format(id=idx, field=e.args[0]))
                 continue
 
             try:
@@ -129,5 +129,5 @@ class UserImportForm(forms.Form):
                 reason = ', '.join('{}: {}'.format(k, v.as_text())
                                    for k, v in form.errors.items())
                 errors.append(_('Invalid row at line {id}: {reason}').format(
-                    id=idx + 1, reason=reason))
+                    id=idx, reason=reason))
         return users, errors[:10]
--- a/ideascube/forms.py
+++ b/ideascube/forms.py
@@ -70,14 +70,44 @@ class CreateStaffForm(forms.ModelForm):
 class UserImportForm(forms.Form):
     source = forms.FileField(required=True)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        formats = getattr(settings, 'USER_IMPORT_FORMATS', None)
+
+        if formats is None:
+            self.fields['format'] = forms.CharField(
+                required=True, initial='ideascube', widget=forms.HiddenInput)
+
+        else:
+            self.fields['format'] = forms.ChoiceField(
+                required=True, initial=formats[0][0], choices=formats)
+
+    def _get_ideascube_reader(self, source):
+        return csv.DictReader(source)
+
+    def _get_ideascube_mapping(self, data):
+        return data
+
     def save(self):
+        format = self.cleaned_data['format']
         source = TextIOWrapper(self.cleaned_data['source'].file)
         qs = User.objects.all()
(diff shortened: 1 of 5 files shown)


bochecha:
I'm not entirely finished, so please don't merge it yet.

However, things already are in a state where a review would be beneficial. 😃 

Please try and test the new widget.

ideascube/widgets.py
```diff
@@ -3,6 +3,58 @@
...
+            '<div class="comboboxentry">',
+            *rendered_widgets,
+            '<ul>',
+            *(
+                '<li data-value="%s">%s</li>' % (o[0], o[1])
+                for o in self._choices
+            ),
+            '</ul>',
+            '</div>',
+        ]
```
mgautierfr: This is not python 3.4 compatible.
You correct it in the last commit 39dbac but you should amend this one.

ideascube/static/ideascube/js/ideascube.js
```diff
@@ -311,3 +311,81 @@ ID.viewablePassword = function () {
...
+                    visible = true;
+                } else {
+                    option.style.display = 'none';
+                }
+
+                if (visible) {
+                    this.option_list.style.display = 'block';
+                } else {
+                    this.hide_options();
+                }
```
bochecha: Oh wow!

Fixed.

ideascube/widgets.py
bochecha: Thanks, fixed.

bochecha:
At this point, it's pretty much all ready except for the actual Llave del Saber mapping.

@mgautierfr if you have some time and want to re-review all the commits except that last one, I'd be grateful. 😃

bochecha:
Alright, so there were some changes.

I spent some time talking this over with Sergio yesterday, and we've realized that the Llave del Saber export is not entirely exploitable right now. He is going to talk with them, and see if they can make some changes.

As a result, the current exports I could get my hands on might not be structured the way will be eventually.

We concluded that it was better to import only the minimum: the serial numbers. This part of the Llave del Saber exports should not change in the future, so it should be safe.

Importing just the user serial allows having the users in the database, which makes it much easier for the librarians to operate the library.

And once we finalize the Llave del Saber export structure, we can implement importing the rest of the data. At that point, running an import of already imported users will just update them.

This provides us with something right now, as well as a path towards a better future.

As a result, I have dropped from this pull request the `ComboBoxEntry` stuff, since it's not used any more. I'm keeping it handy though, to reintroduce it in the future. 😉

ideascube/forms.py
```diff
@@ -70,14 +70,55 @@ def save(self, *args, **kwargs):
...
+
+        formats = getattr(settings, 'USER_IMPORT_FORMATS', None)
+
+        if formats is None:
+            self.fields['format'] = forms.CharField(
+                required=True, initial='ideascube', widget=forms.HiddenInput)
+
+        else:
+            self.fields['format'] = forms.ChoiceField(
+                required=True, initial=formats[0][0], choices=formats)
```
mgautierfr: Maybe we could check here that we know how to handle the proposed format (check that `'_get_%s_mapping' % format` exists).

This would avoid us to show a option we cannot handle to the user (and so, raise a `ValidationError` the user can't fix).

bochecha: So first, it's a case which would only happen if the box is widely misconfigured. I added it for safety, but it really won't ever happen (I know, "famous last words" 😛)

And indeed, you're right that the `ValidationError` as I did it is quite user-hostile.

However, the `ValidationError` tells the user that something went wrong, and it might push them to come and tell us that something is bad with the box, which would lead us to find (and fix) the misconfiguration. Simply hiding the option as you suggest would silently ignore the misconfiguration issue, users might try to import anyway with the wrong mapping, and things might get ugly real quick.

I just thought about something: how about I add a test to `ideascube/tests/test_settings.py` which ensures that all the declared `USER_IMPORT_FORMATS` have a corresponding handler in the code?

This way, we can just remove the `ValidationError` code path, we just know that it won't happen.

mgautierfr: Hum, yes moving the check in the tests is maybe (probably) the best solution.

ideascube/tests/test_settings.py
```diff
@@ -14,6 +14,12 @@ def setting_module(request):
 
 
 def test_setting_file(setting_module):
+    from ideascube.forms import UserImportForm
+
     settings = importlib.import_module(setting_module, package="ideascube")
 
     assert isinstance(getattr(settings, 'IDEASCUBE_NAME', ''), str)
+
+    for name, label in getattr(settings, 'USER_IMPORT_FORMATS', []):
```
mgautierfr: `label` is not used.
(It is surprising that pyflake doesn't complain about this)

bochecha: > (It is surprising that pyflake doesn't complain about this)

Indeed.

Fixed though.

PR #703, opened 2017-01-26
Misc cleanups

### LINKED ISSUE #674: What's the import_legacy command?
https://github.com/ideascube/ideascube/blob/master/ideascube/management/commands/import_legacy.py

bochecha:
Go at it slowly, commit per commit. I'm happy to split this in multiple pull requests if you prefer, just ask.

First commit removes the Django admin. We don't actually use it, so there's no need to load and expose it.

Second commit fixes #674.

And then, when I opened `ideascube/tests/test_views.py` to remove the tests related to the Django admin, I found a few obvious issues. This prompted me to have a better look at the whole file, and I found quite a few more.

As such, all commits starting from the third one (in topological order) are for cleaning up and improving the tests in this file.


ideascube/tests/test_views.py
```diff
@@ -363,7 +363,9 @@ def test_staff_can_set_password(staffapp, client, user):
     response = form.submit(status=302)
     assert response.location.endswith(
         reverse('user_detail', kwargs={'pk': user.pk}))
-    assert user_model.objects.get(pk=user.pk).password != old_password
```
mgautierfr: Hum, we use `user.password` to get the `old_password`, so the comparison is made on both encrypted password so it should be ok. Or I miss something ?

ideascube/tests/test_views.py
```diff
@@ -363,7 +363,9 @@ def test_staff_can_set_password(staffapp, client, user):
...
+    dbuser = user_model.objects.get(pk=user.pk)
+    assert not dbuser.check_password(old_password)
+    assert dbuser.check_password(password)
```
mgautierfr: Here I agree, we need to use `check_password` as `password` is not encrypted.

ideascube/tests/test_views.py
```diff
@@ -363,7 +363,9 @@ def test_staff_can_set_password(staffapp, client, user):
     response = form.submit(status=302)
     assert response.location.endswith(
         reverse('user_detail', kwargs={'pk': user.pk}))
-    assert user_model.objects.get(pk=user.pk).password != old_password
+    dbuser = user_model.objects.get(pk=user.pk)
+    assert not dbuser.check_password(old_password)
```
mgautierfr: `old_password` will probably be encrypted, so `old_password` will probably check with nothing. We should use the plain old `!=` here.

ideascube/tests/test_views.py
bochecha: Nope, you didn't miss anything... I did. 😕 

I got too caught up in the middle of the other changes, somehow.

Fixed.

Issue #685, opened 2017-01-13
Media center : bulk media deletion


barbayellow:
We **need** a command (maybe it is unwise to add it in the UI) to delete massively medias in the media center.

We need to be able to delete: 
- all the medias 
- medias created between xxx and xxx (xxx is a date)

Additionally, it could be nice to have the ability to delete:
- medias type = video / audio / pdf / ...
- medias title = "insertyoursophisticatedregexhere"

letompouce:
> We *need* a command (maybe it is unwise to add it in the UI) to delete
> massively medias in the media center.

For the time being, I can imagine some ugly hack based on `ideascube
dumpdata` / curl POST / `rm -f`. So please don't let me do that :D

bochecha:
Once #684 is fixed, then a quick Django shell would be nicer:

```python
from ideascube.mediacenter.models import Document

for document in Document.objects.all():
    document.delete()
```

(the above is written entirely from memory, and is untested)

In fact, that's exactly what the command would end up doing.

I'm just mentioning it here to show you that the Django shell is not very complex, and is definitely much nicer than playing with `dumpdata` 😉

PR #704, opened 2017-01-26
Add a command to remove orphan files from the mediacenter.

### LINKED ISSUE #457: Delete previous files when reimporting medias

### LINKED ISSUE #684: Removing medias from media center does not erase the file one the dd

mgautierfr:
This PR add a command to remove orphan files from the mediacenter.

After discussion on PR #686, we agree that a command to remove orphan would be a better idea than
add a listener on the `post_delete` signal.
With a command we handle the delete case (#684) and the update case (#457)

Fixes #684 and Fixes #457

### DIFF
--- a/ideascube/mediacenter/management/commands/clean.py
+++ b/ideascube/mediacenter/management/commands/clean.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+import os
+import argparse
+import glob
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from ideascube.mediacenter.models import Document
+from ideascube.utils import printerr
+
+
+class Command(BaseCommand):
+    help = 'Remove files from the mediacenter.'
+
+    def add_arguments(self, parser):
+        subs = parser.add_subparsers(
+            title='Commands', dest='cmd', metavar='',
+            parser_class=argparse.ArgumentParser)
+        clean_leftover = subs.add_parser(
+            'leftover-files',
+            help='Clean mediacenter files not associated with a document.')
+        clean_leftover.set_defaults(func=self.clean_leftover)
+        parser.add_argument('--dry-run', action='store_true',
+                            help='Print the list of medias that would be '
+                                 'removed. Do not actually remove them')
+
+    def handle(self, *args, **options):
+        if 'func' not in options:
+            self.parser.print_help()
+            self.parser.exit(1)
+
+        options['func'](options)
+
+    def clean_leftover(self, options):
+        files_to_remove = self._get_leftover_files()
+        if options['dry_run']:
+            print("Files to remove are :")
+            for _file in files_to_remove:
+                print(" - '{}'".format(_file))
+        else:
+            for f in files_to_remove:
+                try:
+                    os.unlink(f)
+                except Exception as e:
+                    printerr("ERROR while deleting {}".format(f))
+                    printerr("Exception is {}".format(e))
+
+    def _get_leftover_files(self):
+        # List all (original and preview) files in the fs
+        original_files_root_dir = os.path.join(settings.MEDIA_ROOT,
+                                               'mediacenter/document')
+        files_in_fs = set(
+            glob.iglob(os.path.join(original_files_root_dir, '*')))
+
+        preview_files_root_dir = os.path.join(settings.MEDIA_ROOT,
+                                    'mediacenter/preview')
+        files_in_fs.update(
+            glob.iglob(os.path.join(preview_files_root_dir, '*')))
+
+        # Remove known original paths.
+        original_pathes = Document.objects.all().values_list(
+            'original', flat=True)
+        original_pathes = (os.path.join(settings.MEDIA_ROOT, path)
+                           for path in original_pathes)
+        files_in_fs.difference_update(original_pathes)
+
+        # Remove known preview paths.
+        preview_pathes = Document.objects.all().values_list(
+            'preview', flat=True)
+        preview_pathes = (os.path.join(settings.MEDIA_ROOT, path)
+                          for path in preview_pathes if path)
+        files_in_fs.difference_update(preview_pathes)
+
+        return files_in_fs
--- a/ideascube/mediacenter/tests/factories.py
+++ b/ideascube/mediacenter/tests/factories.py
@@ -4,12 +4,15 @@ import factory
 
 from ..models import Document
 
+class EmptyFileField(factory.django.FileField):
+    DEFAULT_FILENAME = None
 
 class DocumentFactory(factory.django.DjangoModelFactory):
     title = factory.Sequence(lambda n: "Test document {0}".format(n))
     summary = "This is a test summary"
     lang = settings.LANGUAGE_CODE
     original = factory.django.FileField()
+    preview = EmptyFileField()
     credits = "Document credits"
     package_id = ""
 
--- a/ideascube/mediacenter/tests/test_clean.py
+++ b/ideascube/mediacenter/tests/test_clean.py
@@ -1,0 +1,140 @@
+import os, stat
+import pytest
+
+from django.core.management import call_command
+from django.core.files import File
+
+from ideascube.mediacenter.models import Document
+from .factories import DocumentFactory
+
+pytestmark = pytest.mark.django_db
+
+DATA_PATH = os.path.join(os.path.dirname(__file__), 'data')
+
+
+def test_clean_leftover_should_not_fail_if_no_media_or_file():
+    call_command('clean', 'leftover-files')
+
+
+def test_clean_leftover_should_not_remove_files_associated_with_documents(settings):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    original_path = document.original.path
+    preview_path = document.preview.path
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    call_command('clean', 'leftover-files')
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+
+
+def test_clean_leftover_should_not_remove_files_associated_with_documents_without_preview(settings):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview=None)
+    original_path = document.original.path
+    assert os.path.exists(original_path)
+    call_command('clean', 'leftover-files')
+    assert os.path.exists(original_path)
+
+
+def test_clean_leftover_should_correctly_remove_orphan_files(settings):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    original_path = document.original.path
+    preview_path = document.preview.path
+    Document.objects.all().delete()
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    call_command('clean', 'leftover-files')
+    assert not os.path.exists(original_path)
+    assert not os.path.exists(preview_path)
+
+
+def test_clean_leftover_should_fail_nicely_if_cannot_remove(settings, capsys):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    original_path = document.original.path
+    preview_path = document.preview.path
+    document.delete()
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    os.chmod(os.path.join(settings.MEDIA_ROOT, 'mediacenter/document'), stat.S_IRUSR | stat.S_IXUSR)
+    call_command('clean', 'leftover-files')
+    assert os.path.exists(original_path)
+    assert not os.path.exists(preview_path)
+    _, err = capsys.readouterr()
+    assert "ERROR while deleting {}".format(original_path) in err
+
+
+def test_clean_leftover_should_not_remove_orphan_files_if_dry_run(settings, capsys):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    original_path = document.original.path
+    preview_path = document.preview.path
+    Document.objects.all().delete()
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    call_command('clean', 'leftover-files', dry_run=True)
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    out, _ = capsys.readouterr()
+    assert original_path in out
+    assert preview_path in out
+
+
+def test_clean_leftover_should_correctly_clean_after_document_update(settings):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    old_original_path = document.original.path
+    preview_path = document.preview.path
+    with open(os.path.join(DATA_PATH, 'a-pdf.pdf'), 'rb') as f:
+        document.original = File(f, name='a-pdf.pdf')
+        document.save()
+    new_original_path = document.original.path
+    assert os.path.exists(old_original_path)
+    assert os.path.exists(new_original_path)
+    assert os.path.exists(preview_path)
+    call_command('clean', 'leftover-files')
+    assert not os.path.exists(old_original_path)
+    assert os.path.exists(new_original_path)
+    assert os.path.exists(preview_path)
+
+
+def test_clean_leftover_should_correctly_clean_after_complex_situation(settings):
+    document0 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    document1 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    document2 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview=None)
+    document3 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+
+    file_that_should_be_deleted = [
+        document1.original.path,
+        document1.preview.path,
+        document3.original.path,
+        document3.preview.path
+    ]
+    document1.delete()
+    with open(os.path.join(DATA_PATH, 'a-pdf.pdf'), 'rb') as f:
+        document3.original = File(f, name='a-pdf.pdf')
+        document3.save()
+    with open(os.path.join(DATA_PATH, 'an-image.jpg'), 'rb') as f:
+        document3.preview = File(f, name='an-image2.jpg')
+        document3.save()
+
+    files_that_should_be_kept = [
+        document0.original.path,
+        document0.preview.path,
+        document2.original.path,
+        document3.original.path,
+        document3.preview.path
+    ]
+
+    call_command('clean', 'leftover-files')
+    for path in file_that_should_be_deleted:
+        assert not os.path.exists(path)
+    for path in files_that_should_be_kept:
+        assert os.path.exists(path)
+


ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+                try:
+                    os.unlink(f)
+                except Exception as e:
+                    printerr("ERROR while deleting {}".format(f))
+                    printerr("Exception is {}".format(e))
+
+    def _get_orphan_files(self):
+        # List all (original and preview) files in the fs
+        original_files_root_dir = pj(settings.MEDIA_ROOT,
+                                     'mediacenter/document')
```
bochecha: Since you have to break the line anyway, is it really worth aliasing `os.path.join` to `pj`?

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+        try:
+            _, _, original_files_in_fs = next(os.walk(original_files_root_dir))
+        except StopIteration:
```
bochecha: Why not directly `original_files_in_fs = os.listdir(original_files_root_dir)`? 😕

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+            original_files_in_fs = []
+        else:
+            original_files_in_fs = (pj(original_files_root_dir, f)
+                                    for f in original_files_in_fs)
+        files_in_fs = set(original_files_in_fs)
```
bochecha: Why do you need a `set`? Is it just so you can use `.union` and `.difference` later on, or am I missing something else?

(I'm not saying this is wrong, I just want to make sure I actually understood)

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+        preview_files_root_dir = pj(settings.MEDIA_ROOT,
+                                    'mediacenter/preview')
+        try:
+            _, _, preview_files_in_fs = next(os.walk(preview_files_root_dir))
+        except StopIteration:
+            preview_files_in_fs = []
+        else:
+            preview_files_in_fs = (pj(preview_files_root_dir, f)
+                                   for f in preview_files_in_fs)
+        files_in_fs = files_in_fs.union(preview_files_in_fs)
```
bochecha: In case you didn't know, `set1.union(set2)` is strictly equivalent to `set1 | set2`, which then allows you to use `set1 |= set2`.

I personally like it, as it reminds me of the logical OR operator: the union of two sets is the result of OR-ing each of their elements.

(same with `set1 = set1.difference(set2)` and `set1 -= set2`)

But that's a personal preference. However if you prefer the longer form, can I suggest `set1.update(set2)` and `set1.difference_update(set2)`?

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Remove known original pathes.
```
bochecha: I think you meant `paths` :wink:

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+        original_pathes = Document.objects.all().values_list(
+            'original', flat=True)
+        original_pathes = (pj(settings.MEDIA_ROOT, path)
+                           for path in original_pathes)
+        files_in_fs = files_in_fs.difference(original_pathes)
```
bochecha: So you can get the difference of a set and a generator... Interesting. 😃

ideascube/mediacenter/tests/test_remove_medias.py
```diff
@@ -0,0 +1,100 @@
...
+
+def test_clean_orphans_should_not_fail_if_no_media_or_file():
+    call_command('remove_medias', 'clean_orphans')
+
+
+def test_clean_orphans_should_not_remove_files_associated_with_documents(settings):
+    metadata = ('title,summary,credits,path,preview\n'
+                'my video,my video summary,BSF,a-video.mp4,an-image.jpg\n')
+    write_metadata(metadata)
+    call_command('import_medias', CSV_PATH)
```
bochecha: We're sort of moving away from the `import_medias` workflow, so eventually, some time in the (hopefully not too) distant future we might end up completely removing that command.

So maybe we shouldn't add new things which rely on it?

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
```
bochecha: Minor nit: the command does not remove media, it removes leftover files.

So maybe the command (and thus the file) name should reflect that?

ideascube/mediacenter/tests/test_remove_medias.py
```diff
@@ -0,0 +1,100 @@
...
+                'my video,my video summary,BSF,a-video.mp4\n')
+    write_metadata(metadata)
+    call_command('import_medias', CSV_PATH)
+    original_path = os.path.join(settings.MEDIA_ROOT, 'mediacenter/document', 'a-video.mp4')
+    assert os.path.exists(original_path)
+    call_command('remove_medias', 'clean_orphans')
+    assert os.path.exists(original_path)
+
+
+def test_clean_orphans_should_correctly_remove_ophan_files(settings):
```
bochecha: I think you meant `orphan`. 😉

ideascube/mediacenter/tests/test_remove_medias.py
```diff
@@ -0,0 +1,100 @@
...
+    metadata = ('title,summary,credits,path,preview\n'
+                'my video,my video summary,BSF,a-video.mp4,an-image.jpg\n')
+    write_metadata(metadata)
+    call_command('import_medias', CSV_PATH)
+    original_path = os.path.join(settings.MEDIA_ROOT, 'mediacenter/document', 'a-video.mp4')
+    preview_path = os.path.join(settings.MEDIA_ROOT, 'mediacenter/preview', 'an-image.jpg')
+    Document.objects.all().delete()
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    call_command('remove_medias', 'clean_orphans')
```
bochecha: Great, now I imagine you working in an orphanage, giving kids their bath.

More seriously though, "orphan" is standard terminology for this kind of things among developers. But it might not be the most obvious things for users, even command line users. Plus it can bring to mind all sorts of unpleasant ideas.

So perhaps that's not the best string to use, at least in user-visible parts?

ideascube/mediacenter/tests/test_remove_medias.py
```diff
@@ -0,0 +1,100 @@
...
+    assert os.path.exists(preview_path)
+    os.chmod(os.path.join(settings.MEDIA_ROOT, 'mediacenter/document'), stat.S_IRUSR | stat.S_IXUSR)
+    call_command('remove_medias', 'clean_orphans')
+    assert os.path.exists(original_path)
+    assert not os.path.exists(preview_path)
+    _, err = capsys.readouterr()
+    assert "ERROR while deleting {}".format(original_path) in err
+
+
+def test_clean_orphans_should_not_remove_ophan_files_if_dry_run(settings, capsys):
```
bochecha: s/ophan/orphans/

ideascube/mediacenter/tests/test_remove_medias.py
```diff
@@ -0,0 +1,100 @@
...
+    preview_path = os.path.join(settings.MEDIA_ROOT, 'mediacenter/preview', 'an-image.jpg')
+    Document.objects.all().delete()
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    os.chmod(os.path.join(settings.MEDIA_ROOT, 'mediacenter/document'), stat.S_IRUSR | stat.S_IXUSR)
+    call_command('remove_medias', 'clean_orphans')
+    assert os.path.exists(original_path)
+    assert not os.path.exists(preview_path)
+    _, err = capsys.readouterr()
+    assert "ERROR while deleting {}".format(original_path) in err
```
bochecha: I'm glad you tested this. 😃

ideascube/mediacenter/tests/test_remove_medias.py
```diff
@@ -0,0 +1,100 @@
...
+    preview_path = os.path.join(settings.MEDIA_ROOT, 'mediacenter/preview', 'an-image.jpg')
+    Document.objects.all().delete()
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    call_command('remove_medias', 'clean_orphans', dry_run=True)
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    out, _ = capsys.readouterr()
+    assert original_path in out
+    assert preview_path in out
```
bochecha: One test that you're missing is to ensure that leftover files are removed after updating a `Document`.

In addition, it would be nice to test the whole thing on a slightly more complex case, mixing files to remove and files to keep, both in originals and previews.

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+from ideascube.mediacenter.models import Document
+from ideascube.utils import printerr
+
+
+class Command(BaseCommand):
+    help = ('Remove files from the mediacenter.')
```
mgautierfr: Relic of a copy/paste of other command where `help` was on several lines.
I will remove it.

ideascube/mediacenter/management/commands/remove_medias.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def add_arguments(self, parser):
+        subs = parser.add_subparsers(
+            title='Commands', dest='cmd', metavar='',
+            parser_class=argparse.ArgumentParser)
+        remove = subs.add_parser(
+            'clean_orphans',
+            help='Clean mediacenter files not associated with a document.')
+        remove.set_defaults(func=self.clean_orphans)
```
mgautierfr: Hum, this is something I wanted to explain in the PR message but I've forget.

The ideas is to add another subparser to the remove_medias command to solve #685.
I really should have explain it in the PR message and you properly point it in several other comments (I will not answer them).

I concede it complicate a bit this PR and I can remove it if you think it's better.
Having two sub-commands may be just not the right solution. This have to be discuss but if we agree we must have two separated command, there should be no subparser here.

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: os.walk has the advantage to separate directories from files.
Saying that, it should be no subdirectory so `os.listdir` also works and simplify the code (no more `try/except`)

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: Yes, just for the `union` and `difference`

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: The longer form can takes any iterable in input but the operator only works on set.

However, I totally agree with you on the use of `difference_update` instead of `difference`.

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: You read my mind :)

ideascube/mediacenter/tests/test_remove_medias.py
mgautierfr: Yes, a direct document creation is better here.

ideascube/mediacenter/tests/test_remove_medias.py
mgautierfr: Yes. "clean_leftover" maybe ?

ideascube/mediacenter/tests/test_remove_medias.py
mgautierfr: Yes, it's missing two tests.

ideascube/mediacenter/management/commands/remove_medias.py
bochecha: I had a feeling you'd want to add other cleanup operations, but I couldn't think of any, so I preferred asking. 😃 

Having subcommands here is fine, and it makes sense to keep the PR as it is in this regard.

However, do add a note about that in the commit message, please.

Also, I still think that `remove_medias` is not really a representative name for this command. It would be more representative of the new subcommand we'd add for #685 though.

So maybe just `clean`?

```
$ ideascube clean leftover-files
$ ideascube clean all-medias
```

They almost look like English sentences. 😛

ideascube/mediacenter/management/commands/remove_medias.py
bochecha: > os.walk has the advantage to separate directories from files.

Sure, I know that... but you're ignoring everything except the files here. 😄

ideascube/mediacenter/management/commands/remove_medias.py
bochecha: > The longer form can takes any iterable

I gladly admit this surprised me. Is it even documented? 😮

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: Agree with your name proposal.

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: Cause I want the files only. Using the `os.walk` avoid me to filter the result of `os.listdir` to remove directories. (And the internal filter of `os.walk` is equivalent or better optimized than a custom filter)

However, this directory should contain no subdirectory or symlink. Either I assume there is none of them and I can use `os.listdir` without filtering the list or I try to be defensive and I have to do a filtering anyway. So I will change this part to a `os.listdir` and a filtering.

ideascube/mediacenter/management/commands/remove_medias.py
mgautierfr: Yes, I discover it when I tried to use the operator :)

ideascube/mediacenter/management/commands/remove_medias.py
bochecha: > Cause I want the files only. Using the os.walk avoid me to filter the result of os.listdir to remove directories.

Oh right! That's what I had missed. Indeed!

> Either I assume there is none of them and I can use os.listdir without filtering the list or I try to be defensive and I have to do a filtering anyway.

Or, you could use `os.scandir()` which is optimized just for this case: `entry.is_file()` doesn't require an additional system call, contrarily to `os.path.is_file(path)`. But that's Python 3.5 😭

(interesting, I just mistyped `Python 3.5` by forgetting the `shift` key, which gave ` Python ";(`, which I found highly appropriate 😆)

Another option would be to just not care about directories. When you try removing the files, it would raise `IsADirectoryError` for these, and you can just safely ignore that error, because who cares about empty directories anyway. (and as you said, there just won't be any)

ideascube/mediacenter/management/commands/clean.py
```diff
@@ -0,0 +1,89 @@
...
+            parser_class=argparse.ArgumentParser)
+        clean_leftover = subs.add_parser(
+            'leftover-files',
+            help='Clean mediacenter files not associated with a document.')
+        clean_leftover.set_defaults(func=self.clean_leftover)
+        parser.add_argument('--dry-run', action='store_true',
+                            help='Print the list of medias that would be '
+                                 'removed. Do not actually remove them')
+
+    def abort(self, msg):
```
bochecha: This function is entirely unused, please remove it. (or use it)

ideascube/mediacenter/management/commands/clean.py
```diff
@@ -0,0 +1,89 @@
...
+        # List all (original and preview) files in the fs
+        original_files_root_dir = os.path.join(settings.MEDIA_ROOT,
+                                               'mediacenter/document')
+        try:
+            original_files_in_fs = os.listdir(original_files_root_dir)
+        except FileNotFoundError:
+            original_files_in_fs = []
+        original_files_in_fs = (os.path.join(original_files_root_dir, f)
+                                for f in original_files_in_fs)
+        files_in_fs = set(original_files_in_fs)
```
bochecha: I just thought about something: you can of course replace the 7 previous lines by:

```
set(glob.iglob(os.path.join(original_files_root_dir, '*')))
```

ideascube/mediacenter/tests/test_clean.py
```diff
@@ -0,0 +1,141 @@
...
+def test_clean_leftover_should_not_remove_files_associated_with_documents_without_preview(settings):
+    document = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                               preview=None)
+    original_path = document.original.path
+    assert os.path.exists(original_path)
+    call_command('clean', 'leftover-files')
+    assert os.path.exists(original_path)
+
+
+def test_clean_leftover_should_correctly_remove_ophan_files(settings):
```
bochecha: s/ophan/orphan/

ideascube/mediacenter/tests/test_clean.py
```diff
@@ -0,0 +1,141 @@
...
+    assert os.path.exists(preview_path)
+    os.chmod(os.path.join(settings.MEDIA_ROOT, 'mediacenter/document'), stat.S_IRUSR | stat.S_IXUSR)
+    call_command('clean', 'leftover-files')
+    assert os.path.exists(original_path)
+    assert not os.path.exists(preview_path)
+    _, err = capsys.readouterr()
+    assert "ERROR while deleting {}".format(original_path) in err
+
+
+def test_clean_leftover_should_not_remove_ophan_files_if_dry_run(settings, capsys):
```
bochecha: s/ophan/orphan/

ideascube/mediacenter/tests/test_clean.py
```diff
@@ -0,0 +1,141 @@
...
+    call_command('clean', 'leftover-files', dry_run=True)
+    assert os.path.exists(original_path)
+    assert os.path.exists(preview_path)
+    out, _ = capsys.readouterr()
+    assert original_path in out
+    assert preview_path in out
+
+
+def test_clean_leftover_should_correctly_clean_after_document_update(settings):
+    from django.core.files import File
```
bochecha: I recall last time we discussed this, we agreed that local (as in, from the code being tested, i.e ideascube in our case) imports should be done in the test functions, for lots of reasons we both agreed on.

But this is an external import, so maybe import it at the top?

ideascube/mediacenter/tests/test_clean.py
```diff
@@ -0,0 +1,141 @@
...
+    ]
+    document1.delete()
+    with open(os.path.join(DATA_PATH, 'a-pdf.pdf'), 'rb') as f:
+        document3.original = File(f, name='a-pdf.pdf')
+        document3.save()
+    with open(os.path.join(DATA_PATH, 'an-image.jpg'), 'rb') as f:
+        document3.preview = File(f, name='an-image2.jpg')
+        document3.save()
+
+    file_that_should_be_keep = [
```
bochecha: s/keep/kept/

(and file**s** as well 😛 )

ideascube/mediacenter/tests/test_clean.py
```diff
@@ -0,0 +1,141 @@
...
+def test_clean_leftover_should_correctly_clean_after_complex_situation(settings):
+    from django.core.files import File
+    document0 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    document1 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
+    document2 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview=None)
+    document3 = DocumentFactory(original__from_path=os.path.join(DATA_PATH, 'a-video.mp4'),
+                                preview__from_path=os.path.join(DATA_PATH, 'an-image.jpg'))
```
bochecha: Oh, those paths are "before the upload".

You got me for a moment here, I thought the paths in database would be identical, which then made me wonder why the paths from `document1` would end up being deleted, since `document0` and document2` would still reference them.

In fact, that's one more good argument in favour of having it done this way rather than in the ORM/forms: we could have removed the path of a document when deleting another document if the two documents shared a path!

ideascube/mediacenter/tests/factories.py
```diff
@@ -4,12 +4,15 @@
...
 
+class EmptyFileField(factory.django.FileField):
+    DEFAULT_FILENAME = None
 
 class DocumentFactory(factory.django.DjangoModelFactory):
     title = factory.Sequence(lambda n: "Test document {0}".format(n))
     summary = "This is a test summary"
     lang = settings.LANGUAGE_CODE
     original = factory.django.FileField()
+    preview = EmptyFileField()
```
bochecha: Seems the only alternative would have been to change all the existing tests so that they'd pass `preview=None` when instanciating the factory.

meh...

ideascube/mediacenter/tests/test_clean.py
mgautierfr: It's the ORM/forms who create the files in MEDIA_ROOT (and ensure that two documents cannot share the same file even if they share the same **original** path)

ideascube/mediacenter/tests/factories.py
mgautierfr: > Seems the only alternative would have been to change all the existing tests so that they'd pass preview=None when instanciating the factory?

Yes :/ I've preferred to not change the API of the factory.

bochecha: >> Seems the only alternative would have been to change all the existing tests so that they'd pass preview=None when instanciating the factory?
> 
> Yes :/ I've preferred to not change the API of the factory.

Well, the API wouldn't have changed, it's just that tests which really don't want a `preview` had to set it themselves.

In fact, with your change, you changed the behaviour of the factory for all the previously existing tests.

I do think this is a much more sensible default value though, since it's the same as the one from the model (as `preview` has `blank=True`).

mgautierfr: > it's just that tests which really don't want a preview had to set it themselves.
That's what I call a API change :)

Issue #707, opened 2017-01-27
Increase media-center upload max file size


fheslouin:
Currently the max upload file size is 200M. We can go to 500M as tests has been done and Nginx + Uswgi can handle this.

fheslouin:
See PR #708

bochecha:
Fixed by #708

PR #708, opened 2017-01-27
Increase upload max file size on Nginx

fheslouin:
Can it be merge in the next release ? 

Thanks

### DIFF
--- a/extras/nginx/ideascube
+++ b/extras/nginx/ideascube
@@ -15,7 +15,7 @@ server {
     charset     utf-8;
 
     # max upload size
-    client_max_body_size 200M;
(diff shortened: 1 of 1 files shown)


bochecha:
@fheslouin for next time, you can write `Fixes #707` in your commit message, and github will close the ticket automatically when we merge. 😃 

In this case, that would have given the following message:

```
Increase upload max file size on Nginx

Fixes #707
```

No worries for this time, I'm merging and closing the ticket manually, but I just wanted to let you know we can save some manual work. 😉

fheslouin:
Good to know, thanks ;)

PR #709, opened 2017-01-27
Fix for idb_jor_hostcommunity user form fields.

letompouce:
Narjis says the hostcommunity conf should import its USER_FORM_FIELDS
from alhashmi instead of azraq's ones.

### DIFF
--- a/ideascube/conf/idb_jor_hostcommunity.py
+++ b/ideascube/conf/idb_jor_hostcommunity.py
@@ -1,4 +1,4 @@
 """Ideasbox for Host Community, Jordan"""
-from .idb_jor_azraq import *  # pragma: no flakes
+from .idb_jor_alhashmi import *  # pragma: no flakes
 
 ENTRY_ACTIVITY_CHOICES = []


bochecha:
> Narjis says the hostcommunity conf should import its USER_FORM_FIELDS
from alhashmi instead of azraq's ones.

But your change imports **everything** from the alhashmi config.

Right now, the alashmi config imports the azraq one and just overrides the `USER_FORM_FIELDS` so that achieves your goal.

However we have **5** levels of configuration inheritance here: hostcommunity inherits from alashmi which inherits from azraq which inherits from idb which inherits from base.

This looks like a recipe for trouble if we're not careful.

letompouce:
cc @Narjis89

letompouce:
Our options so far:
* Repeat the `USER_FORM_FIELDS` tweak in both `idb_jor_alhashmi.py` & `idb_jor_hostcommunity.py`. Redundant code.
* Create a generic `idb_jor.py` file to import in all the others. This only adds another layer of inclusion. Not to talk about the future JOR projects and their potential willing to tweak `USER_FORM_FIELDS` or something else.
* Live with it and be careful.

I'd tend to apply the third option.

But I wanna check about the `ENTRY_ACTIVITY_CHOICE` micmac as well before we merge...

Narjis89:
I'm with choice number 3 also 
And for the  ENTRY_ACTIVITY_CHOICE no one has any complain about it , we can keep it as it

bochecha:
Alright, thanks for thinking about this, let's go with it. 😃

Issue #701, opened 2017-01-26
Better try and weed out silly tags


bochecha:
Some people seem to not understand the "tags" field in the new media form, and just type a description sentence.

There are a few things we could do to limit the problems caused by that:

* separate punctuation from words

    We have some tags like `foo;`, which we should tokenize into `foo` and `;`. Then we can just remove tags that are purely punctuation.

* try lower-casing the tag name

    This could help de-duplicate some tags.

A bigger effort might be needed on the UX of this field, but that's for another ticket.

PR #710, opened 2017-02-01
Mediacenter misc

mgautierfr:
Several small commits on the mediacenter to improve it.
Fixes few already opened bugs.

### DIFF
--- a/ideascube/mediacenter/models.py
+++ b/ideascube/mediacenter/models.py
@@ -67,7 +67,9 @@ class Document(SearchMixin, TimeStampedModel):
                             default=OTHER)
 
     objects = DocumentQuerySet.as_manager()
-    tags = TaggableManager(blank=True, manager=SortedTaggableManager)
+    tags = TaggableManager(verbose_name=_('Topics'),
+                           blank=True,
+                           manager=SortedTaggableManager)
 
     package_id = models.CharField(verbose_name=_('package'), max_length=100,
                                   blank=True)
--- a/ideascube/mediacenter/models.py
+++ b/ideascube/mediacenter/models.py
@@ -55,7 +55,8 @@ class Document(SearchMixin, TimeStampedModel):
     lang = LanguageField(verbose_name=_('Language'), max_length=10, blank=True)
     original = models.FileField(verbose_name=_('Source file'),
                                 upload_to='mediacenter/document',
-                                max_length=10240)
+                                max_length=10240,
+                                help_text=_('Maximum size : 500 Mo.'))
     preview = models.ImageField(verbose_name=_('Preview picture'),
                                 upload_to='mediacenter/preview',
                                 max_length=10240,
--- a/ideascube/mediacenter/models.py
+++ b/ideascube/mediacenter/models.py
@@ -60,7 +60,7 @@ class Document(SearchMixin, TimeStampedModel):
                                 upload_to='mediacenter/preview',
                                 max_length=10240,
                                 blank=True)
-    credits = models.CharField(verbose_name=_('credit'), max_length=300)
+    credits = models.CharField(verbose_name=_('Authorship'), max_length=300)
     kind = models.CharField(verbose_name=_('type'),
                             max_length=5,
                             choices=KIND_CHOICES,
--- a/ideascube/mediacenter/models.py
+++ b/ideascube/mediacenter/models.py
@@ -53,10 +53,10 @@ class Document(SearchMixin, TimeStampedModel):
     title = models.CharField(verbose_name=_('title'), max_length=100)
     summary = models.TextField(verbose_name=_('summary'))
     lang = LanguageField(verbose_name=_('Language'), max_length=10, blank=True)
-    original = models.FileField(verbose_name=_('original'),
+    original = models.FileField(verbose_name=_('Source file'),
                                 upload_to='mediacenter/document',
                                 max_length=10240)
-    preview = models.ImageField(verbose_name=_('preview'),
+    preview = models.ImageField(verbose_name=_('Preview picture'),
                                 upload_to='mediacenter/preview',
                                 max_length=10240,
                                 blank=True)
--- a/ideascube/mediacenter/templates/mediacenter/index.html
+++ b/ideascube/mediacenter/templates/mediacenter/index.html
@@ -34,5 +34,5 @@
     {% include 'search/order_by.html' %}
     {% include 'search/filter_by_kind.html' %}
     {% include 'search/filter_by_lang.html' %}
-    {% tag_cloud url='mediacenter:index' model=view.model tags=available_tags %}
+    {% tag_cloud url='mediacenter:index' model=view.model limit=100 tags=available_tags %}
 {% endblock third %}
--- a/ideascube/templates/ideascube/includes/tag_cloud.html
+++ b/ideascube/templates/ideascube/includes/tag_cloud.html
@@ -1,7 +1,7 @@
 {% load i18n ideascube_tags %}
 {% spaceless %}
 <div class="card tinted">
-    <h4><span class="theme learn">{% trans "browse" %}</span> {% trans "Topics" %}</h4>
+    <h4><span class="theme discover">{% trans "browse" %}</span> {% trans "Topics" %}</h4>
     {% for tag in tags %}
         {% if tag.slug %}
             {% is_in_qs 'tags' tag.slug as in_qs %}
--- a/ideascube/tests/test_utils.py
+++ b/ideascube/tests/test_utils.py
@@ -2,7 +2,7 @@ from io import BytesIO
 
 import pytest
 
-from ideascube.utils import to_unicode, get_file_sha256, get_file_size
+from ideascube.utils import to_unicode, get_file_sha256, get_file_size, tag_splitter
 from hashlib import sha256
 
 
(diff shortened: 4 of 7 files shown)


ideascube/mediacenter/models.py
```diff
@@ -60,7 +60,7 @@ class Document(SearchMixin, TimeStampedModel):
                                 upload_to='mediacenter/preview',
                                 max_length=10240,
                                 blank=True)
-    credits = models.CharField(verbose_name=_('credit'), max_length=300)
+    credits = models.CharField(verbose_name=_('autorship'), max_length=300)
```
bochecha: It's spelled "authorship".

This is a user-visible string, and it's also visible by translators, so it's important to get it right.

bochecha: I'm wondering about the all lower-casing here... In other `verbose_name`s we sometimes capitalize the first letter.

Maybe we should make a pass on all these at once eventually.

ideascube/mediacenter/forms.py
```diff
@@ -17,6 +18,9 @@ class Meta:
         widgets = {
             'lang': LangSelect,
         }
+        labels = {
+            'original': _('Source file (maximum size : 500 Mo)'),
```
bochecha: Sneaked the limit in there, didn't you? 😆 

Shouldn't the limit instead be in a help text?

And why not change the `verbose_name` on the model instead?

ideascube/templates/ideascube/includes/tag_cloud.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n ideascube_tags %}
 {% spaceless %}
 <div class="card tinted">
-    <h4><span class="theme learn">{% trans "browse" %}</span> {% trans "Topics" %}</h4>
+    <h4><span class="theme discover">{% trans "browse" %}</span> {% trans "Topics" %}</h4>
```
bochecha: I don't understand why this commit is necessary. Can you provide a before/after screenshot? Or maybe explain where in the application are those "other titles" displayed, so I can go and look at the difference for myself?

ideascube/mediacenter/templates/mediacenter/index.html
```diff
@@ -34,5 +34,5 @@
     {% include 'search/order_by.html' %}
     {% include 'search/filter_by_kind.html' %}
     {% include 'search/filter_by_lang.html' %}
-    {% tag_cloud url='mediacenter:index' model=view.model tags=available_tags %}
+    {% tag_cloud url='mediacenter:index' model=view.model limit=40 tags=available_tags %}
```
bochecha: I'm not sure that's really going to mitigate the issue all that much. :-/

```
$ for DB in *.sqlite; do sqlite3 $DB 'select count(id) from taggit_tag'; done
132
119
160
452
348
124
124
172
119
125
```

Granted, most of these tags are wrong (duplicates like `Sylvain` and `Sylvain;`, punctuation-only tags like `.`, etc.). But if we don't display them, library staff won't even know those bad tags exist and need to be fixed.

I'm not sure we can truly fix the issue without thinking hard about the form UI (auto-completion, better display of tags as they are parsed, ...), but simply the limit to 40 seems like it wouldn't achieve much at all. :-/

Maybe we should bump to 100? Maybe we should order the tags by most "frecent"? (i.e some heuristics based on "frequency" and "recency" of use)

(removing the limit entirely and displaying the 452 tags would make Ideascube look terrible, and we probably shouldn't do it)

ideascube/templates/ideascube/includes/tag_cloud.html
mgautierfr: That was a ask from @swbsf from internal BSF people (we should have open a bug for that)

Before : 
capture_tag_different_color

After :
capture_tag_same_color

ideascube/mediacenter/templates/mediacenter/index.html
mgautierfr: Hum, with so much tags, yes, change the limit to 40 will not help a lot :/

I agree with you on your other remarks. I especially like the "frecent" order, but it may be tricky to implement.

We really have to do something about this, all the easy solutions we've got do not fix anything on this point and it is a real problem for users.

ideascube/mediacenter/forms.py
mgautierfr: Yes, it is better to have it in the help text.
I didn't change the verbose_name as this can be use elsewhere and the size limit make sense only in the upload form. (I don't know where it is used (and if), but it could)

bochecha: Right, but this is kind of the django way: forms are generated from the model attributes. (it's horrendous if you ask me, but it's idiomatic)

Models even have the [`help_text`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#help-text) attribute for forms! 🙄 

I don't really see any reason not to modify the model's `verbose_name` here. Especially since it already has a `verbose_name`, which means translators would now have to translate both "original" (the `verbose_name`) and "source file" (the `label`)

ideascube/templates/ideascube/includes/tag_cloud.html
bochecha: Oh, I see. That's an eyesore indeed. 😆 

What tripped me up is that the CSS classes use the `discover` and `leanr` stuff from the home cards, which is really an abuse.

Anyway, changing all that would be too big a change, let's wait for the UI rework.

This specific change is approved, end of the thread as far as I'm concerned.

ideascube/mediacenter/templates/mediacenter/index.html
bochecha: Proposal: We bump to 100 for now, we work on #701 (the splitting), and we wait for the designers to help us with a UI rework? The UI rework might entail JS (autocompletion, display of chosen tags in "boxes", ...), or a new page "view all tags", or who knows what. At that point we might revisit the 100 limit, lower it down, remove it, or anything else.

ideascube/tests/test_utils.py
```diff
@@ -55,3 +55,30 @@ def test_size(tmpdir):
...
+    'black & white,stuff,fuzzy',
+    'black & white;fuzzy;stuff',
+    ';stuff; black & white,stuff,fuzzy',
+    '    black & white   ,,,, stuff  ,fuzzy',
+])
+def test_tag_splitter_tags_with_spaces(string):
+    oracle = set(['black & white', 'stuff', 'fuzzy'])
+    tags = tag_splitter(string)
+    assert len(tags) == len(oracle)
+    assert set(tags) == oracle
```
bochecha: Nice tests.

Great use of `parametrize`, so we can add more weird cases as we find (and fix) issues in the future.

ideascube/utils.py
```diff
@@ -87,3 +88,7 @@ def get_file_sha256(path):
 
 def get_file_size(path):
     return os.stat(path).st_size
+
+def tag_splitter(tag_string):
+    tags = set(t.strip().lower() for t in re.split('[;,]+', tag_string) if t.strip())
```
bochecha: I'm not sure we should lower-case the tags like this. 😕 

It has the potential to be one of these "features" which feels strongly like a bug for users: "come on, I keep using a capital letter and this dumb thing lower-cases it!!!!" followed by cries of rage.

Can't we lower-case the slug, to avoid duplicates, but keep the visible value the exact way the user typed it?

ideascube/mediacenter/forms.py
```diff
@@ -17,6 +18,9 @@ class Meta:
         widgets = {
             'lang': LangSelect,
         }
+        help_texts = {
+            'original': _('Maximum size : 500 Mo.'),
+        }
```
bochecha: Why not put the `help_text` on the model field?

I know it feels wrong... but it's the idiomatic Django way...

ideascube/utils.py
mgautierfr: > Can't we lower-case the slug, to avoid duplicates, but keep the visible value the exact way the user typed it?

I'm not sure we can do this. Both `name` and `slug` are "`unique`" field in the `Tag` model.
We could not have a `Tag(name="Bar", slug="bar")` and a `Tag(name="bar", slug="bar")` at the same time.
And the user will asking itself : 
> Why sometime the case isn't changed [no tag already exist with the same slug] and sometime the case is changed, sometime lowering everything, sometime "titleling" it [the tag already exist] ?

ideascube/mediacenter/forms.py
mgautierfr: It's not cause django's people does strange things that we have to do the same ;)

This `help_texts` is only useful in this specific form (upload form), not everywhere.
`help_texts` is a "valid" way to set the help text only for one form.

However, there is no other form using this model, so it is not so important for now. I will not argue too long on this. If you prefer to have the `help_text` in the model, I will do it. Make as every one has advantage also :)

ideascube/utils.py
bochecha: > We could not have a Tag(name="Bar", slug="bar") and a Tag(name="bar", slug="bar") at the same time.

I know, and that's a good thing.

The reason I had suggested lower-casing the tags in the first place was to deduplicate them.That is, if someone entered "Cinema" as a tag for a media document, and then "cinema" as a tag for another document, the goal was to consider them the same tag.

You're right that the user experience this provides is tricky (if not impossible) to get right. It has the exact same problems as I was complaining with your lower-casing everything.

Therefore, let's not lower-case at all.

ideascube/utils.py
```diff
@@ -87,3 +88,7 @@ def get_file_sha256(path):
 
 def get_file_size(path):
     return os.stat(path).st_size
+
+def tag_splitter(tag_string):
+    tags = set(t.strip() for t in re.split('[;,]+', tag_string) if t.strip())
```
bochecha: Should you make this a raw-string with `r'...'`?

It wouldn't change anything with the regex you used, but if we end up modifying it in the future it might make for some hilarious debugging.

PR #712, opened 2017-02-01
Cosmetics

swbsf:
Cosmetics.

### DIFF
--- a/ideascube/locale/fr/LC_MESSAGES/django.po
+++ b/ideascube/locale/fr/LC_MESSAGES/django.po
@@ -1578,11 +1578,11 @@ msgid "browse"
 msgstr "parcourir"
 
 #: ideascube/search/templates/search/filter_by_kind.html:4
-msgid "by kind"
-msgstr "par genre"
+msgid "By kind"
+msgstr "Par genre"
 
 #: ideascube/search/templates/search/filter_by_lang.html:4
-msgid "by lang"
+msgid "By lang"
 msgstr "Par langue"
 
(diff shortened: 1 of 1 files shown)


mgautierfr:
We must not modify `.po` by hand.

- If the original string is wrong, you should modify it at (in) the source, where we use and declare it.
- If the translation is wrong, you must modify it on transifex (https://www.transifex.com/ideascube).

The `.po` are extracted from transifex and any modification we made by hand will be overwrited the next time we update the translation from transifex.

I also wonder if this is a good idea to change the source string for such a small changes.
The source string are used as id to find the translated string. If we change them, we have to update all translations. It's a tradeoff problem and I don't especially know what is the best idea. Maybe @bochecha have a clue on this.

swbsf:
oh ok :(

bochecha:
Agreed with @mgautierfr here: those are valid fixes, done in the wrong place.

@swbsf Can you go to Transifex and translate the `Thèmes` one there?

As for modifying the source strings in the code... It goes with [my recent comment on another pull request](): we definitely lack consistency in these strings and we should have a go at fixing that at some point.

But as @mgautierfr mentioned, that invalidates all existing translations, so we should do it carefully and sparingly.

Maybe we should wait for the big UI rework for these, since we're most likely going to have lots of string changes at that point.

---

I'm going to close this since there is nothing to do in Ideascube for the issues you found @swbsf. Thanks for having submitted it though, it shows you pay attention to the little details, which is a great thing as they often end up making most of the impression users have on a product. 😃

PR #715, opened 2017-02-03
Fix the clean command

bochecha:
Without this, it couldn't print its own help text when invoked without a
subcommand.


mgautierfr:
Nothing to say but it's ok for me.

PR #716, opened 2017-02-09
Add a command to remove all media from the mediacenter.

mgautierfr:
Media installed with a package are not removed.

Fix partially #685.

### DIFF
--- a/ideascube/mediacenter/management/commands/clean.py
+++ b/ideascube/mediacenter/management/commands/clean.py
@@ -17,13 +17,17 @@ class Command(BaseCommand):
         subs = parser.add_subparsers(
             title='Commands', dest='cmd', metavar='',
             parser_class=argparse.ArgumentParser)
+
+        dry_run = argparse.ArgumentParser('dry_run', add_help=False)
+        dry_run.add_argument('--dry-run', action='store_true',
+                             help='Print the list of medias that would be '
+                                  'removed. Do not actually remove them')
+
         clean_leftover = subs.add_parser(
             'leftover-files',
+            parents = [dry_run],
             help='Clean mediacenter files not associated with a document.')
         clean_leftover.set_defaults(func=self.clean_leftover)
-        parser.add_argument('--dry-run', action='store_true',
-                            help='Print the list of medias that would be '
-                                 'removed. Do not actually remove them')
 
         self.parser = parser
 
--- a/ideascube/mediacenter/management/commands/clean.py
+++ b/ideascube/mediacenter/management/commands/clean.py
@@ -29,6 +29,12 @@ class Command(BaseCommand):
             help='Clean mediacenter files not associated with a document.')
         clean_leftover.set_defaults(func=self.clean_leftover)
 
+        clean_media = subs.add_parser(
+            'media',
+            parents = [dry_run],
+            help='Remove all medias')
+        clean_media.set_defaults(func=self.clean_media)
+
         self.parser = parser
 
     def handle(self, *args, **options):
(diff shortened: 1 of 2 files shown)


ideascube/mediacenter/management/commands/clean.py
```diff
@@ -17,13 +17,19 @@ def add_arguments(self, parser):
...
+                            help='Print the list of medias that would be '
+                                 'removed. Do not actually remove them')
+
         clean_leftover = subs.add_parser(
             'leftover-files',
             help='Clean mediacenter files not associated with a document.')
         clean_leftover.set_defaults(func=self.clean_leftover)
-        parser.add_argument('--dry-run', action='store_true',
-                            help='Print the list of medias that would be '
-                                 'removed. Do not actually remove them')
```
bochecha: This means the command now is:

```
$ ideascube clean --dry-run leftover-files
$ ideascube clean --dry-run media
```

However, the `--dry-run` option applies to the subcommand, so I'd much prefer we had:

```
$ ideascube clean leftover-files --dry-run
$ ideascube clean clean --dry-run
```

You can do that with [a common parser](https://github.com/ideascube/ideascube/blob/87aa1c444dde784fd7fbf6ae3a09d040e2046bd8/ideascube/serveradmin/management/commands/catalog.py#L28-L33) and have it be [a parent of the commands](https://github.com/ideascube/ideascube/blob/87aa1c444dde784fd7fbf6ae3a09d040e2046bd8/ideascube/serveradmin/management/commands/catalog.py#L59).

It's a better UX, because it means you can run:

```
$ ideascube catalog clean media --dry-run
```

And then quickly press ↑ followed by `ctrl`+`w` then ↵.

(it would also be an API break if we had released the existing command, or if the Ansiblecube folks had already created the cron job for the `leftover-files`)

ideascube/mediacenter/management/commands/clean.py
```diff
@@ -32,6 +38,16 @@ def handle(self, *args, **options):
...
         options['func'](options)
 
+    def clean_media(self, options):
+        Document.objects.filter(package_id='').delete()
+        left_media_count = Document.objects.all().count()
+        if left_media_count:
+            print("{} media have been installed by packages."
+                  " They have been not deleted.\n"
+                  "You must delete the corresponding package if you want to "
+                  "remove them.".format(left_media_count))
```
bochecha: Maybe print the command? `'%s catalog remove pkgid [pkgid ...]' % args[0]` (not sure how to get the program name)

ideascube/urls.py
```diff
@@ -43,3 +43,4 @@
         name='jsi18n'),
     url(r'^i18n/', include('django.conf.urls.i18n')),
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+
```
bochecha: Huh?

ideascube/mediacenter/management/commands/clean.py
```diff
@@ -38,6 +45,19 @@ def handle(self, *args, **options):
...
+        Document.objects.filter(package_id='').delete()
+        left_media_count = Document.objects.all().count()
+        if left_media_count:
+            print("{left_media_count} media have been installed by packages."
+                  " They have been not deleted.\n"
+                  "You must delete the corresponding package if you want to "
+                  "remove them. Use the command :\n"
+                  "{program_name} catalog remove pkgid+".format(
+                      left_media_count=left_media_count,
+                      program_name=sys.argv[0]))
```
bochecha: Does `sys.argv[0]` return the right thing with `manage.py` and `ideascube`?

mgautierfr: Alas, not in all case :/

I've removed the `sys.argv[0]`

PR #717, opened 2017-02-09
Use a non-UTF8 locale to run the tests

bochecha:
The default encoding on Python 3 is the one from the current locale.
This almost always means UTF-8, at least for our target deployment
(Debian Jessie with a UTF-8 locale)

And that's a great thing. It's not necessary to think about encoding
issues most of the time, and that's liberating.

However, BSF manages the boxes in two ways.

The first is to connect to them from their (usually Debian, Ubuntu or
Mint) workstations with SSH, through a bastion. The default
configuration for the SSH client on these OS makes the remote session
use the locale of the local session (usually fr_FR.UTF8).

But on Debian, not all locales are installed by default. This means that
the remote session will try to use a locale (from the local session)
which is not installed on the server.

When that happens, the session falls back to the C locale, for which the
default encoding is ASCII.

The second management method is through a NetworkManager dispatcher
script, which runs ansible-pull to get the latest ansiblecube state and
run it. This also tends to run in a C locale, for which the default
encoding is ASCII.

This means that in Ideascube we can't be our usual lazy Python
developers and assume a UTF-8 encoding. We must instead start specifying
explicitly the correct encoding.

This commit makes sure we run the tests under the C locale, and
therefore with the ASCII default encoding, to catch lots of these
issues.

It won't catch all of them, but it will help us eliminate some of them
and get into the habit of always explicitly specifying the UTF-8
encoding.

And when we find an issue in production which wasn't caught by tests, we
can add more tests to prevent it from coming back.

---

And in fact, it already catches some, so the subsequent commits fix them. 😄

### DIFF
--- a/ideascube/serveradmin/tests/test_catalog.py
+++ b/ideascube/serveradmin/tests/test_catalog.py
@@ -16,6 +16,11 @@ from ideascube.mediacenter.models import Document
             'url': 'http://foo.fr/catalog.yml',
         },
         {
+            'id': 'bibliothèque',
+            'name': 'Le contenu de la bibliothèque',
+            'url': 'http://foo.fr/catalog.yml',
+        },
+        {
             'name': 'Content provided by Foo',
             'url': 'http://foo.fr/catalog.yml',
         },
@@ -30,6 +35,7 @@ from ideascube.mediacenter.models import Document
     ],
     ids=[
         'foo',
+        'utf8',
         'missing-id',
         'missing-name',
         'missing-url',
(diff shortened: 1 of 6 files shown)


ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -30,21 +35,27 @@
...
 
     if 'name' in request.param:
-        path.write('name: {name}\n'.format(**request.param), mode='a')
+        lines.append('name: {name}\n'.format(**request.param))
 
     if 'url' in request.param:
-        path.write('url: {url}'.format(**request.param), mode='a')
+        lines.append('url: "{url}"'.format(**request.param))
+
+    path.write_text('\n'.join(lines), encoding='utf-8')
```
mgautierfr: line in `lines` already have a `\n`.
Either we just `''.join(...)` or we remove the `\n` in the `lines.append` before.

ideascube/serveradmin/tests/test_catalog.py
```diff
@@ -130,6 +141,31 @@ def test_remote_to_file(tmpdir):
...
+    from ideascube.serveradmin.catalog import Remote
+
+    path = tmpdir.join('foo.yml')
+
+    remote = Remote(
+        'bibliothèque', 'Le contenu de la bibliothèque',
+        'http://foo.fr/catalog.yml')
+    remote.to_file(path.strpath)
+
+    lines = path.readlines(cr=False)
```
mgautierfr: No encoding here ?

ideascube/serveradmin/tests/test_catalog.py
bochecha: Right, fixed.

Somehow I also left some debug `print` in there, removed as well.

ideascube/serveradmin/tests/test_catalog.py
bochecha: Huh. I'm surprised this works, indeed. I guess `.readlines()` must be doing something magic... 😕 

I'm not super comfortable witht hat, so I've replaced it by a more explicit `read_text('utf-8').split('\n')`.

PR #718, opened 2017-02-09
Check that index file is really present before adding it to library.xml

mgautierfr:
Not all zim come with a corresponding index.
If there is no index file, we should not add it to library xml.

### DIFF
--- a/ideascube/serveradmin/catalog.py
+++ b/ideascube/serveradmin/catalog.py
@@ -163,7 +163,10 @@ class Kiwix(Handler):
 
                 book = books[0]
                 book.set('path', 'data/content/%s' % zimname)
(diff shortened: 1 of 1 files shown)


ideascube/serveradmin/catalog.py
```diff
@@ -163,7 +163,10 @@ def commit(self):
 
                 book = books[0]
                 book.set('path', 'data/content/%s' % zimname)
-                book.set('indexPath', 'data/index/%s.idx' % zimname)
+
+                index_path = 'data/index/%s.idx' % zimname
+                if os.path.exists(os.path.join(self._install_dir, index_path)):
```
bochecha: `os.path.isdir` ?

PR #720, opened 2017-02-10
Handle our dependencies with pip-tools

bochecha:
This is pretty nifty.

Keeping our dependencies up to date is as simple as running:

    make upgrade-deps

And it generates **complete** pip requirements files, so that we know
exactly what gets installed, everybody (including the Debian packages)
gets the same versions of everything, no matter when they installed.

If we need to add a new dependency, or remove an old one, we do that in
the new .in files.

And from time to time, each one of us needs to sync their virtual
environment:

    make sync-deps

More details about pip-tools:

    https://github.com/nvie/pip-tools


bochecha:
I think this not only makes it simpler to manage our dependencies, it also provides a lot of value: with this, we can know for sure that everybody has the exact same versions of every dependency, including the dependencies of our dependencies.

This is really important because currently, depending on when someone installs Ideascube, they might have different Python modules installed, even if they did it with the same `requirements.txt`!

That means we can test our code with an existing virtual environment and everything works fine, but then the Debian package is broken because it was built later, and thus got a newer version of a transitive dependency.

The solution is clearly to list **all** dependencies and their versions in the `requirements.txt` file, including transitive ones. The `pip-tools` allow us to do that in a super simple way.

I don't know if the `pip-tools` are the absolute right answer, but I'd like to move to such a workflow, whether with `pip-tools` or not:

* always list **all** the dependencies (including transitive deps) and their versions in the `requirements.txt` file, so that everyone uses the same thing;
* right after a release, bump it all to the latest and greatest;
* if we didn't bump soon after a release, then don't bump right before a release, wait for the next release;

I'm open to discussion on this, of course. 😄

Makefile
```diff
@@ -3,7 +3,7 @@ test:
 testcov:
 	py.test --cov=ideascube/ --cov-report=term-missing --migrations
 develop:
-	pip3 install -r requirements-dev.txt
+	pip3 install --no-use-wheel -r requirements-dev.txt
```
bochecha: We can't do anything here unfortunately.

Older pip (as on Debian Jessie) uses the `--no-use-wheel` argument.

Newer pip (as on our Fedora workstations, or on Debian Stretch) use `--no-binary :all:`.

We can't move to the new one or else `make develop` won't work on Debian Jessie any more.

Right now I'm just hoping that we'll move to Stretch before pip upstream deletes this flag.

Checking the version of pip in the Makefile to use the right option would be pretty painful to implement (because Makefile targets aren't a shell script) so for now I'd rather just punt on the issue, and do it if it ever becomes a problem.

mgautierfr: That's a pity but we have to go this way :/

PR #724, opened 2017-02-13
Remove Kalite until Arabic language is available

fheslouin:
Modification has been done directly on the server.

Can it merged on the next release  ?


ideascube/conf/idb_irq_bardarash.py
```diff
@@ -14,3 +14,19 @@
 
 
 ENTRY_ACTIVITY_CHOICES = []
+
+STAFF_HOME_CARDS = [c for c in STAFF_HOME_CARDS  # pragma: no flakes
```
bochecha: You don't need the `# pragma: no flakes` comment here.

PR #729, opened 2017-02-15
Csv import check non existant path

mgautierfr:


ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -94,7 +94,7 @@ def add(self, metadata):
             return
 
         path = os.path.join(self.ROOT, original)
-        if not os.path.exists(path):
+        if not os.path.isfile(path):
```
bochecha: This is indeed better than what we had. However, it is also racy.

It would be better to not even check for the file, and just handle opening failures.

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -96,13 +96,21 @@ def add(self, metadata):
         path = os.path.join(self.ROOT, original)
         if not os.path.isfile(path):
             self.report.error(u'Path not found', path)
+            preview = metadata.get('preview')
+            if preview and not os.path.isfile(os.path.join(self.ROOT, preview)):
```
bochecha: Same note about the potential for race-conditions here. BTAFTP

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -93,21 +93,32 @@ def add(self, metadata):
...
+                    try:
+                        with open(preview_path, 'rb') as f:
+                            preview = File(f, name=preview)
+                            self.save(metadata, original, instance, preview)
+                    except IOError:
+                        self.report.error("Cannot open path", preview_path)
+                else:
+                    self.save(metadata, original, instance)
+        except IOError:
+            self.report.error("Cannot open path", original_path)
```
bochecha: Maybe say what's wrong with the path?

It could be a permission issue, a file not found one, ... Just adding the exception to the output would probably be enough.

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -93,21 +93,32 @@ def add(self, metadata):
...
+                    preview_path = os.path.join(self.ROOT, preview)
+                    try:
+                        with open(preview_path, 'rb') as f:
+                            preview = File(f, name=preview)
+                            self.save(metadata, original, instance, preview)
+                    except IOError:
+                        self.report.error("Cannot open path", preview_path)
+                else:
+                    self.save(metadata, original, instance)
+        except IOError:
```
bochecha: I'm pretty sure you want to catch `OSError` here.

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -93,21 +93,32 @@ def add(self, metadata):
...
+                original = File(f, name=original)
+
+                preview = metadata.get('preview')
+                if preview:
+                    preview_path = os.path.join(self.ROOT, preview)
+                    try:
+                        with open(preview_path, 'rb') as f:
+                            preview = File(f, name=preview)
+                            self.save(metadata, original, instance, preview)
+                    except IOError:
```
bochecha: I'm pretty sure you want to catch OSError here.

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -93,21 +93,32 @@ def add(self, metadata):
...
-                    preview = File(f, name=preview)
-                    self.save(metadata, original, instance, preview)
-            else:
-                self.save(metadata, original, instance)
+                preview_path = os.path.join(self.ROOT, preview)
+                try:
+                    with open(preview_path, 'rb') as f:
+                        pass
+                except IOError:
+                    self.report.error("Cannot open path", preview_path)
```
bochecha: If you're checking for the existence of the `preview_path` both in the `try` and in the `except` blocks, then why not just make it its own block?

```python
try:
    with open(original_path, 'rb') as f:
        ...

except OSError as e:
    self.report.error(...)

if preview:
    try:
        with open(preview_path, 'rb') as f:
            ...

    except OSError:
        self.report.error(...)
```

That would avoid some indentation levels, as well as some code duplication.

ideascube/serveradmin/catalog.py
```diff
@@ -394,11 +394,12 @@ def install(self, download_path, install_dir):
...
-            except:
+            except Exception as e:
                 # This can lead to installed package with uninstall media.
                 # We sould handle this somehow.
                 printerr("Cannot install media {} from package {}".format(
                     media['title'], self.id))
+                printerr(e)
                 continue
 
     def _install_media(self, media_info, pseudo_install_dir):
```
bochecha: Is this commit really supposed to be in this pull request?

bochecha: Hahaha, I just realized we both did the same thing: https://github.com/ideascube/ideascube/pull/733/commits/f6cac3da3c4048559b97dfea1ba92c7fa22d7154

😆 

I sort of prefer having the error on the same line as the "Cannot install..." summary, otherwise it's not entirely clear to the user that they are related.

I'm happy to remove it from my pull request since you had done it first.

ideascube/mediacenter/management/commands/import_medias.py
mgautierfr: The main difficulty here is that we need to have both files opened when we are calling the `save()` method.
So we need to have the enclosed `with open(...)`.

Or, we can have a simple `open` and close it in a `finaly` close.

bochecha: > The main difficulty here is that we need to have both files opened when we are calling the save() method.

Argh, right...

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -87,23 +89,36 @@ def add(self, metadata):
...
-                path = os.path.join(self.ROOT, preview)
-                with open(path, 'rb') as f:
-                    preview = File(f, name=preview)
-                    self.save(metadata, original, instance, preview)
-            else:
-                self.save(metadata, original, instance)
+                preview_path = os.path.join(self.ROOT, preview)
+                try:
+                    with open(preview_path, 'rb') as f:
+                        pass
```
bochecha: So, since you're not going to import anything here anyway, you don't need to open the preview path at all.

It's certainly nice to let the user know that the preview path doesn't exist here (otherwise they'd fix the original paths then try again and we'd tell them this time the preview paths are missing).

So how about just `if not os.path.isfile(preview_path)`?

ideascube/mediacenter/management/commands/import_medias.py
```diff
@@ -93,21 +93,38 @@ def add(self, metadata):
...
+                return
+            if preview and not pf:
+                return
+
+            original = File(of, name=original)
+            if pf:
+                preview = File(pf, name=preview)
+                self.save(metadata, original, instance, preview)
             else:
                 self.save(metadata, original, instance)
```
bochecha: Put `preview = None` in the `else` block, and then you can call `self.save(metadata, original, instance, preview=preview)` outside of the `if`/`else`?

bochecha: Whatever, that's not very important, and this branch has some important fixes we need asap, so I'll discount my remark as personal preference and merge this.

Issue #118, opened 2015-03-30
arabic tags have no slug


yohanboniface:
We need a more powerfull slug algo that can do ascii folding on all utf-8 (unidecode may be a good candidate).

yohanboniface:
cf https://github.com/alex/django-taggit/pull/273#issuecomment-66955534

yohanboniface:
cf https://github.com/alex/django-taggit/issues/179

goingeast:
I have issue that first Chinese tag has no slug, then second Chinese tag's slug is "_1".

yohanboniface:
Are you using ideascube, or are you just using django-taggit?

bochecha:
```
$ pew new -p python3 test-slugs
...
Launching subshell in virtual environment. Type 'exit' or 'Ctrl+D' to return.
test-slugs$ pip install django
...
Successfully installed django-1.9.7
test-slugs$ python
Python 3.4.3 (default, Mar 31 2016, 20:42:37) 
[GCC 5.3.1 20151207 (Red Hat 5.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from django.template.defaultfilters import slugify
>>> slugify('林')
''
>>> 
test-slugs$ pip install python-slugify
...
Successfully installed Unidecode-0.4.19 python-slugify-1.2.0
test-slugs$ python
Python 3.4.3 (default, Mar 31 2016, 20:42:37) 
[GCC 5.3.1 20151207 (Red Hat 5.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from slugify import slugify
>>> slugify('林')
'lin'
```

https://pypi.python.org/pypi/python-slugify

I've used this on other projects, and it bugs me to no end that Django provides such a poor `slugify` implementation instead of just taking in python-slugify. :smile:

bochecha:
Ah, and now I read the comments you mentioned above and see this module has already been mentioned. So much for arriving late to the party. :)

PR #730, opened 2017-02-16
Add a command to clean tags.

mgautierfr:
Before ideascube version 0.9, we were using case-sensitive taggit.
Then at this time, taggit were creating two different tags for 'Foo' and
'foo'.

But now, with more recent versions, we are in case-insensitive. Then
taggit assumes there is only one tag for a (insensitive) string.
So if we try to tag a object with tag 'foo', taggit crashes because the
assumption is not fulfill.

This new command "tags sanitize" fixes this by merging (insensitive) equal
tags. After running it, we end with the same situation as if we were all
the time in case-insensitive setting.

This command also clean the tag names. It remove punctuation characters and
remove tags with empty names.

### DIFF
--- a/ideascube/management/commands/tags.py
+++ b/ideascube/management/commands/tags.py
@@ -1,10 +1,13 @@
 import argparse
 import sys
+from itertools import groupby
+from operator import itemgetter
 
 from django.core.management.base import BaseCommand
 from django.utils.termcolors import colorize
 
 from taggit.models import Tag, TaggedItem
+from ideascube.utils import sanitize_tag_name
 
 
 def log(text, **kwargs):
@@ -49,6 +52,11 @@ class Command(BaseCommand):
         replace.add_argument('old', help='Old tag name.')
         replace.add_argument('new', help='New tag name.')
         replace.set_defaults(func=self.replace)
+
+        sanitize = subs.add_parser('sanitize',
+            help=('Sanitize existing tags.\n'
+                  'Remove duplicates, clean characters...'))
+        sanitize.set_defaults(func=self.sanitize)
 
         list_ = subs.add_parser('list', help='List tags')
         list_.set_defaults(func=self.list)
@@ -121,3 +129,57 @@ class Command(BaseCommand):
         print(row.format('.' * 40, '.' * 40, '.' * 40))
         for tag in Tag.objects.order_by('slug'):
             print(row.format(tag.name, tag.slug, self._count(tag.name)))
+
+    def sanitize(self, options):
+        all_tags = ((sanitize_tag_name(t.name), t) for t in Tag.objects.all())
+        all_tags = sorted(all_tags, key=itemgetter(0))
+        all_tags = groupby(all_tags, key=itemgetter(0))
+        for new_tag_name, tags in all_tags:
+            tags = (t[1] for t in tags)
+            if not new_tag_name:
+                # No need to delete relation, happy cascading !
+                for tag in tags:
+                    tag.delete()
+                continue
+
+            tag = next(tags)
+            other_equivalent_tags = list(tags)
+
+            # All the relations we need to redirect to `tag`
+            other_relations = TaggedItem.objects.filter(
+                tag__in=other_equivalent_tags)
+            for relation in other_relations:
+                # if an object `o` is tagged with tag `foo` and `Foo`, the
+                # relation `o-Foo` must be change to `o-foo`. But this relation
+                # already exists, so, instead, we must delete `o-Foo`,
+                # not change it.
+                existing_relations = TaggedItem.objects.filter(
+                        tag=tag,
+                        object_id=relation.content_object.id,
+                        content_type = relation.content_type)
+                if not existing_relations.exists():
+                    # We must change the relation
+                    relation.tag = tag
+                    relation.save()
+                else:
+                    # We have existing relation(s).
+                    # We should not have more than one because we cannot have
+                    # the *exact* same relation twice but let's be safe :
+                    # delete any extra relation.
+                    extra_relations = list(existing_relations)[1:]
+                    for rel in extra_relations:
+                        rel.delete()
+                    # Then delete the current relation because we know we have
+                    # an existing relation.
+                    relation.delete()
+
+            # There is no relation to other tags left, delete them.
+            for t in other_equivalent_tags:
+                t.delete()
+
+            # Be sure our tag is correctly renamed.
+            # We do it at the end because the tag's name is unique and so,
+            # we want to be sure that all potential duplicates have been
+            # deleted/changed.
+            tag.name = new_tag_name
+            tag.save()
--- a/ideascube/mediacenter/models.py
+++ b/ideascube/mediacenter/models.py
@@ -106,7 +106,7 @@ class Document(SearchMixin, TimeStampedModel):
 
     @property
     def index_tags(self):
-        return self.tags.slugs()
+        return list(self.tags.slugs()) + list(self.tags.names())
 
     @property
     def slug(self):
--- a/ideascube/search/tests/test_models.py
+++ b/ideascube/search/tests/test_models.py
@@ -97,6 +97,18 @@ def test_search_Document_multiple_tag():
 
 
 @pytest.mark.usefixtures('cleansearch')
+def test_search_Document_on_tag_name_and_slug():
+    doc1 = DocumentFactory(tags=["aé"])
+    doc2 = DocumentFactory(tags=["ae"])
+
+    assert doc1 in Search.search(tags__match=["ae"])
+    assert doc2 in Search.search(tags__match=["ae"])
+
+    assert doc1 in Search.search(tags__match=["aé"])
+    assert doc2 not in Search.search(tags__match=["aé"])
+
+
+@pytest.mark.usefixtures('cleansearch')
 def test_more_relevant_should_come_first():
     second = ContentFactory(title="About music and music")
     third = ContentFactory(title="About music")
--- a/ideascube/search/tests/test_models.py
+++ b/ideascube/search/tests/test_models.py
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
 import pytest
 
+from operator import attrgetter
+
 from ideascube.blog.tests.factories import ContentFactory
 from ideascube.blog.models import Content
 from ideascube.mediacenter.models import Document
 from ideascube.mediacenter.tests.factories import DocumentFactory
+from ideascube.utils import sanitize_tag_name
 from ..models import Search
 
 
@@ -106,6 +109,23 @@ def test_search_Document_on_tag_name_and_slug():
 
     assert doc1 in Search.search(tags__match=["aé"])
     assert doc2 not in Search.search(tags__match=["aé"])
+
+
+@pytest.mark.usefixtures('cleansearch')
+def test_search_Document_is_case_insensitive():
+    # The custom function to parse tag string (and so sanitize tag's names)
+    # is used is form only, not if we use `tag.set()` or `tag.add()`.
+    # DocumentFactory use `tag.add()` to set the tags of the document and so
+    # the tag's names are not sanitized.
+    doc1 = DocumentFactory(tags=[sanitize_tag_name("aé")])
+    doc2 = DocumentFactory(tags=[sanitize_tag_name("AÉ")])
+    doc3 = DocumentFactory(tags=[sanitize_tag_name("Bar")])
+
+    assert sorted(Search.search(tags__match=["aé"]), key=attrgetter('id')) \
+        == [doc1, doc2]
+    assert sorted(Search.search(tags__match=["AÉ"]), key=attrgetter('id')) \
+        == [doc1, doc2]
+    assert doc3 in Search.search(tags__match=["baR"])
 
 
 @pytest.mark.usefixtures('cleansearch')
--- a/ideascube/tests/test_tags_command.py
+++ b/ideascube/tests/test_tags_command.py
@@ -79,3 +79,36 @@ def test_list_should_list_tags_and_slugs(capsys):
     out, err = capsys.readouterr()
     assert tag.name in out
     assert tag.slug in out
+
+
+def test_sanitize_tags():
+    foo = Tag.objects.create(name='foo')
+    Foo = Tag.objects.create(name='Foo')
+    Bar = Tag.objects.create(name='Bar') # Create a tag with upper case first.
+    bar = Tag.objects.create(name='bar')
+    bar_ = Tag.objects.create(name='bar;')
+    Bar_ = Tag.objects.create(name='Bar;')
+    tag_to_delete = Tag.objects.create(name=':')
+    clean = Tag.objects.create(name="Other:")
+    half_clean1 = Tag.objects.create(name="Other:Foo,")
+    half_clean2 = Tag.objects.create(name="Other:foo")
+
+    doc1 = DocumentFactory(tags=[foo, bar, clean])
+    doc2 = DocumentFactory(tags=[foo, Bar, clean, half_clean2])
+    doc3 = DocumentFactory(tags=[Foo, bar])
+    doc4 = DocumentFactory(tags=[Foo, Bar_, half_clean1, half_clean2])
+    doc5 = DocumentFactory(tags=[Foo, foo, bar_, Bar])
+    doc6 = DocumentFactory(tags=[Foo, foo, Bar_, tag_to_delete])
+
+    call_command('tags', 'sanitize')
+
+    all_tag_names = list(Tag.objects.all().order_by('name')
+                             .values_list('name', flat=True))
+    assert all_tag_names == ['bar', 'foo', 'other', 'other:foo']
+    assert sorted(doc1.tags.names()) == ['bar', 'foo', 'other']
+    assert sorted(doc2.tags.names()) == ['bar', 'foo', 'other', 'other:foo']
+    assert sorted(doc3.tags.names()) == ['bar', 'foo']
+    assert sorted(doc4.tags.names()) == ['bar', 'foo', 'other:foo']
+    assert sorted(doc5.tags.names()) == ['bar', 'foo']
+    assert sorted(doc6.tags.names()) == ['bar', 'foo']
+
--- a/ideascube/utils.py
+++ b/ideascube/utils.py
@@ -89,6 +89,11 @@ def get_file_sha256(path):
 def get_file_size(path):
     return os.stat(path).st_size
 
+def sanitize_tag_name(tag_name):
+    tag_name = tag_name.strip(';:.,?!+-@+-/* \t')
+    tag_name = tag_name.lower()
+    return tag_name
+
 def tag_splitter(tag_string):
     tags = set(t.strip() for t in re.split(r'[;,]+', tag_string) if t.strip())
     return list(tags)
--- a/ideascube/utils.py
+++ b/ideascube/utils.py
@@ -95,5 +95,5 @@ def sanitize_tag_name(tag_name):
     return tag_name
 
 def tag_splitter(tag_string):
-    tags = set(t.strip() for t in re.split(r'[;,]+', tag_string) if t.strip())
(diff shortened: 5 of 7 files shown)


ideascube/management/commands/tags.py
```diff
@@ -20,6 +21,13 @@ def exit(text, **kwargs):
     sys.exit(1)
 
 
+ponctuation_regex = re.compile(r"([;:.,?!+-@+-/*]+)")
```
bochecha: I'm not sure.

`Culture: musique` seems like a very valid tag to me. So does `audio-visuel`.

We might want to keep some punctuation marks?

ideascube/management/commands/tags.py
```diff
@@ -20,6 +21,13 @@ def exit(text, **kwargs):
...
+def sanitize_tag_name(tag_name):
+    tag_name = ponctuation_regex.sub('', tag_name)
+    tag_name = tag_name.strip()
```
bochecha: At this point, what if the tag name is empty?

ideascube/management/commands/tags.py
```diff
@@ -50,6 +58,13 @@ def add_arguments(self, parser):
         replace.add_argument('new', help='New tag name.')
         replace.set_defaults(func=self.replace)
 
+
+        sanitize = subs.add_parser('sanitize',
+            help=('Sanitize existing tags.\n'
+                  'Remove duplicates, clean characters...'))
+        sanitize.set_defaults(func=self.sanitize)
+
+
```
bochecha: Too many blank lines for pep8?

ideascube/management/commands/tags.py
```diff
@@ -121,3 +136,44 @@ def list(self, options):
         print(row.format('.' * 40, '.' * 40, '.' * 40))
         for tag in Tag.objects.order_by('slug'):
             print(row.format(tag.name, tag.slug, self._count(tag.name)))
+
+    def sanitize(self, options):
+        for relation in TaggedItem.objects.all():
+            tag = relation.tag
+            new_tag_name = sanitize_tag_name(tag.name)
+            if not new_tag_name:
```
bochecha: Ah, that's where you handle empty tags. Ok then.

ideascube/management/commands/tags.py
```diff
@@ -121,3 +136,44 @@ def list(self, options):
...
+            if not tags.count():
+                # There is none.
+                # This can happen if we change the tag name.
+                # Then, we just want to rename the tag.
+                tag.name = new_tag_name
+                tag.save()
+            else:
+                new_tag = tags[0]
+                if tag == new_tag:
+                    continue
```
bochecha: I can't understand what your intention is with these 3 lines. 😕

ideascube/management/commands/tags.py
```diff
@@ -121,3 +136,44 @@ def list(self, options):
...
+
+                # Is the object already related to the new tag ?
+                already_existing_relations = TaggedItem.objects.filter(
+                    tag=new_tag,
+                    object_id=relation.content_object.id,
+                    content_type = relation.content_type)
+                if already_existing_relations.count():
```
bochecha: Remember, if you just want a boolean, then `.exists()` is supposed to be faster (in addition to being semantically more correct)

ideascube/management/commands/tags.py
mgautierfr: Yes, we could remove some.

ideascube/management/commands/tags.py
mgautierfr: `new_tag` is the tag we want to keep. 
But `tag` may be already the tag we want to keep (no change in the name, the only one (or the first created) with this name). 
In this case everything is ok, just continue to the next relation.

ideascube/management/commands/tags.py
bochecha: Not just remove them from the regex.

We do want to weed out tags like `;` after all.

I think a regex is not going to cut it.

ideascube/management/commands/tags.py
bochecha: Right, I understand what those lines do. What I don't get is why they are here in this form.

They look awfully specific, but I just can't make sense out of them.

For example, what about the other items of the `tags` list? Why ignore them?

mgautierfr: > For example, what about the other items of the tags list? Why ignore them?

Because this especially what we want : remove the case-insensitive duplicated tags.
 We remove them by always taking the first created one. All the others should be ignored and the `relation` should be "redirected" to the right one.  At the end, the tags to remove will have no relation linking to them. In the second loop (on the `Tags` themselves) we delete them.

ideascube/management/commands/tags.py
```diff
@@ -50,6 +57,13 @@ def add_arguments(self, parser):
         replace.add_argument('new', help='New tag name.')
         replace.set_defaults(func=self.replace)
 
+
```
bochecha: Too many blank lines for pep8?

ideascube/management/commands/tags.py
bochecha: This is still an issue.

ideascube/tests/test_tags_command.py
```diff
@@ -79,3 +79,56 @@ def test_list_should_list_tags_and_slugs(capsys):
...
+    doc2 = DocumentFactory(tags=[foo, Bar, clean, half_clean2])
+    doc3 = DocumentFactory(tags=[Foo, bar])
+    doc4 = DocumentFactory(tags=[Foo, Bar_, half_clean1, half_clean2])
+    doc5 = DocumentFactory(tags=[Foo, foo, bar_, Bar])
+    doc6 = DocumentFactory(tags=[Foo, foo, Bar_, tag_to_delete])
+
+    call_command('tags', 'sanitize')
+
+    for name in ('Foo', 'Bar', 'Bar;', 'bar;', ';', 'Other:', 'Other:Foo,', 'Other:foo'):
+        assert not Tag.objects.filter(name=name)
```
bochecha: `.exists()`?

ideascube/tests/test_tags_command.py
```diff
@@ -79,3 +79,56 @@ def test_list_should_list_tags_and_slugs(capsys):
...
+
+    for doc in (doc1, doc2):
+        assert clean in doc.tags.all()
+    for doc in (doc3, doc4, doc5, doc6):
+        assert clean not in doc.tags.all()
+
+    for doc in (doc2, doc4):
+        assert half_clean1 in doc.tags.all()
+    for doc in (doc1, doc3, doc5, doc6):
+        assert half_clean1 not in doc.tags.all()
```
bochecha: I think I'd have preferred something simpler, like:

```python
assert [t.name for t in doc1.tags.order_by('name')] == ['foo', 'bar', 'other']
assert [t.name for t in doc2.tags.order_by('name')] == ['foo', 'bar', 'other', 'other:foo']
...
```

In addition to being much simpler (and explicit), it removes a doubt: as you noted in a commend below, Django keeps the Python objects cached and they need to be refreshed from the db; which could lead one to wonder what gets tested by the `assert some_tag not in doc.tags.all()`.

It turns out that the tests are correct here (good job), but it's still a doubt that completely disappears, nobody will ever wonder about it when reading the code I suggested.

ideascube/tests/test_tags_command.py
```diff
@@ -79,3 +79,56 @@ def test_list_should_list_tags_and_slugs(capsys):
...
+    doc4 = DocumentFactory(tags=[Foo, Bar_, half_clean1, half_clean2])
+    doc5 = DocumentFactory(tags=[Foo, foo, bar_, Bar])
+    doc6 = DocumentFactory(tags=[Foo, foo, Bar_, tag_to_delete])
+
+    call_command('tags', 'sanitize')
+
+    for name in ('Foo', 'Bar', 'Bar;', 'bar;', ';', 'Other:', 'Other:Foo,', 'Other:foo'):
+        assert not Tag.objects.filter(name=name)
+    for name in ('foo', 'bar', 'other', 'other:foo'):
+        assert Tag.objects.get(name=name)
```
bochecha: To be honest here I'd replace lines 105-108 by:

```python
assert [t.name for t in Tag.objects.order_by('name')] == ['bar', 'foo', 'other', 'other:foo']
```

(or something with `values_list` but I always forget how to use it)

It ensures the same thing (only these 4 tag names are left), but it probably does it faster (only one query), and it's just much simpler and more explicit to read (no for loops involved).

ideascube/tests/test_tags_command.py
```diff
@@ -79,3 +79,56 @@ def test_list_should_list_tags_and_slugs(capsys):
...
+        assert half_clean1 in doc.tags.all()
+    for doc in (doc1, doc3, doc5, doc6):
+        assert half_clean1 not in doc.tags.all()
+
+    # We have to reload the tag from the db as django cache the attribute in
+    # the python object.
+    assert Tag.objects.get(id=foo.id).name == 'foo'
+    assert Tag.objects.get(id=Bar.id).name == 'bar'
+    assert Tag.objects.get(id=clean.id).name == 'other'
+    assert Tag.objects.get(id=half_clean1.id).name == 'other:foo'
```
bochecha: I'd just remove these lines.

What we actually care about is that only the right tags are left, and document-tag relationships were removed/fixed.

The fact that this happened by renaming some tags and keeping their `id` seems entirely irrelevant to me, it's just an implementation detail which is highly dependent on the order in which you created the tags above and the `order_by` used in the `sanitize` command.

ideascube/management/commands/tags.py
```diff
@@ -121,3 +135,53 @@ def list(self, options):
         print(row.format('.' * 40, '.' * 40, '.' * 40))
         for tag in Tag.objects.order_by('slug'):
             print(row.format(tag.name, tag.slug, self._count(tag.name)))
+
+    def sanitize(self, options):
+        all_tags = list(Tag.objects.all().order_by('id').reverse())
```
bochecha: No need to `.all()` if you follow it with `.order_by('id')`

bochecha: Why do you need to order the query by `id`?

ideascube/management/commands/tags.py
```diff
@@ -121,3 +135,53 @@ def list(self, options):
...
+            tag = all_tags.pop()
+            new_tag_name = sanitize_tag_name(tag.name)
+
+            if not new_tag_name:
+                # No need to delete relation, happy cascading !
+                tag.delete()
+                continue
+
+            other_equivalent_tags = {t for t in all_tags
+                if new_tag_name == sanitize_tag_name(t.name)}
```
bochecha: You're going to be doing that a lot (although `all_tags` is going to get smaller every time).

Why not do, before the `while` loop:

```python
        all_tags = ((sanitize_tag_name(t.name), t) for t in Tag.objects.all())
        all_tags = sorted(all_tags, key=itemgetter(0))
        all_tags = groupby(all_tags, key=itemgetter(0))
```

Then you can just iterate over `all_tags`:

```python
        for sanitized_name, tags in all_tags:
            if not sanitized_name:
                # Delete the tags
                continue

            # Keep the first tag as the one and only
            tags = list(tags)  # Or `tags = sorted(tags, key=attrgetter('id'))` if you really want to keep the first created one, but I don't think it matters at all
            tag_to_keep = tags[0]
            other_equivalent_tags = tags[1:]
            
            # Then do what you have to do with the `other_relations`, drop the `other_equivalent_tags` and rename the tag
```

Wouldn't that be simpler?

ideascube/management/commands/tags.py
```diff
@@ -121,3 +135,53 @@ def list(self, options):
...
+                        object_id=relation.content_object.id,
+                        content_type = relation.content_type)
+                except ObjectDoesNotExist:
+                    # We must change the relation
+                    relation.tag = tag
+                    relation.save()
+                except MultipleObjectsReturned:
+                    # This should not happend, we cannot have the *exact* same
+                    # relation twice. But let's be safe.
+                    relation.delete()
```
bochecha: In addition to deleting the now-undesirable relation (because of tag sanitation), you should probably also delete all the additional-relations `o-foo` except one, so that this `MultipleObjectsReturned` never happens again?

ideascube/management/commands/tags.py
```diff
@@ -121,3 +135,53 @@ def list(self, options):
...
+                else:
+                    # We must delete the relation
+                    relation.delete()
+
+            # There is no relation to other tags, delete them.
```
bochecha: This gave me a pause: I wondered why there would be no relation to the other tags, until I realized that there is no relation to other tags **left** at this point (because we deleted or moved them just above).

What a difference a word can make. 😄

ideascube/utils.py
```diff
@@ -95,5 +95,6 @@ def sanitize_tag_name(tag_name):
     return tag_name
 
 def tag_splitter(tag_string):
-    tags = set(t.strip() for t in re.split(r'[;,]+', tag_string) if t.strip())
+    tags = set(sanitize_tag_name(t) for t in re.split(r'[;,]+', tag_string))
+    tags = filter(None, tags)
```
bochecha: Aren't you the one who convinced me to stop using `map`, `filter` and their friends? 😛

ideascube/mediacenter/models.py
```diff
@@ -106,7 +106,7 @@ def index_source(self):
 
     @property
     def index_tags(self):
-        return self.tags.slugs()
+        return self.tags.names()
```
bochecha: Why not search on both?

We now lower-case all tags, which means the `slug` will now be the `name` with a few differences:

* punctuation and spaces replaced by `-`
* non-ascii characters replaced by their `unidecode`-ed counterpart (modulo the fact that some languages are poorly handled by our slugifying process, see #118)

As a result, if we index/search on both the `name` and the `slug`, then a search for the term "marche" will return both tags containing "marche" and "marché".

Oh look! That feature we wanted! For free! 😃

ideascube/management/commands/tags.py
mgautierfr: Hum, it was when we wanted to keep the case of the first created tag.
But this is not relevant now.

ideascube/management/commands/tags.py
mgautierfr: Whouaaa, this is so much simpler this way. Thanks.

ideascube/management/commands/tags.py
mgautierfr: Hum, yes. But we should all *but* one.

ideascube/utils.py
mgautierfr: Huhu. changed :)

ideascube/management/commands/tags.py
```diff
@@ -121,3 +129,57 @@ def list(self, options):
         print(row.format('.' * 40, '.' * 40, '.' * 40))
         for tag in Tag.objects.order_by('slug'):
             print(row.format(tag.name, tag.slug, self._count(tag.name)))
+
+    def sanitize(self, options):
+        all_tags = ((sanitize_tag_name(t.name), t) for t in Tag.objects.all())
+        all_tags = sorted(all_tags, key=itemgetter(0))
+        all_tags = groupby(all_tags, key=itemgetter(0))
+        for new_tag_name, tags in all_tags:
+            tags = (t[1] for t in tags)
```
bochecha: Oh right, I forgot about that!

So you end up twice with the sanitized tag name in the iterator, one that was used by sorted/groupby, and one which is the result of the groupby. 😞

ideascube/tests/test_tags_command.py
bochecha: You didn't address this comment.

(it's fine if you disagree with me, but just say it in that case 😉)

ideascube/tests/test_tags_command.py
```diff
@@ -79,3 +79,38 @@ def test_list_should_list_tags_and_slugs(capsys):
...
+    doc6 = DocumentFactory(tags=[Foo, foo, Bar_, tag_to_delete])
+
+    call_command('tags', 'sanitize')
+
+    for name in ('Foo', 'Bar', 'Bar;', 'bar;', ';', 'Other:', 'Other:Foo,', 'Other:foo'):
+        assert not Tag.objects.filter(name=name).exists()
+    for name in ('foo', 'bar', 'other', 'other:foo'):
+        assert Tag.objects.get(name=name)
+
+    assert sorted(doc1.tags.names()) == ['bar', 'foo', 'other']
```
bochecha: Oh nice! I had no idea Taggit provided a `.names()` manager method. 😃

bochecha: Of course now I realize that you were using it in the `index_tags` method and I feel like I didn't pay sufficient attention. :shame:

ideascube/search/tests/test_models.py
```diff
@@ -97,6 +97,18 @@ def test_search_Document_multiple_tag():
 
 
 @pytest.mark.usefixtures('cleansearch')
+def test_search_Document_on_tag_name_and_slug():
+    doc1 = DocumentFactory(tags=["aé"])
+    doc2 = DocumentFactory(tags=["ae"])
+
+    assert doc1 in Search.search(tags__match=["ae"])
+    assert doc2 in Search.search(tags__match=["ae"])
```
bochecha: See, I had no idea this would actually work. 😆 

I'm glad I suggested it though, and you proved it did. Thanks!

ideascube/search/tests/test_models.py
```diff
@@ -109,6 +110,23 @@ def test_search_Document_on_tag_name_and_slug():
...
+    # is used is form only, not if we use `tag.set()` or `tag.add()`.
+    # DocumentFactory use `tag.add()` to set the tags of the document and so
+    # the tag's names are not sanitized.
+    doc1 = DocumentFactory(tags=[sanitize_tag_name("aé")])
+    doc2 = DocumentFactory(tags=[sanitize_tag_name("AÉ")])
+    doc3 = DocumentFactory(tags=[sanitize_tag_name("Bar")])
+
+    for d in (doc1, doc2):
+        assert d in Search.search(tags__match=["aé"])
+        assert d in Search.search(tags__match=["AÉ"])
```
bochecha: You're essentially running the search query for each document.

Why not run it once, and then ensure it is **exactly** the list of doc1 and doc2?

That advantageously replaces the above 3 lines, but it also removes the need for the next one.

ideascube/tests/test_tags_command.py
mgautierfr: Hum, I don't understand your comment here.
Have you tried to comment something else ?

bochecha: Your 4 lines (105-108) essentially ensure two things:

* that the "invalid" tag names don't exist any more
* that the "correct" tag names exist

You'd get the exact same thing by doing:

```python
assert sorted(Tag.objects.names()) == ['bar', 'foo', 'other', 'other:foo']
```

(yes, now that I know of the `.names()` manager method, I use it 😛 )

It's less code, it does 1 query instead of 12, and in addition to the above two things it ensures one more: there are no other tags. (e.g one with an empty `name`, because somehow the `;` one ended up sanitized but not removed)

Less code which does more and faster, I don't see a reason not to do it. :)

ideascube/search/tests/test_models.py
```diff
@@ -109,6 +112,24 @@ def test_search_Document_on_tag_name_and_slug():
...
+    # the tag's names are not sanitized.
+    doc1 = DocumentFactory(tags=[sanitize_tag_name("aé")])
+    doc2 = DocumentFactory(tags=[sanitize_tag_name("AÉ")])
+    doc3 = DocumentFactory(tags=[sanitize_tag_name("Bar")])
+
+    assert sorted(Search.search(tags__match=["aé"]), key=attrgetter('id')) \
+        == [doc1, doc2]
+    assert sorted(Search.search(tags__match=["AÉ"]), key=attrgetter('id')) \
+        == [doc1, doc2]
+    assert doc3 not in Search.search(tags__match=["aé"])
```
bochecha: That's already `assert`-ed by line 124.

mgautierfr: Yes. Removed in last git push -f

PR #732, opened 2017-02-17
Remove the csv2pkg command

bochecha:
We needed it outside of Ideascube, so that it is easier to install
without installing all of Ideascube.

It has been extracted to its own repository:

    https://framagit.org/ideascube/csv-to-pkg

---

Please review not only this pull request (to ensure I didn't remove anything important, or that I didn't forget to remove anything) as well as the new csv-to-pkg project. (I had to modify the code to make it work standalone)


mgautierfr:
This technically looks good.

However, on the long term, we have (had ?) the plan to create a package for the user content and install it with the catalog. So, I wonder if we should remove the `MediaCenterPackage` from ideascube.
But in the same time, if we keep it, we will have dead code because nothing but csv2pkg use `MediaCenterPackage`.

I have no real idea on what is better. So I will not block this PR for this. But it worth to mention this point, at least for the record.

bochecha:
> However, on the long term, we have (had ?) the plan to create a package for the user content and install it with the catalog. So, I wonder if we should remove the MediaCenterPackage from ideascube.
But in the same time, if we keep it, we will have dead code because nothing but csv2pkg use MediaCenterPackage.
> 
> I have no real idea on what is better. So I will not block this PR for this. But it worth to mention this point, at least for the record.

Good points, and I'm glad you made them because I had completely forgotten about that.

In any case, there are parts of this `MediaCenterPackage` class which wouldn't be needed in the future (`from_csv` comes to mind).

In addition to it being dead code right now, it is possible that after using the external `csv2pkg` more and more, we realize that changes are needed to this class. I'd rather bring it back wholesale in Ideascube when we end up needing it, than making those changes both here and in the external project.

mgautierfr:
You're right, I will merge it right now. If we need the `MediaCenterPackage` later, we can copy it back from `csv-to-pkg` repository.

Issue #675, opened 2017-01-06
What are those .old cards?


bochecha:
I see quite a few setting files with card ids ending in `.old`, for example `ideascube/conf/idb_fra_cultura.py`:

```python
HOME_CARDS = HOME_CARDS + [
    {
        'id': 'cpassorcier.old',
    },
    {
        'id': 'wikisource.old',
    },
    {
        'id': 'ted.old',
    },
    {
        'id': 'ubuntudoc.old',
    },
]
```

What are these?

They seem to have been introduced in this commit:

```
commit 94003e8b3c75231d0226dd88fc94692b7b0a9445
Author: Yohan Boniface <yb@enix.org>
Date:   Thu Mar 10 19:37:13 2016 +0100

    Package manager aware static cards
```

The message is not very explicit, so I'm not sure what to understand.

There are even some of those `.old` cards in `ideascube/templates/ideascube/includes/cards/`.

Is that old stuff that became obsolete with the move to the package manager for content? Should we remove all of this?

fheslouin:
These `.old` cards are used for zim files that are not managed by the catalog. 

These settings have to be keep until ansiblecube is going to be able to move zim files from old path to the new path managed by the catalog. It should happen soon !

bochecha:
Thanks for the explanation. 👍 

Let's keep this open until that time then, so we don't forget to delete them in the glorious future. (who said "next August"?!)

fheslouin:
AnsibleCube use `/etc/default/ideascube` and the `IDEASCUBE_ID=` variable
If a file in `ideascube/conf/` is found matching the device name, this file will be use and possibly some old cards `.old`

We have to take care of this and remove those `.old` lines and possibly `.py` files as zim migration are going

PR #735, opened 2017-02-20
Removing custom conf. Not needed anymore

fheslouin:
The server has been migrated to the last version of ideascube. 
this specific conf is not needed anymore.


ideascube/conf/idb_fra_aurore.py
```diff
@@ -1,28 +0,0 @@
-"""Ideasbox Aurore in Paris Clichy, France"""
-from .idb_fra import *  # pragma: no flakes
-
-
-IDEASCUBE_NAME = u"Aurore"
```
bochecha: If we do this, then the server will be called « Ideas Cube » in the UI (top-left corner of the app) instead of « Aurore ».

Is that what we want?

If not, you must run `ideascube config set server site-name Aurore` on the server, to get the name set properly.

bochecha:
This relates to #675 .

ideascube/conf/idb_fra_aurore.py
fheslouin: This can be removed as the user can now rename the server himself from UI

bochecha: > This can be removed as the user can now rename the server himself from UI

I know, I implemented that feature. :)

My point is that **right now**, they see "Aurore" in the top-right corner, and if we merely remove the configuration file they will suddenly see "Ideas Cube".

I think that's needlessly hostile towards our users, and I'm suggesting we run the command (maybe through Ansiblecube?) to set the name to what it currently is so that users don't see a difference after the upgrade.

The fact that it's configurable in the UI doesn't mean they will go and configure it, and we should try and limit disruptions whenever possible. This is a trivial one to avoid.

Did you already add the command in Ansiblecube and we can merge this without regrets? Or am I over-thinking this and we should not care?

bochecha: @fheslouin just told me on IRC that he's doing it on the server. Merging.

PR #736, opened 2017-02-22
No Kalite needed, so specific config file

fheslouin:


ideascube/conf/kb_fra_crfaix.py
```diff
@@ -0,0 +1,13 @@
+"""KoomBook conf"""
+from .kb import *  # pragma: no flakes
+
+LANGUAGE_CODE = 'fr'
+IDEASCUBE_NAME = 'CRF AIX'
```
bochecha: The commit message only says you want to remove Kalite from the home, but then you define these two above.

Are they necessary?

fheslouin: They are not really necessary actually, but because I can specify variable I did it...

bochecha: > They are not really necessary actually, but because I can specify variable I did it...

I'd rather you didn't, then.

Eventually we want to move to a point where we have almost no site-specific configuration file in Ideascube, and the first step is to stop adding new ones unless necessary.

fheslouin: Ok, I removed them

PR #737, opened 2017-02-23
Mismatch type in mediacenter : .mobi are considered as text files 

barbayellow:
Mobi files are considered as text files in the media center. 

image

The form to uplaod files does not allow either to add a mobi file
image

mobi is the Amazon proprietary ebook format. It is used on the prorietary Kindle device. Kindles can only read .mobi and .pdf.  As a user  looking for book to read on my kindle, I want to have only mobi files and noot epub files. This is why we need to add this new format.


bochecha:
I'm not sure that's enough.

Your change ensures there is a `mobi` option in the form. However, it does nothing to properly display the mobi files?

bochecha:
I mentioned this on IRC, but let's move the discussion here so it's archived.

I'm wondering why you want to add mobi files to the mediacenter.

Unless I'm missing something, those are ebooks, and as such would be better in the library?

For example, I just created this book on my local development instance, and added two electronic specimens to it: one epub and one mobi:

ideas_cube_-_2017-03-16_16 56 25

You can clearly see the two download links, one for the epub and one for the mobi (and we probably should order these alphabetically).

Doesn't that fit the use-case of uploading .mobi books to Ideascube?

That doesn't solve the second part of your use-case, though: listing books which have a .mobi specimen. But here as well, we should probably add this feature to the library, not the mediacenter?

barbayellow:
The pb is : sometimes we need to push remotely e-books on the kombooks or ideasbox.  And the only way we can do this is to use the catalog. And the catalog installs the books in the media center. So as  long as we don't have a possibility to install remotely book packages in the library, we need a mobi format in ideasbox.

A bigger question could be : do we want to have ebook files living in 2 different places, in the media center and in the library app ? I think this is not a good idea because we sometimes need to create packages with video, ebooks, images, and this kind of package can only live in the media center. And this is confusing.

bochecha:
> The pb is : sometimes we need to push remotely e-books on the kombooks or ideasbox. And the only way we can do this is to use the catalog. And the catalog installs the books in the media center. So as long as we don't have a possibility to install remotely book packages in the library, we need a mobi format in ideasbox.

Right, I understand the problem better now, thanks.

> A bigger question could be : do we want to have ebook files living in 2 different places, in the media center and in the library app ? I think this is not a good idea because we sometimes need to create packages with video, ebooks, images, and this kind of package can only live in the media center. And this is confusing.

Yeah, there's definitely something to think about, here. 😕 

I also think it's confusing having content in multiple places. People will never know where to find a book they want to read: is it in the library? is it in the mediacenter?

One way to get out of this is if we made the mediacenter into the canonical store of all digital content in Ideascube. That is, digital book specimens (epubs, pdf, mobi, audio books, ...) would all be stored in the mediacenter, along with all other content.

Digital book specimens would thus not be **stored** in the library any more, but they would still be **accessible** from the library.

In essence, the library would become a front dedicated to part of the content from the mediacenter: digital books.

That would need some serious rearchitecturing of Ideascube though, and some pretty hairy data migration.

It would also need some changes to the `zipped-medias` packages, so that media files corresponding to digital books would have some more metadata, so that we create the book/specimen when installing those packages.

But then the benefits would be pretty interesting: all digital content would always be available in the mediacenter, and all digital books would **also** be available in the library.

Users wouldn't need to know where to find which book, they would be able to find them wherever they prefer looking:

* in the mediacenter, if they are searching for medias about a certain topic, and find an interesting related book;
* in the library, if they were explicitly searching for a book

How does that sound? Any other idea to get us out of this conundrum?

(note that I'm talking about a long term solution here, if we were to embark in this kind of big project to improve the long term UX of Ideascube, then it goes without saying that adding the .mobi format to the mediacenter as a first step becomes more appealing)

barbayellow:
That sounds interesting. However, this is a long run shot. And we got a lot in our plate. We could take an intermediary step that would be to store every digital content in the mediacenter and not adding any books in the library app.

bochecha:
> We could take an intermediary step that would be to store every digital content in the mediacenter and not adding any books in the library app.

But... then people won't find books in the library? 😕

barbayellow:
no they won't. But they'll find books in the media center.  At the moment I'm not even sure anyone knows that library app can host both physical and digital books.

bochecha:
> At the moment I'm not even sure anyone knows that library app can host both physical and digital books.

Ouch. >_<

If indeed that's the case, then it makes the idea of going towards storing everything in the mediacenter more palatable.

This definitely deserves more thought in any case, I'll open a dedicated issue, so that we can get this pull request back to its intended purpose, since that's going to be needed anyway.

bochecha:
I rebased this pull request, and added a commit so that the mobi files are properly displayed in the mediacenter list view.

This now seems to work the same as the epub files.

@mgautierfr can you review?

mgautierfr:
Long and interesting discussion here :) We really have to think about this in the future.

The code itself looks good. I'm merging.

PR #738, opened 2017-02-23
Some minor test improvements

bochecha:
This isn't much, just some stuff I found while working on #691

### DIFF
--- a/ideascube/tests/test_settings.py
+++ b/ideascube/tests/test_settings.py
@@ -6,13 +6,11 @@ import pytest
 def test_all_settings(request):
     confdir = request.fspath.dirpath().dirpath().join('conf')
     assert confdir.check(dir=True)
-    count = 0
 
     for path in confdir.listdir():
         if path.ext != '.py':
             continue
 
-        count += 1
         module = path.purebasename
         module = '.conf.%s' % module
 
@@ -25,6 +23,3 @@ def test_all_settings(request):
             pytest.fail(str(e))
 
         assert isinstance(getattr(settings, 'IDEASCUBE_NAME', ''), str)
-
-    # Make sure we have actually tested some files.
-    assert count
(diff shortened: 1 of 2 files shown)


ideascube/tests/test_settings.py
```diff
@@ -14,12 +14,6 @@ def test_all_settings(request):
...
         module = '.conf.%s' % module
 
-        try:
-            settings = importlib.import_module(module, package="ideascube")
-
-        except Exception as e:
-            print(path)
-            print(module)
-            pytest.fail(str(e))
+        settings = importlib.import_module(setting_module, package="ideascube")
```
mgautierfr: Why not try/catch the exception anymore ?

There is no `setting_module` here. On latter commit yes, but not yet.

bochecha: > Why not try/catch the exception anymore ?

Because if there's an exception, `pytest` considers the test to have failed and displays the exception, which is exactly what we want.

If we catch it and tell `pytest.fail(e)` as we were doing, then we get an output covering several pages of terminal, with lots of unnecessary boilerplate, making finding the actual error much harder.

> There is no setting_module here. On latter commit yes, but not yet.

The joys of rebasing. 😆

ideascube/tests/test_settings.py
```diff
@@ -1,19 +1,20 @@
...
 import importlib
 
 import pytest
 
 
-def test_all_settings(request):
-    confdir = request.fspath.dirpath().dirpath().join('conf')
-    assert confdir.check(dir=True)
+@pytest.fixture(params=glob.glob('ideascube/conf/*.py'))
+def setting_module(request):
```
mgautierfr: I like this. This is far better with a fixture.

ideascube/tests/test_settings.py
bochecha: And fixed.

ideascube/tests/test_settings.py
bochecha: Me too!

PR #740, opened 2017-02-24
Remove -e in requirements(-dev).txt

mgautierfr:
pip-compile doesn't support URLs as packages unless they are editable.
So we need the '-e' options in *.in files.

But, then when dh-virtualenv	install the package as editable with a
egg-link pointing to the source. As the sources change location when we
package everything, the egg-link points to nothing and we cannot import
the packages.

This is a bit silly, but the sed after the requirements.txt generation
removes the '-e' option and everything work again.
This is definitively a workaround and we should remove it once this bug is
fixed : https://github.com/spotify/dh-virtualenv/issues/200

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,10 @@ migrate:
 upgrade-deps:
 	pip-compile --rebuild --header --index --annotate  requirements.in
 	pip-compile --rebuild --header --index --annotate  requirements-dev.in
+	# Remove -e in the requirements(-dev).txt.
+	# See issue : https://github.com/spotify/dh-virtualenv/issues/200
+	sed -i 's/^-e //g' requirements.txt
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,6 +15,8 @@ migrate:
 upgrade-deps:
 	pip-compile --rebuild --header --index --annotate  requirements.in
 	pip-compile --rebuild --header --index --annotate  requirements-dev.in
+	sed -i 's/^-e //g' requirements.txt
```
bochecha: I'd like a comment referencing the issue above the 2 sed lines. It's fine in the commit message, but in such a case, more is better.

mgautierfr: Done

PR #743, opened 2017-03-02
Dependencies

bochecha:


bochecha:
> It's a pity we have to handle all this stuff with this kind of hack and workaround

Agreed.

I do think the benefit is worth it though: reproducible build/test/install environments, identical on our development workstations, in the CI, or in the Debian packages.

But yeah, it's all super fiddly. I guess that's the state of Python packaging/distribution, though. 😞 

Some of these hacks will go away at some point: if we can get dh-virtualenv fixed then we can remove the `sed` line and just use `-e` happily.

Eventually the PyPA folks will save us all with the `pyproject.toml` file, but that might take a while still.

PR #745, opened 2017-03-02
Add missing empty dir in git.

mgautierfr:
The commit a94cd6e9 add a test using a empty directory.
This empty directory should have been added to git but I forgot it.

As the test checks that the command inded fails, it passed. But it was
not testing the right thing.


bochecha:
I'm not sure I really like this...

To be honest, I'd much prefer it if the tests didn't all use the same `CSV_PATH`.

Because of it, we must put all the test data in the git tree:

```
$ ls -1 ideascube/mediacenter/tests/data/
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAan-image.jpg
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA.mp4
an-image.jpg
a-pdf.pdf
a-video.mp4
```

But the import tests don't really need to care about whether the file really contains an image or video, so we could just go with text files.

And those test files could be written at the same time as the `metadata.csv` file in a `tmpdir` for each test.

This would remove the need for an empty directory in the Git tree (the directory would be created by the test function in its `tmpdir`), and we could also remove the data files. (unless they are used by other tests which actually do need the right file content/type)

What do you think?

mgautierfr:
I globally agree with you. Tests have to be rewritten to avoid all those data in the git repository.

But for now, we have a test checking that we fail nicely if we have a directory instead of a file but it tests that we fail nicely if we have nothing instead of a file.

Those tests are not perfect at all, but this commit is really a quick fix for a for a forgotten directory in the previous commit.

I suggest that we merge this PR and open an issue to rewrite those tests later.

bochecha:
> this commit is really a quick fix for a for a forgotten directory in the previous commit.

I'm not sure I would have merged the previous commit if it came with an empty directory like this. 😄 

> I suggest that we merge this PR and open an issue to rewrite those tests later.

Is this really that urgent that we need to merge this now? Can't we just directly do the right thing, even if it takes longer?

These days we're not really in a rush to get things out, as we've been in the past (e.g with the Colombian milestone). So maybe now is the right moment to take the time needed to fix this properly? 

If there was a bug in the code that this change exposes, and the fix was part of this PR then I probably would agree with the urgency to merge this. But... for once, can we try and take our time? 😃

bochecha:
Ideascube moved to Gitlab.

https://framagit.org/ideascube/ideascube/merge_requests/745

PR #747, opened 2017-03-03
Misc model stuff

bochecha:
Some small things I encountered while working on migration testing in the CI.

See the individual commits for details.


mgautierfr:
Looks good to me.

I'm impressed by the fixes you've made on the JSON field. We were three on it (write and/or review) and we still have written something that obviously doesn't work...

Issue #670, opened 2017-01-05
Ensure migrations work with real data


bochecha:
We currently can't update a lot of boxes because of #626. We need to do better in the future, and try hard for this situation to not happen again.

One thing we could do is have the CI run something like this:

```
git checkout $TAG_OF_LATEST_RELEASE
python manage.py migrate

# import some data, representative of what a box could have

git checkout $BRANCH_OF_THE_PULL_REQUEST
python manage.py migrate
```

This certainly won't catch all issues. It will only be as good as the data we load into the database.

But if it helps finding **some** broken migrations, then it's already a good thing.

And we can make that data more and more comprehensive with time, so that it keeps getting better at finding our mistakes.

PR #750, opened 2017-03-09
Add some test data

bochecha:
Eventually we will be able to use this data to verify that migrations work from the last release.

The goal is to have a CI job which runs at every pull request and:

* checks out the tag of the last release
* loads this data into the database
* checks out the branch being tested
* run the migrations

It should give us more confidence into our migrations, especially with corner cases. This data will need to live with our releases, as documented.

Of course, we don't have this data in a release yet, so we can't enable a CI job. But we can do that starting with the next release.

We can also run the checks manually with a simple make target, just like the CI would.

Relates to #670


bochecha:
Note that I still have the command I wrote to generate this random data. I didn't feel like it was useful to add it to Ideascube, since I don't see us reusing it in the future (adding it to Ideascube means we'd have to maintain it, port it to model changes, etc...), but I could be convinced of the contrary.

letompouce:
Well, one may be curious about it, and you could put it in this PR so it remains readable for the next century if we ever need it.

Makefile
```diff
@@ -45,3 +45,18 @@ test:
...
+test-data-migration:
+	set -e ; \
+	BRANCH=$$(git rev-parse --abbrev-ref HEAD); \
+	LATEST_TAG=$$(git describe --abbrev=0 --tags); \
+	\
+	echo "# Loading some data at $$LATEST_TAG"; \
+	git checkout $$LATEST_TAG; \
+	rm -fr storage; \
+	make migrate; \
+	python manage.py loaddata --database=default test-data/data.json; \
```
mgautierfr: Add a reindex here ?

docs/release_process.md
```diff
@@ -40,8 +40,18 @@ Let's make sure there are no migration conflicts with the previous release:
     $ git checkout master
     $ make migrate
```
mgautierfr: Do we keep those lines ?
This documentation version will be in the 0.22.0, we don't need to do (and document) manual migration tests with imperfect data while we have a automated migration test with better data.

test-data/README.md
```diff
@@ -0,0 +1,14 @@
+The data file in this directory has been originally created with the
+`dumpadata` command.
```
mgautierfr: `dumpdata` command.

Makefile
bochecha: To me, `dummydata` is mostly for developers who want to test the Ideascube UI/UX with some reasonable data into it. As a result, reindexing the content after `dummydata` makes sense, because without it searching in the app does not work.

However, `test-data-migration` will mostly be used in the CI, and as manual sanity checks (e.g when releasing), but only to test the migrations, not to load some data for a developer to then test the UI/UX of Ideascube. I mean, the data being loaded is complete gibberish, nothing makes sense in the UI after loading it. 😄 

Therefore, I didn't find it necessary to reindex here, as it wouldn't really be useful, but it would make the CI jobs longer.

Unless you meant for us to test reindexation in the CI as well, which is an interesting thought on its own. /me goes and ponder on this a bit...

docs/release_process.md
bochecha: > This documentation version will be in the 0.22.0

And 0.22.0 won't be able to use `make test-data-migrations`, since the previous release didn't have the test data file.

That's why for 0.22.0 I preferred keeping both documentations. I already have a branch ready which removes it, and was thinking of doing it real soon after the 0.22.0 release, so that 0.23.0 only has the new documentation.

test-data/README.md
bochecha: Good eye!

Fixed.

Makefile
mgautierfr: > Unless you meant for us to test reindexation in the CI as well, which is an interesting thought on its own. /me goes and ponder on this a bit...

This is what I meant. We had problems in the past with index and migration. I would prefer to test it also. (But, I're right, it will make the CI jobs longer. I don't know how many time it add to the job to reindex the content)

docs/release_process.md
mgautierfr: Hum. After we've make the `0.22.0` the data will be there no ? So for the `0.23.0` release (using the `0.22.0` documentation) the `make test-data-migrations` will work.
Or I'm missing something ? (In this case , we should change the next line `Starting with Ideascube 0.22.0, you can instead run:`  who imply that it will work with `0.22.0`)

bochecha: The person making the 0.22.0 release will follow the documentation as it is written at the time of the release.

Therefore, the first part of the instructions (the current ones) are kept so that we can make the 0.22.0 release. :)

As I said, I'm planning to remove them as soon as the 0.22.0 release is out.

mgautierfr: Hum, I had the documentation on readthedocs.org in mind. But I've just check and the default one is the `lastest` version not the `stable` one.

So yes, let's keep it for now and remove it after the `0.22.0` release.

Makefile
bochecha: We agreed on IRC that using this as an occasion to test the reindexation process was a good idea, as it only adds ~4 seconds to the job. (it's almost a factor 2, though)

PR #751, opened 2017-03-10
Factorize some command-related code

bochecha:
Less code, better test coverage, what's not to like?


mgautierfr:
Looks good. Nice clean.

PR #752, opened 2017-03-10
Some fixes for the catalog/mediacenter interaction

bochecha:
See individual commits.

The second one deserves some discussion though. It seems to fix the issue, and I can then view the imported media files in the mediacenter.

However, I'm not sure of the implications this might have on everything else.

@mgautierfr Please review carefully. Especially since you wrote this code and know better its ins-and-outs.


mgautierfr:
No problem for the first commit. But I'm not sure of the implication of the `recursive=True` argument for the second commit.

Can you split this PR ? This way we will merge the first commit right now and think about the second one.

bochecha:
Done, I've removed the second commit from this PR, and will open a new one for it shortly.

PR #757, opened 2017-03-16
Media cleanup with filters

bochecha:
See individual commits.

### DIFF
--- a/ideascube/mediacenter/management/commands/clean.py
+++ b/ideascube/mediacenter/management/commands/clean.py
@@ -37,11 +37,12 @@ class Command(BaseCommandWithSubcommands):
         Document.objects.filter(package_id='').delete()
         left_media_count = Document.objects.all().count()
         if left_media_count:
-            print("{} media have been installed by packages."
-                  " They have been not deleted.\n"
-                  "You must delete the corresponding package if you want to "
-                  "remove them. Use the command :\n"
-                  "catalog remove pkgid+".format(left_media_count))
+            self.stdout.write(
+                "{} media have been installed by packages and have not been deleted.\n"
+                "Remove the corresponding package(s) if you want to delete "
+                "them with the command:\n"
+                "catalog remove pkgid+".format(left_media_count))
+
         self.clean_leftover(options)
 
     def clean_leftover(self, options):
--- a/ideascube/mediacenter/management/commands/clean.py
+++ b/ideascube/mediacenter/management/commands/clean.py
@@ -58,7 +58,7 @@ class Command(BaseCommandWithSubcommands):
 
         if from_packages_count:
             self.stdout.write(
-                "{} media have been installed by packages and {} deleted.\n"
+                "\n{} media have been installed by packages and {} deleted.\n"
                 "Remove the corresponding package(s) if you want to delete "
                 "them with the command:\n"
                 "catalog remove pkgid+".format(from_packages_count, have_not))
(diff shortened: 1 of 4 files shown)


ideascube/mediacenter/management/commands/clean.py
```diff
@@ -34,7 +34,13 @@ def add_arguments(self, parser):
         clean_media.set_defaults(func=self.clean_media)
 
     def clean_media(self, options):
-        Document.objects.filter(package_id='').delete()
+        documents = Document.objects.filter(package_id='')
+        documents_count = documents.count()
+
+        self.stdout.write(
```
mgautierfr: Maybe move the write after the `delete` ?
This way, if the `delete` fails for any reason and raise a exception, we don't print that x documents have been deleted.

ideascube/mediacenter/management/commands/clean.py
```diff
@@ -34,7 +34,13 @@ def add_arguments(self, parser):
...
+            '{} documents have been deleted'.format(documents_count))
```
mgautierfr: Add a final dot to the sentence.

ideascube/mediacenter/management/commands/clean.py
```diff
@@ -37,14 +38,28 @@ def add_arguments(self, parser):
             '--type', action='append', choices=document_types,
             help='The type of media to remove, e.g "--types=image". Can be '
                  'specified multiple times to clean multiple types.')
+        clean_media.add_argument(
+            '--created-before', type=date_argument,
+            help='Only remove media created before this date.')
+        clean_media.add_argument(
+            '--created-after', type=date_argument,
+            help='Only remove media created after this date.')
```
mgautierfr: Maybe inverse the two options order. It seem more natural when specifying a range to give first the start (`--created-after`) of the range and then the end (`--created-before`). And this is the order you've used when you're filtering the queryset.

You could also improve the help the `create-after` to explicit that it is media created after **or at** this date. (`created_at__gte`)

ideascube/mediacenter/management/commands/clean.py
bochecha: Good idea, done in the latest push to this branch.

ideascube/mediacenter/management/commands/clean.py
bochecha: Done in the latest push to this branch.

ideascube/mediacenter/management/commands/clean.py
bochecha: > Maybe inverse the two options order. It seem more natural when specifying a range to give first the start (--created-after) of the range and then the end (--created-before).

Yeah, maybe... Alphabetical order is overrated anyway.

> You could also improve the help the create-after to explicit that it is media created after or at this date. (created_at__gte)

Do note that this is all `datetime` objects, not `date` ones. As a result, that `__gte` is sort of a gimmick, because realistically, it's just not going to happen that a media is created at 00:00:00 on a given day. 😆 

But yeah, sure, why not. It's still a weird corner case, so let's test it and document it.

bochecha: To be honest, the fact that it will match medias created exactly at `datetime(YYYY, MM, DD, 0, 0, 0)` when specifying `--created-after=YYYY-MM-DD` is a bit of an easter egg.

I mean, statistically, no media will ever be created exactly at midnight, **0 minutes and 0 seconds**.

But better safe than sorry, I added a test as requested. 😃

bochecha:
I think I've taken care of all your remarks, now.

mgautierfr:
This is ok for me.

PR #773, opened 2017-03-17
Move from the setting files for IDEASCUBE_NAME

bochecha:
We've had the server.site-name in-database configuration for a few
releases now, and ansiblecube sets it for new boxes.

This means the time has come to start removing it from the settings
files, to eventually remove those settings files entirely.

But before removing them from the settings files, the first thing to get
old boxes to have the configuration set in their database.

This commit adds a migration which searches for an IDEASCUBE_NAME in the
settings. If it finds one, and there was no in-database server-name set,
then the configuration gets set to what was in the settings file.

Once a box will have been upgraded to a release including this
migration, then we can remove the IDEASCUBE_NAME option from their
settings file starting with the next release.

---

This is long overdue, but better late than never. 😄 

After this, we can start removing more things from the existing settings files.

And possibly even remove some settings files for some boxes, if that's the last thing they had in it.

---

I originally thought about making a management command which ansiblecube would run after an upgrade.

However, I figured a migration would make it easier to know which boxes have made the switch:

* with a migration, we just need to look at the Ideascube version on the box
* with a management command, we would need to go through the ansible logs to figure out on which box the command has been run

I'm not entirely happy though, as I think this would be better suited for a management command, since that's a one-time thing to run, and a migration would have a (minor) overhead forever in the future (e.g when running tests, creating a new box, etc...).

I'd be happier with a management command, if we can find a simple way to know whether the command has been run for a box. Any idea?

### DIFF
--- a/ideascube/configuration/migrations/0003_ideascube_name_to_config.py
+++ b/ideascube/configuration/migrations/0003_ideascube_name_to_config.py
@@ -1,0 +1,38 @@
+from django.conf import settings
+from django.db import migrations
+
+from ideascube.configuration import set_config
+
+
+def do_migrate(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+
+    Configuration = apps.get_model('configuration', 'Configuration')
+    configs = Configuration.objects.using(db_alias)
+
+    User = apps.get_model('ideascube', 'User')
+    users = User.objects.using(db_alias)
+
+    try:
+        configs.get(namespace='server', key='site-name')
+
+    except Configuration.DoesNotExist:
+        # No value was set in the DB
+        ideascube_name = getattr(settings, 'IDEASCUBE_NAME', '')
+
+        if ideascube_name:
+            # Use the value set in the settings file
+            system_user = users.get_system_user()
+            set_config('server', 'site-name', ideascube_name, system_user)
+
+
+class Migration(migrations.Migration):
+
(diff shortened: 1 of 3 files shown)


mgautierfr:
Hum, I'm not sure about this.

Let's say we merge it now (in release 0.23) and then we remove the value from the setting latter (in release 0.30).
If a box make a update directly from 0.22 to 0.31 in one step, the migration will not found the value in the setting file (because removed in 0.30) and we will lose the value.

We will have to be sure that a box has update to a version >= 0.23 before removing the value from the setting file.
Maybe it is more on the ansiblecube side to handle this synchronization as this is BSF folks who better now witch box have been updated or not.
@letompouce @fheslouin @swbsf any idea about the better way to keep all of this in sync ?

bochecha:
> Let's say we merge it now (in release 0.23) and then we remove the value from the setting latter (in release 0.30).

This is not what I had in mind.

Let's say we release it in 0.23.

Then, when a box gets updated to a version >= 0.23, we remove `IDEASCUBE_NAME` from its settings file.

Which means that if there are boxes which never get updated, then we'll keep their settings file with an `IDEASCUBE_NAME` in it forever.

bochecha:
Ideascube moved to Gitlab.

https://framagit.org/ideascube/ideascube/merge_requests/773

PR #783, opened 2017-03-24
Better cards settings

bochecha:
This pull request doesn't change anything at all: all boxes will have the exact same cards displayed on the home page as before the changes.

That is... if I didn't make any mistake. Do check! 😛 

The main change however is that it gets us to a place where it will be easier to gradually move towards configuring the cards in the UI: since this separates the `HOME_CARDS` into 3 independent settings, we will be able to move them to the in-database configuration independently.

As it turns out, a nice side-effect of this split is that it becomes easier to override only one or another. And since most boxes were only interested in adding/removing to only one of these settings, this has a pretty-nice side-effect:

```
$ git diff --stat BEFORE_THE_SPLIT AFTER_THE_SPLIT -- ideascube/conf
 ideascube/conf/base.py              |  5 ++++-
 ideascube/conf/idb.py               |  2 +-
 ideascube/conf/idb_irq_bardarash.py | 13 +------------
 ideascube/conf/idb_rwa_mahama.py    |  5 +----
 ideascube/conf/kb.py                |  5 ++++-
 ideascube/conf/kb_bsfcampus.py      |  2 +-
 ideascube/conf/kb_cdf_fra.py        |  9 ++-------
 ideascube/conf/kb_fra_crfaix.py     | 10 ++--------
 ideascube/conf/kb_gin_conakry.py    |  2 +-
 ideascube/conf/kb_ifb_bdi.py        |  2 +-
 ideascube/conf/kb_mooc_cog.py       |  2 +-
 ideascube/conf/kb_rwa_giz.py        |  9 +--------
 12 files changed, 20 insertions(+), 46 deletions(-)
```

Not bad!

The only downside is that it changes the way we override the `HOME_CARDS`, which imposes learning a new thing for the next config files. (until we get to the point where this is all in the database, of course)

But with the rest of the cleanups in this pull request, we end up with a grand total of 108 lines removed from the settings files! That's 108 less things to consider when we move to the in-database configuration. 😄 

In addition, this pull request improves the test which verifies the settings files. We never found that they had a problem, but it still is one nevertheless.

### DIFF
--- a/ideascube/conf/base.py
+++ b/ideascube/conf/base.py
@@ -318,9 +318,6 @@ HOME_CARDS = STAFF_HOME_CARDS + [
     {
         'id': 'mediacenter',
     },
-    {
-        'id': 'khanacademy',
-    },
     # {
     #     'category': 'learn',
     #     'url': 'http://mydomain.fr',
--- a/ideascube/conf/base.py
+++ b/ideascube/conf/base.py
@@ -308,7 +308,7 @@ STAFF_HOME_CARDS = [
     },
 ]
 
-HOME_CARDS = STAFF_HOME_CARDS + [
+HOME_CARDS = [
     {
         'id': 'blog',
     },
--- a/ideascube/conf/base.py
+++ b/ideascube/conf/base.py
@@ -322,6 +322,18 @@ BUILTIN_APP_CARDS = [
 
 EXTRA_APP_CARDS = [
     # {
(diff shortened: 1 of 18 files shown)


bochecha:
I'm having second-thoughts on this one, please don't merge for now.

bochecha:
Ideascube moved to Gitlab.

https://framagit.org/ideascube/ideascube/merge_requests/783

PR #791, opened 2017-04-05
Upgrade our dependencies.

mgautierfr:
`python-networkmanager` just releases a new 2.0.1 who integrates the commit https://github.com/seveas/python-networkmanager/commit/cad4dc79efd8d29e9fe1f1cd03f76e61534af808
This commit fixes python-networkmanager not handling old networkManager version.

As so, this PR invalidates the `nm-dep` branch (commit dd6cbe309).
But commit message mentions other issues. We have to fix them.

PR #792, opened 2017-04-06
Downgrade our python-NetworkManager dependency

bochecha:
The latest releases are giving us trouble. Version 2.0.0 is a full
rewrite, which seems to have broken compatibility with the old
NetworkManager on Debian Jessie.

Git master has a patch to restore compatibility:

    https://github.com/seveas/python-networkmanager/commit/cad4dc79efd8d29e9fe1f1cd03f76e61534af808

However, even with that applied we still seem to encounter issues. So
for now, we've decided to stick with the older version.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -45,7 +45,7 @@ pytest-flakes==1.0.1
 pytest-mock==1.5.0
 pytest==3.0.7
 python-dateutil==2.6.0    # via faker, freezegun
-python-networkmanager==2.0.0
+python-networkmanager==1.2.1
 PyYAML==3.12              # via mkdocs
 requests==2.13.0
 resumable-urlretrieve==0.1.5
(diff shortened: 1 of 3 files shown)


bochecha:
> What are the issues we encounter with the python-networkmanager patch ?

Something about an invalid DBus method signature.

The problem having been found on a Monday, I admit to not having searched for it in more details.

In any case, fixing the issue might take some time, and right now ~20 boxes have the broken python-NetworkManager 2.0.0. A revert to version 1.2.1 with a quick release of Ideascube would allow us to fix these boxes while we take our time to move towards python-NetworkManager 2.0.1.

bochecha:
We've talked about this, and we agree this is the short-term way to go.

Let's try and fix things later with #791 .

mgautierfr:
+1

PR #794, opened 2017-06-16
Adding necessary files for Kolibri card

swbsf:
If anything is missing, just say so :)


bochecha:
We moved to Framagit.org, please resubmit there.

